PR #2, opened 2016-03-01
Add missing install targets

severin-lemaignan:


mkhansen-intel:
Thanks for this, we will be integrating it soon.

reaganlo:
Severin, currently this code change causes a failure in our unit tests while sourcing the "devel" area because of the change in the package.xml file. Could you please update the pull request to move the "src/realsense_camera_nodelet_plugins.xml" from the "src" directory to the top level package directory (e.g at the same level as package.xml). So that the change works for both sourcing the "devel" or "install". Thanks.

reaganlo:
Severin, we have incorporated your pull request, internally, and added my above mentioned fix. It will be part of our next release. Thanks for contributing.

reaganlo:
Severin, we have incorporated your pull request, internally, and added my above mentioned fix. It is part of our Release 1.0.2.

Thanks for contributing.

Issue #7, opened 2016-03-08
Multiple cameras question


prarobo:
I have some questions on using the ros wrapper for multiple cameras. Currently I see in your ros code a rs_context object that contains all camera device information. Your source calls the first realsense device (index 0) and pushes the data from the device into ros messages.

It will be convenient to spawn multiple nodelets where each nodelet brings up a single realsense camera. Currently if I try that, the nodelets crash. I believe this is due a single rs_context object controlling all cameras and when a new rs_context object is created (from a different nodelet), it causes some low level permissions issue. Please correct me if I am wrong.

One way I can think of getting around this problem is to just use a single nodelet to get all datastreams from all realsenses using a single rs_context object.  Is there a more elegant way to solve this problem?

I would prefer a solution where each realsense camera can be brought up from a separate nodelet. Please let me know if you want further information or clarifications.

reaganlo:
As a user, which is the preferred use case:
1) Stream data from a realsense camera by giving the unique "serial number" of the camera as input (like in the launch file)
OR 
2) The ability for the nodelet to traverse all the connected realsense cameras, fetch their unique serial numbers and stream data from them.

prarobo:
I would go with 1.

In my use case, I have multiple realsenses. I would like to have the device IDs of each realsense connected to be repeatable over system reboots. I feel connecting using a unique serial number would help.

liz-murphy:
I would also like (1). I've tried to implement this functionality in the PR above, but it is flaky.  As @prarobo said, opening multiple contexts in different processes does not seem to work well.

reaganlo:
We are currently working on implementing this feature. The first release with respect to this would be to uniquely identify a camera. Please refer to #18

reaganlo:
@liz-murphy @prarobo If you are interested in the progress of this issue, please refer to the closing comments of issue #18

reaganlo:
Fixed in release [1.0.3](https://github.com/intel-ros/realsense/releases/tag/1.0.3)

PR #8, opened 2016-03-10
Address cameras by serial number and support running multiple concurr…

liz-murphy:
…ent processes

Addresses #7

### DIFF
--- a/camera/launch/includes/realsense_r200_nodelet.launch.xml
+++ b/camera/launch/includes/realsense_r200_nodelet.launch.xml
@@ -2,6 +2,7 @@
     <!-- Name of nodelet manager, must be fully resolved -->
     <arg name="manager" />
 
+    <arg name="device_id"           default="" />
     <arg name="depth_frame_id"      default="camera_depth_optical_frame" />
     <arg name="rgb_frame_id"        default="camera_rgb_optical_frame" />
     <arg name="publish_tf"          default="false" />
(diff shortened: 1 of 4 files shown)


liz-murphy:
I see a lot of this error and often only one process starts up successfully:

[ERROR] [1457594957.962800200]: RealSense Camera - Error calling rs_create_context ( api_version:4 ): 
UVCIOC_CTRL_QUERY:UVC_SET_CUR error 5, Input/output error

reaganlo:
@liz-murphy The pull request is giving compilation errors; so I am not able to test it. Could you please look into it.

reaganlo:
@liz-murphy 
I am **assuming** that you are getting the "UVCIOC_CTRL_QUERY:UVC_SET_CUR error 5" while using **multiple cameras**. Our current code does not support multiple cameras. We have this requirement in our backlog. So with multiple cameras, you may come across errors like "UVCIOC_CTRL_QUERY:UVC_SET_CUR error 2" and "UVCIOC_CTRL_QUERY:UVC_SET_CUR error 5". Even the librealsense examples (cpp-capture) would give the similar errors with multiple cameras connected.

But you **should not** be getting this error with a **single camera**. If yes, I would suggest running the librealsense example (cpp-capture). If you get the same error with the librealsense example too, then it could be either:
- The camera firmware is not updated. The latest stable firmware version is **1.0.72.06**. To update the firmware you will need to download and run this [executable](https://downloadmirror.intel.com/25044/eng/intel_rs_dcm_r200_2.1.24.6664.exe) on a windows machine with the camera connected to a USB 3.0 port. OR
- The uvc driver patches were not applied correctly. Refer to the [README](https://github.com/intel-ros/realsense/blob/indigo-devel/camera/README.md#installation) for installation steps.

Hope this helps.

reaganlo:
@liz-murphy , FYI I have created an Issue for the enabling ros params at https://github.com/intel-ros/realsense/issues/11

Issue #25, opened 2016-04-12
xyzrgb point cloud freezes in rviz after displaying for 1 second


jgoppert:
### System Configuration

| Version          | Known Good   | Your Configuration |
| ---------------- | ------------:| ------------------:|
| Operating System | ubuntu 14.04 |                ubuntu 14.04 |
| Kernel Version   |     4.4.0-13 |             4.4.0-13 |
| R200 FW          |    1.0.72.06 |            1.0.72.06 |
| librealsense     |        0.9.1 |              master  f019a  > 0.9.1 |
| ROS                  |     Indigo  |                 Jade       |

### Expected Behavior

Start rgbd node, should display xyzrgb point cloud. (Point cloud with color from rgb image).

### Actual Behavior

xyzrgb point cloud  ( /camera/depth_registered/points ) displays for 1 second before crashing

xyz point cloud displays fine ( /camera/depth/points )

### Steps to Reproduce

Unplug the realsense usb connection and reconnect it (it won't reliably start /camera/depth_registered/points at all otherwise.

```bash
roslaunch realsense_camera realsense_r200_rgbd.launch 
rviz
```
Add /camera/depth_registered/points as PointCloud2

mdhorn:
By crash to you mean rviz stops streaming? Please confirm this is what you are seeing.

If yes, we could reproduce this and it will be investigated further.

jgoppert:
Yes, it stops streaming.  If I select the other point cloud topic it starts streaming again without rgb, but won't start rgbxyz point cloud stream again.

ffurrer:
We experienced a very similar behavior on a Ubuntu machine with an identical configuration as the one above except on Indigo. Sporadically we were able to get a single (or very few) colored point cloud message(s). In rviz, this lead to the behavior described above.
We also tested it on an OSX 10.11 machine running Jade where we got no colored point cloud at all.

mrshpot:
@jgoppert Could you check if PR #38 helps in your case?

mdhorn:
@jgoppert Have you tried release 1.0.4?  Did that resolve the issue?
If not, can you please add your system configuration information: CPU, Memory, type of storage (disk/emmc/sdcard...)

jgoppert:
@mdhorn @mrshpot fixed for me on 1.0.4

mdhorn:
Great!

jgoppert:
@mdhorn This seems to be back on latest master of realsense-ros and librealsense

```bash

[...]

PR #38, opened 2016-05-10
Work around RGBXYZ stream hangs (issue #25)

mrshpot:
Fixes Issue: #25

- Shift TF timestamps into the future to work around synchronization issues with the rgbd_launch pipeline

Apparently, this is triggered when the depth_image_proc/register nodelet
fails with a TF2 'timestamp is in the future' error and doesn't publish
the registered depth image properly. This seems to cause the ROS
Synchronizer filter to choke.

Work around by moving the published timestamp further into the future.

### DIFF
--- a/camera/src/realsense_camera_nodelet.cpp
+++ b/camera/src/realsense_camera_nodelet.cpp
@@ -926,11 +926,12 @@ namespace realsense_camera
     checkError();
 
     ros::Duration sleeper(0.1); // 100ms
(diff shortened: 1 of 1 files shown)


mkhansen-intel:
@mrshpot Can we close this PR now that the original issue #25 has been fixed?

mrshpot:
@mkhansen-intel Yes, go ahead.

PR #59, opened 2016-06-28
Prepare for Package Release

mdhorn:
- Install launch, rviz, and config files.
- Migrate package.xml to format 2.
- Update the maintainers
- Update the package dependencies

### DIFF
--- a/realsense_navigation/CMakeLists.txt
+++ b/realsense_navigation/CMakeLists.txt
@@ -9,3 +9,18 @@ find_package(catkin REQUIRED)
 
 # Declare this project as a catkin package
 catkin_package()
+
+# Install launch files
+install(DIRECTORY launch/
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/launch
+)
+
+# Install rviz files
+install(DIRECTORY rviz/
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/rviz
+)
+
+# Install config files
+install(DIRECTORY config/
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/config
+)
--- a/realsense_navigation/package.xml
+++ b/realsense_navigation/package.xml
@@ -1,23 +1,27 @@
 <?xml version="1.0"?>
-<package>
+<package format="2">
   <name>realsense_navigation</name>
   <version>1.1.0</version>
   <description>RealSense Camera navigation package allowing use of Intel 3D cameras with the navigation stack</description>
 
-  <maintainer email="amit.moran@intel.com">Amit Moran</maintainer>
   <maintainer email="matthew.c.curfman@intel.com">Matt Curfman</maintainer>
+  <maintainer email="mark.d.horn@intel.com">Mark Horn</maintainer>
 
   <license>BSD 3-clause. See license attached</license>
 
   <url type="website">http://www.ros.org/wiki/RealSense</url>
 
   <author email="amit.moran@intel.com">Amit Moran</author>
-  <author email="matthew.c.curfman@intel.com">Matt Curfman</author>
 
   <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
-  <build_depend>rospy</build_depend>
- 
-  <run_depend>roscpp</run_depend>
-  <run_depend>rospy</run_depend>
+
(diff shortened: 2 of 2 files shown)


realsense_navigation/CMakeLists.txt
```diff
@@ -9,3 +9,18 @@ find_package(catkin REQUIRED)
...
+
+# Install rviz files
+install(DIRECTORY rviz/
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/rviz
+)
+
+# Install config files
+install(DIRECTORY config/
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/config
+)
```
reaganlo, line 26: Do we need to install the package.xml too?

realsense_navigation/package.xml
```diff
@@ -1,23 +1,21 @@
...
   <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
-  <build_depend>rospy</build_depend>
- 
-  <run_depend>roscpp</run_depend>
-  <run_depend>rospy</run_depend>
+
+  <exec_depend>roslaunch</exec_depend>
+  <exec_depend>realsense_camera</exec_depend>
+  <exec_depend>rgbd_launch</exec_depend>
```
kevincwells: Do we need this exec_depend on rgbd_launch?  Shouldn't we have an exec_depend on turtlebot, since we directly include launch files in that package?

realsense_navigation/package.xml
```diff
@@ -1,23 +1,21 @@
...
-  <author email="matthew.c.curfman@intel.com">Matt Curfman</author>
 
   <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
-  <build_depend>rospy</build_depend>
- 
-  <run_depend>roscpp</run_depend>
-  <run_depend>rospy</run_depend>
+
+  <exec_depend>roslaunch</exec_depend>
```
reaganlo, line 18: Reading about "run_depend", it says in format 2, it should be replaced by two tags.
"build_export_depend"
"exec_depend"

realsense_navigation/CMakeLists.txt
mdhorn, line 26: The package.xml is automatically placed in the install directory by catkin.

realsense_navigation/package.xml
mdhorn, line 18: Are you referring to the <exec_depend>? <exec_depend> is a new tag in format 2.
`If the dependency is only used at run-time, only the <exec_depend> is needed.`

realsense_navigation/package.xml
mdhorn: Yes, we should replace rgbd_launch with packages we are directly referencing in our launch files; my mistake as I was modeling after the [ros_astra_launch](https://github.com/orbbec/ros_astra_launch) example which only used 'rgbd_launch'.
We seem to have a long list:
`turtlebot_bringup, turtlebot_navigation, map_server, rviz, rosbag, amcl, gmapping`
Does this seem to be complete?

realsense_navigation/CMakeLists.txt
reaganlo, line 26: Thanks for clarifying.

realsense_navigation/package.xml
reaganlo, line 18: Thanks for clarifying.

reaganlo:
Approved once all the exec_depends are identified and updated.

realsense_navigation/package.xml
```diff
@@ -1,23 +1,27 @@
...
+  <exec_depend>realsense_camera</exec_depend>
```
kevincwells, line 19: Correct me if I'm wrong, but I think we don't actually have a direct dependency on realsense_camera package. We have a dependency on turtlebot_bringup, and THAT will have a dependency on realsense_camera (because it launches the 3d sensor).

Otherwise this list looks accurate and complete.

mdhorn, line 19: @kevincwells Considering we have the realsense_navigation package in the same repository and we know it will indirectly require the realsense_camera package, are you opposed to leaving that dependency?

kevincwells, line 19: I'm fine with leaving it in.

I approve this request.

PR #64, opened 2016-07-06
Correct ROS Dependencies Install command

mdhorn:
Correct ROS Dependencies Install command

Fix Issue #63


reaganlo:
Approved :+1:

PR #72, opened 2016-07-17
[realsense_camera] follow ROS manner to include files in C++ codes

furushchev:
- change structure to follow more ROS manner
  - move header files to `include/realsense_camera`.
  - avoid to use #include in relative directory


mdhorn:
This is a good catch and something we should fix.
As part of this patch, we should correct the installation of the header files as well; would you apply the attached patch to your branch as part of the same PR?
[0001-Fix-the-Install-for-Includes.patch.txt](https://github.com/intel-ros/realsense/files/370354/0001-Fix-the-Install-for-Includes.patch.txt)

mdhorn:
I've rebased your pull request against the latest code, add my install fixes, and resubmitted this as PR #77 
https://github.com/intel-ros/realsense/pull/77

PR #73, opened 2016-07-17
[realsense_camera/CMakeLists.txt] fix: finding librealsense library

furushchev:
## Issue
- No clear error if library was not found in CMake.
- `find_package(catkin OPTIONAL_COMPONENTS ...)` was always overridden by `find_package(catkin REQUIRED COMPONENTS ...)`. (library path for `librealsense` was always not written in `catkin_LIBRARIES`)
- In current `CMakeLists.txt`, build system at first finds `realsense` (assumed to be installed without catkin), then re-finds `librealsense` (catkin package version). This could occur runtime error (`undefined symbol`) when `realsense` library (which is installed without catkin) is old.
## Solution
- fix: avoid to find_package catkin twice
- find `librealsense` catkin package at first, and if not found then find `realsense` library by manual

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -14,10 +14,26 @@ set(CMAKE_EXE_LINKER_FLAGS "-pie -z noexecstack -z relro -z now")
 # Flags shared libraries
 set(CMAKE_SHARED_LINKER_FLAGS "-z noexecstack -z relro -z now")
 
-find_library(REALSENSE_LIB realsense)
-
 # This is optional until the first release of librealsense as a ROS package
-find_package(catkin OPTIONAL_COMPONENTS librealsense)
+find_package(librealsense)
+
+# This should be removed once librealsense is a required component
+if(NOT librealsense_FOUND)
+  find_library(REALSENSE_LIB realsense)
+  if(NOT REALSENSE_LIB)
+    message(FATAL_ERROR "librealsense library not found")
+  endif()
+  find_path(REALSENSE_INCLUDE_DIR
+    NAMES rs rsutil
+    PATHS /usr/include /usr/local/include
+    PATH_SUFFIXES include)
+  if(REALSENSE_INCLUDE_DIR)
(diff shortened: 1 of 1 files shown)


mdhorn:
We are about to release librealsense and realsense_camera as packages. Right now they are only in the Shadow Repository.

The existing code has been tested building with the follow configurations:
- librealsense only installed without catkin (/usr/local) 
- librealsense only as a package via apt 
- librealsense installed without catkin (/usr/local) and as a package via apt [same versions]
- librealsense installed without catkin (/usr/local) and as a package via apt [/usr/local older version]
  All configuration work as expected. There is a Warning generated if you only have /usr/local version, but the build still completes successfully. 
  If there is no occurrence of librealsense, then the build will fail.

This method was done to support the cross over time of users having a /usr/local version and still working once switching to the packaged version. The packaged version will take precedence over a locally installed version. If the user builds librealsense from source using catkin_make, this version will be used.

Once the first packages are released from the ROS package server, the CMakeLists.txt will be updated to require librealsense from a package.

mdhorn:
@furushchev 
When I attempt to test your pull request on a system with only /usr/local librealsense (installed without catkin), the build fails.

`CMake Warning at ros-realsense/realsense_camera/CMakeLists.txt:18 (find_package):
  By not providing "Findlibrealsense.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "librealsense", but CMake did not find one.`

`Could not find a package configuration file provided by "librealsense" with
  any of the following names:`

```
librealsenseConfig.cmake
librealsense-config.cmake
```

`Add the installation prefix of "librealsense" to CMAKE_PREFIX_PATH or set
  "librealsense_DIR" to a directory containing one of the above files.  If
  "librealsense" provides a separate development package or SDK, be sure it
  has been installed.`

`CMake Error at ros-realsense/realsense_camera/CMakeLists.txt:33 (message):
  librealsense header files not found`

`-- Configuring incomplete, errors occurred!`

Since only have a locally installed version of librealsense is the default case, this is a requirement.

Due to this problem, this PR will not be merged.

PR #74, opened 2016-07-17
[realsense_camera/CMakeLists.txt] fix: testing

furushchev:
- testing is not needed on build time.
- we can use `CATKIN_ENABLE_TESTING` variable to disable testing on build time.
- we can run test by `catkin run_tests realsense_camera`.

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -29,7 +29,6 @@ find_package(catkin REQUIRED COMPONENTS
   message_generation
   std_msgs
   sensor_msgs
-  rostest
   pcl_ros
 )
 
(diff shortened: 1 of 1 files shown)


mdhorn:
Can you provide more information of why this change is needed?
Are you just trying to remove the requirement of having rostest installed on the system in order to build the package?

The tests that are in this package require the use of physical camera attached to the system under test. As the ROS build farm will not have the required camera, it will fail. This is why we build the test executable, but do not automatically run tests via `catkin_make run_tests`.

LorenzMeier:
@mdhorn The test build fails on a normal install. I can't build the driver without this diff.

mdhorn:
@LorenzMeier Can you provide the command you are running which is failing please?
We are building this code in our Continuous Integration for both Indigo and Kinetic as well as building the packages in the ROS Build Farm, so I want to understand what we might be missing.
thank you.

LorenzMeier:
@mdhorn Will do

PR #79, opened 2016-08-02
Change to Static Transforms for camera

mdhorn:
- Switch to using a static transform for the camera as the data does not change.
- Fixes the warning generated when the dynamic transform was used.


reaganlo:
You may want to remove this from the README or update it with a command relevant to static transforms. https://github.com/intel-ros/realsense/tree/782b2a5261c3f8063287c6cb747193e7f41581ab/realsense_camera#transform-frames

**On a side note**, not related to this PR, the current description for "tf_monitor" is incorrect. It prints on screen instead of generating a pdf. It is "view_frames" that generates the pdf.

mdhorn:
@reaganlo README updated; good catch.

rjingar:
Looks good to me. Approved.

reaganlo:
Approved :+1:

PR #96, opened 2016-08-29
use nodehandles to enable easier remapping

tfoote:
Remapping is not supported for substrings. By creating intermediate nodehandles the namespaces can be remapped in aggregate. Without this my testing of https://github.com/turtlebot/turtlebot_apps/pull/152 was failing due to partial remappings.


kevincwells:
@tfoote -- In the short term, this change as-is breaks the way we are handling multi-camera support, as we've been handling the camera namespace remapping ourselves. We are looking into tweaking this so our regression tests still pass backward compatibility.

For getting the turtlebot integration namespacing to work, we have a separate fix in the works that uses our rgbd-compatibile launch file that does not insert the additional camera namespace level.

kevincwells:
#117 incorporates this change, and thus this pull request can be closed.

PR #117, opened 2016-09-16
Topic name publishing and remapping

kevincwells:
This PR incorporates @tfoote's change to use NodeHandles to correctly publish topic names using namespaces, allowing for easier remapping of topics.

This resulted in an overall change to the topic names, where the leading "camera" namespace was no longer erroneously hardcoded in the code.  To accommodate this, the default launch files have been updated throughout to use a group namespace to properly handle camera name remapping and multiple camera support.  Unit tests have been updated accordingly.

This change brings our package more in line with the standards and conventions used by the other 3D sensor packages.

Note: This PR incorporates changes in PR #96, and thus trumps that PR.

### DIFF
--- a/realsense_camera/README.md
+++ b/realsense_camera/README.md
@@ -88,13 +88,13 @@ IR2 camera: Available only for <b>R200</b> cameras.
         Calibration data
 
 ####Services
-    get_settings (camera/get_settings)
+    get_settings (camera/driver/get_settings)
         Gets the current value of the supported camera options in "options:value" format separated by semicolon.
-    set_power < true/false > (camera/set_power < true/false >) # true - to power on the camera, false - to power off the camera
+    set_power < true/false > (camera/driver/set_power < true/false >) # true - to power on the camera, false - to power off the camera
         Sets camera power to turn camera on/off. It turns off the camera only if camera has no subscribers. Otherwise it returns false.
-    force_power < true/false > (camera/force_power < true/false >) # true - to power on the camera, false - to forcefully power off the camera
+    force_power < true/false > (camera/driver/force_power < true/false >) # true - to power on the camera, false - to forcefully power off the camera
         Sets camera power to turn camera on/off. It turns off the camera regardless of number of subscribers.
-    is_powered (camera/is_powered)
+    is_powered (camera/driver/is_powered)
         Gets current state of camera power. It checks whether camera is on or off and returns true or false respectively.
 
 ####Transform Frames
--- a/realsense_camera/launch/includes/nodelet.launch.xml
+++ b/realsense_camera/launch/includes/nodelet.launch.xml
@@ -9,7 +9,7 @@
   <arg name="ir2"          default="ir2" />
   <arg name="serial_no"    default="" />
   <arg name="usb_port_id"  default="" />
-  <node pkg="nodelet" type="nodelet" name="$(arg camera)"
+  <node pkg="nodelet" type="nodelet" name="driver"
     args="load realsense_camera/$(arg camera_type)Nodelet $(arg manager)">
     <param name="serial_no"               type="str"  value="$(arg serial_no)" />
     <param name="usb_port_id"             type="str"  value="$(arg usb_port_id)" />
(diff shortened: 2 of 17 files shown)


realsense_camera/README.md
```diff
@@ -88,13 +88,13 @@ IR2 camera: Available only for <b>R200</b> cameras.
         Calibration data
 
 ####Services
-    get_settings (camera/get_settings)
+    get_settings (camera/driver/get_settings)
```
reaganlo, line 91: @kevincwells is this a **standard** to have the nodelet name part of the service name? e.g "camera/driver/get_settings" instead of just "camera/get_settings"? or has this been done because the code has been changed to advertise the services using a private node handle "pnh_" instead of just the node handle "nh_"?

realsense_camera/launch/includes/nodelet.launch.xml
```diff
@@ -22,18 +22,9 @@
...
-    <remap from="camera/depth/points"       to="$(arg camera)/$(arg depth)/points" />
-    <remap from="camera/depth/camera_info"  to="$(arg camera)/$(arg depth)/camera_info" />
-    <remap from="camera/color/camera_info"  to="$(arg camera)/$(arg rgb)/camera_info" />
-    <remap from="camera/ir/camera_info"     to="$(arg camera)/$(arg ir)/camera_info" />
-    <remap from="camera/ir2/camera_info"    to="$(arg camera)/$(arg ir2)/camera_info" />
-    <remap from="camera/get_settings"       to="$(arg camera)/get_settings" />
-    <remap from="camera/set_power"          to="$(arg camera)/set_power" />
-    <remap from="camera/force_power"        to="$(arg camera)/force_power" />
-    <remap from="camera/is_powered"         to="$(arg camera)/is_powered" />
+    <remap from="depth"    to="$(arg depth)" />
```
reaganlo, line 25: @kevincwells Does this actually give the user an option to rename the namespaces? Instead, shouldnt they be set as parameters in this file and the code updated to read those parameters and create the node handles accordingly.
E.g  
- Add **param name="depth"             type="str"  value="$(arg depth)"**
- In "void BaseNodelet::getParameters()", add **pnh_.param("depth", depth_, DEPTH_NAMESPACE);** where depth_ is a global string variable.
- In "void BaseNodelet::advertiseTopics()", add **ros::NodeHandle depth_nh(nh_, depth_);** instead of ros::NodeHandle depth_nh(nh_, DEPTH_NAMESPACE);

realsense_camera/README.md
kevincwells: In short, yes.  And yes, this is a side effect of using the private NodeHandle ("pnh_").  If the service a node is providing is global and not specific to this node, it should be using a public or even global namespace. If the service is specific to the node (like these), then it should be using a private namespace, to encapsulate it and associate it with this node.  If you run two of these nodes at the same time at the same level, you need their service names to not interfere.

realsense_camera/launch/includes/nodelet.launch.xml
kevincwells: Yup, it does.  Give it a try -- try adding  "&lt;arg name="depth" value="foo" /&gt;" to r200_nodelet_default.launch (or any other launch file that includes nodelet.launch.xml), and you'll see that depth gets appropriately remapped to foo, within the camera group namespace.

Basically, no, we shouldn't be handling the topic remapping ourselves in code -- we should be publishing and subscribing to our published API, and let the ROS name remapping tools handle the remapping for us should a user choose to use it.  That's a large part of what this whole change is about -- by not hardcoding the topic names, this kind of remapping actually works.

mdhorn:
Assuming all of the unit tests are passing, and the "driver" is a ROS standard, then approved.

I would you you to shorten the title of the commit to be less than 50 characters as per our Git standards.

👍  approved with comments.

realsense_camera/launch/includes/nodelet.launch.xml
reaganlo: @kevincwells Awesome. Thanks for clarifying this.

kevincwells:
@mdhorn -- Thanks for the head's up; I've renamed the commit and updated.

realsense_camera/src/base_nodelet.cpp
```diff
@@ -246,23 +246,30 @@ namespace realsense_camera
...
    */
   void BaseNodelet::advertiseServices()
   {
-    get_options_service_ = nh_.advertiseService(SETTINGS_SERVICE, &BaseNodelet::getCameraOptionValues, this);
-    set_power_service_ = nh_.advertiseService(CAMERA_SET_POWER_SERVICE, &BaseNodelet::setPowerCameraService, this);
-    force_power_service_ = nh_.advertiseService(CAMERA_FORCE_POWER_SERVICE, &BaseNodelet::forcePowerCameraService, this);
-    is_powered_service_ = nh_.advertiseService(CAMERA_IS_POWERED_SERVICE, &BaseNodelet::isPoweredCameraService, this);
-}
+    get_options_service_ = pnh_.advertiseService(SETTINGS_SERVICE, &BaseNodelet::getCameraOptionValues, this);
+    set_power_service_ = pnh_.advertiseService(CAMERA_SET_POWER_SERVICE, &BaseNodelet::setPowerCameraService, this);
```
reaganlo, line 269: @tfoote Is it OK to have a standard where users are expected to make the topics, services and parameters of a nodelet instance unique, by grouping them using "**group ns**"; thereby avoiding the usage of **PrivateNodeHandle** in the code and also avoiding the usage of the nodelet name to access them.

E.g. **group ns = "camera"** will make the **topic as "camera/depth/image_raw"**, the **service as "camera/get_settings"** and **parameter as "camera/mode"**.

tfoote: Certainly you can setup services on a namespace. You should use a NodeHandle with that scope "camera"  to support appropriate remapping using the system tools. And then you would advertise the topics or services from the NodeHandle

As @kevincwells [pointed out](https://github.com/intel-ros/realsense/pull/117#r79445964) you want to differentiate driver API from interface API. So if it's a generic API such as `set_camera_info` it should be advertised in the `camera` namespace, but if it's an R200 implementation specific option then it is good to put it into a different namespace.

With this you could launch two drivers, remapping driver1 and driver2 to and both publishing to `camera` namespace but using `camera/driver1` and `camera/driver2` for their private namespaces. And conversely you could have one driver publishing two types of camera interfaces `camera1` and `camera2` and the driver could use the `multi_camera_driver` name.

I don't think we really have a standard for where the node implementation lives compared to the advertised namespace when they are different. Usually the NodeHandle and the camera namespace align and share the remapping.

PR #124, opened 2016-09-29
Add multiple R200 RGBD launch

knorth55:
This is for multiple usage of R200 camera.
You can get pointcloud with this launch.


knorth55:
no need

Issue #132, opened 2016-10-18
Depth Control Preset not settable via launch file


mdhorn:
### System Configuration

| Version | Your Configuration |
| :-- | :-- |
| Operating System | Ubuntu 16.04.1 LTS |
| Kernel | 4.4.0-42-generic |
| Backend | video4linux |
| ROS | kinetic |
| ROS RealSense | 1.5.0 |
| librealsense | 1.11.0 |
| Camera Type-Firmware | Intel RealSense R200, Camera FW: 1.0.72.06 |
### Expected Behavior

As a ROS user, I should be able to select a different default value for the R200 Depth Control Preset from a launch file by setting parameter' r200_dc_preset'.
`<param name="$(arg camera)/driver/r200_dc_preset"   type="int"  value="3" />`

> $ rosparam dump | grep r200_dc
>     r200_dc_estimate_median_decrement: 5, r200_dc_estimate_median_increment: 5, r200_dc_lr_threshold: 14,
>     r200_dc_median_threshold: 185, r200_dc_neighbor_threshold: 45, r200_dc_preset: 3,
>    r200_dc_score_maximum_threshold: 505, r200_dc_score_minimum_threshold: 5, r200_dc_second_peak_threshold: 45,
>     r200_dc_texture_count_threshold: 6, r200_dc_texture_difference_threshold: 35,
### Actual Behavior

When the parameter is added to a launch file:
`<param name="$(arg camera)/driver/r200_dc_preset"   type="int"  value="3" />`
the new preset value is set, but the individual depth control values are left as the defaults from the dynamic reconfiguration defaults.

> $ rosparam dump | grep r200_dc
>    r200_dc_estimate_median_decrement: 5, r200_dc_estimate_median_increment: 5, r200_dc_lr_threshold: **12**,
>    r200_dc_median_threshold: **235**, r200_dc_neighbor_threshold: **90**, r200_dc_preset: 3,
>    r200_dc_score_maximum_threshold: **420**, r200_dc_score_minimum_threshold: **27**, r200_dc_second_peak_threshold: **70**,
>    r200_dc_texture_count_threshold: **8**, r200_dc_texture_difference_threshold: **80**,
### Steps to Reproduce

Create r200_nodelet_dc_preset.launch

```
<!-- Sample launch file for using RealSense R200 camera with default configurations -->
<launch>
  <arg name="camera"       default="camera" />
  <arg name="camera_type"  default="R200" /> <!-- Type of camera -->
[...]

PR #137, opened 2016-10-22
Fixed a logic bug in the dc_preset override

mdhorn:
Fixes Issue: #132

General case: individual depth control parameter changes causes
the depth control preset to be ignored and changed to unused.

Special case: During initialization, first time the callback is called,
the user may wish to override the default depth control preset by
setting a new value in the launch file as a parameter -- this is the
only time the depth control preset supersedes individual values.


tripzero (approved):
LGTM

reaganlo:
Approved :+1:

PR #143, opened 2016-10-27
Fix ROS Camera Tests

mdhorn:
- Disable checks for IMU and FishEye cameras by default
- Enabled the checks for IMU and FishEye on ZR300
- Correct ZR300 to only check for 1st coefficient in FishEye distortion model.


reaganlo:
Approved :+1:

mkhansenbot (approved):
Looks good to me 👍

PR #146, opened 2016-11-02
Make building ROS unit tests optional

rojkov:
ROS users don't always need unit tests built with the driver together,
so make them conditional.

Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -26,7 +26,6 @@ find_package(catkin REQUIRED COMPONENTS
   message_generation
   std_msgs
   sensor_msgs
-  rostest
   pcl_ros
 )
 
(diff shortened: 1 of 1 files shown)


mdhorn:
@rojkov Fails to builds in the ROS Build Farm. Please fix.

mdhorn:
@rojkov I'm not opposed to the idea of this change, but I do think that if someone is downloading this code, and building it for ROS, then they have an intention to make changes. If they are making changes, then they should be running the unit tests before submitting a pull request. 
At the moment this doesn't build.

rojkov:
@mdhorn Well, the problem with the unit tests is that it's not an easy task to reproduce the environment the tests are run in. :)

I spent yesterday to do exactly that, but after many bumps I've finally ended up with the problem fixed in https://github.com/IntelRealSense/librealsense/pull/342. I wonder if there is a way to work it around without having it propagated to the `http://packages.ros.org/ros/ubuntu` package repository or setting up an Ubuntu host.

rojkov:
Fixed the build.
Turned out to be a peculiarity of `find_package`'s syntax.

rojkov:
@mdhorn is this PR ok now to merge?

mdhorn (approved):
Unit tests still available and building in the ROS Build Farm.  Approved

PR #150, opened 2016-11-14
IR stream enable

mkhansenbot:

Fixes Issue: #

- Adds enable parameters for IR and IR2 streams
- Adds new RVIZ files for use with default and RGBD launch files
- Updates tests and adds tests for IR streams

### DIFF
--- a/realsense_camera/launch/r200_nodelet_modify_params.launch
+++ b/realsense_camera/launch/r200_nodelet_modify_params.launch
@@ -9,6 +9,8 @@
   <!-- These 'arg' tags are just place-holders for passing values from test files.
   The recommended way is to pass the values directly into the 'param' tags. -->
   <arg name="enable_depth"      default="true" />
+  <arg name="enable_ir"         default="true" />
+  <arg name="enable_ir2"        default="true" />
   <arg name="enable_color"      default="true" />
   <arg name="enable_pointcloud" default="true" />
   <arg name="enable_tf"         default="true" />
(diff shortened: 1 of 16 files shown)


realsense_camera/src/r200_nodelet.cpp
```diff
@@ -459,7 +460,7 @@ namespace realsense_camera
       }
       ts_[RS_STREAM_INFRARED2] = -1;
     }
-    else if (enable_[RS_STREAM_DEPTH] == false)
+    else if (enable_[RS_STREAM_INFRARED] == false)
```
mdhorn: Should this be RS_STEAM_INFRARED2?

realsense_camera/launch/r200_nodelet_modify_params.launch
```diff
@@ -9,6 +9,8 @@
   <!-- These 'arg' tags are just place-holders for passing values from test files.
   The recommended way is to pass the values directly into the 'param' tags. -->
   <arg name="enable_depth"      default="true" />
+  <arg name="enable_ir"         default="true" />
```
mdhorn, line 12: Should these default to False?

mdhorn: These probably need to be added to the nodelet_rgbd.launch.xml file as well.

mdhorn (changes requested):
A few minor changes needed.

realsense_camera/launch/r200_nodelet_modify_params.launch
mkhansenbot: I had these as true to show how to override them to true. This was part of the request from Dave.

mkhansenbot: I don't think they belong in the rgbd launch file. AFAIK they aren't used by RGBD, and they default to false in the header file.

realsense_camera/src/r200_nodelet.cpp
mkhansenbot: Yes, good catch. I'll fix that.

realsense_camera/launch/r200_nodelet_modify_params.launch
mkhansenbot: I stand corrected, I found that they are used in the RGBD file when ir_processing is set to true, so I updated the RGBD launch file also

mdhorn (approved):
👍

reaganlo (approved):
Approved :+1:

Issue #145, opened 2016-10-31
Nodelet for ZR300 doesn't launch with default settings


rojkov:
### System Configuration

| Version               | Your Configuration   |
|:--------------------- |:---------------------|
| Operating System      | Ostro XT                |
| Kernel                |  4.4.20-yocto-standard               |
| ROS                   |      indigo               |
| ROS RealSense         |   1.6.0              |
| librealsense          | 1.10.0                |
| Camera Type-Firmware  |  Name: Intel RealSense ZR300, Camera FW: 2.0.71.13, Adapter FW: 0.0.1.16           |

### Expected Behavior

`roslaunch realsense_camera zr300_nodelet_default.launch` starts successfully

### Actual Behavior

The command `roslaunch realsense_camera zr300_nodelet_default.launch` reports the error:

    [ERROR] [1477912810.031223078]: /camera/driver - Error calling rs_get_device_extrinsics (device:0x555f955f7860, from:FISHEYE, to:COLOR, extrin:0x7fff56df48f0 ): 
    The extrinsic from FISHEYE to COLOR is not valid

Then it segfaults in the code handling the error:

    #0  0x00007ffff6033038 in __GI_raise (sig=sig@entry=6) at /usr/src/debug/glibc/2.24-r0/git/sysdeps/unix/sysv/linux/raise.c:54
    #1  0x00007ffff603448a in __GI_abort () at /usr/src/debug/glibc/2.24-r0/git/stdlib/abort.c:89
    #2  0x00007ffff607075b in __libc_message (do_abort=do_abort@entry=2, fmt=fmt@entry=0x7ffff6163dc8 "*** Error in `%s': %s: 0x%s ***\n") at /usr/src/debug/glibc/2.24-r0/git/sysdeps/posix/libc_fatal.c:175
    #3  0x00007ffff6076a36 in malloc_printerr (action=3, str=0x7ffff6160a11 "free(): invalid size", ptr=<optimized out>, ar_ptr=<optimized out>) at /usr/src/debug/glibc/2.24-r0/git/malloc/malloc.c:5046
    #4  0x00007ffff607723e in _int_free (av=0x7ffff6397b00 <main_arena>, p=<optimized out>, have_lock=0) at /usr/src/debug/glibc/2.24-r0/git/malloc/malloc.c:3902
[...]

PR #154, opened 2016-11-18
Add catches

mdhorn:
- Addresses segmentation fault reported in Issue #145
- Improved code to avoid additional double free
- Improved error messages for users
- Exception handling to catch throws from included libraries

### DIFF
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -59,7 +59,7 @@ namespace realsense_camera
   /*
    * Initialize the nodelet.
    */
-  void BaseNodelet::onInit()
+  void BaseNodelet::onInit() try
   {
     getParameters();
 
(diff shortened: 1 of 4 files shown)


realsense_camera/src/base_nodelet.cpp
```diff
@@ -45,8 +45,12 @@ namespace realsense_camera
...
-    checkError();
+    if (rs_context_)
+    {
+      rs_delete_context(rs_context_, &rs_error_);
+      rs_context_ = NULL;
+      checkError();
+    }
 
     ROS_INFO_STREAM(nodelet_name_ << " - Stopping...");
     ros::shutdown();
```
mkhansenbot: Can you please add a check ros::isShuttingDown() here so that we don't call the shutdown again?

kevincwells (approved):
These all seem like good checks, and I appreciate the new error messages.

mkhansenbot (approved):
Looks good to me, did it pass the test suite? 👍

PR #159, opened 2016-12-01
Changed nodelet to use camera timestamps

mkhansenbot:
- Changes nodelet to use camera timestamps + camera start time instead of ros::Time::now() for frames


mdhorn (approved):
Looks good to me. 👍

reaganlo (approved):
Approved, assuming that the ros::Time::now() used in the tf_static timestamp is not way off from the timestamp generated from the cameras to cause any issues in downstream packages like navigation.

PR #160, opened 2016-12-01
Frame callbacks

mkhansenbot:

Fixes Issue: #

- Changes all streams to using callbacks instead of using rs_wait_for_frames
- Adds needed mutex to imu callback for zr300
- Refactors code to remove duplicated code
- Removes the topic_thread since rs_wait_for_frames is removed and everything done via callbacks

### DIFF
--- a/realsense_camera/include/realsense_camera/base_nodelet.h
+++ b/realsense_camera/include/realsense_camera/base_nodelet.h
@@ -162,7 +162,7 @@ namespace realsense_camera
     virtual std::string startCamera();
     virtual std::string stopCamera();
     virtual void publishTopic(rs_stream stream_index, rs::frame &  frame);
-    virtual void getStreamData(rs_stream stream_index, rs::frame &  frame);
+    virtual void setImageData(rs_stream stream_index, rs::frame &  frame);
     virtual void publishPCTopic();
     virtual void publishStaticTransforms();
     virtual void checkError();
--- a/realsense_camera/include/realsense_camera/base_nodelet.h
+++ b/realsense_camera/include/realsense_camera/base_nodelet.h
@@ -133,7 +133,6 @@ namespace realsense_camera
     bool enable_pointcloud_;
     bool enable_tf_;
     const uint16_t *image_depth16_;
-    float depth_scale_meters_;
     cv::Mat cvWrapper_;
     boost::mutex frame_mutex_[STREAM_COUNT];
 
--- a/realsense_camera/include/realsense_camera/base_nodelet.h
+++ b/realsense_camera/include/realsense_camera/base_nodelet.h
@@ -82,7 +82,7 @@ namespace realsense_camera
     // Interfaces.
     virtual void onInit();
     virtual ~BaseNodelet();
-    virtual void prepareTopics();
+    virtual void setDepthEnable(bool &enable_depth);
     virtual bool getCameraOptionValues(realsense_camera::CameraConfiguration::Request & req,
         realsense_camera::CameraConfiguration::Response & res);
     virtual bool setPowerCameraService(realsense_camera::SetPower::Request & req,
@@ -133,11 +133,8 @@ namespace realsense_camera
     bool enable_pointcloud_;
     bool enable_tf_;
     const uint16_t *image_depth16_;
-    boost::shared_ptr<boost::thread> topic_thread_;
     float depth_scale_meters_;
     cv::Mat cvWrapper_;
-    bool start_camera_ = true;
-    bool start_stop_srv_called_ = false;
     boost::mutex frame_mutex_[STREAM_COUNT];
 
     struct CameraOptions
--- a/realsense_camera/include/realsense_camera/base_nodelet.h
+++ b/realsense_camera/include/realsense_camera/base_nodelet.h
@@ -132,13 +132,13 @@ namespace realsense_camera
     float max_z_ = -1.0f;
     bool enable_pointcloud_;
     bool enable_tf_;
-    bool duplicate_depth_color_;
     const uint16_t *image_depth16_;
     boost::shared_ptr<boost::thread> topic_thread_;
     float depth_scale_meters_;
     cv::Mat cvWrapper_;
     bool start_camera_ = true;
     bool start_stop_srv_called_ = false;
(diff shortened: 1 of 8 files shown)


mkhansen-intel:
I'm not sure why this is failing. I think ros.build.org is down. I can't access their web server. Investigating...

realsense_camera/src/base_nodelet.cpp
```diff
@@ -514,68 +517,56 @@ namespace realsense_camera
...
+    checkError();
+  }
+
+  /*
+   * Set the streams according to their corresponding flag values.
+   */
+  void BaseNodelet::setStreams()
+  {
+    // Enable streams
+    for (int stream=0; stream < STREAM_COUNT; stream++)
```
mdhorn, line 560: Change the type of stream to be type 'rs_stream' and remove the casts.

realsense_camera/src/base_nodelet.cpp
```diff
@@ -724,81 +726,94 @@ namespace realsense_camera
...
-  void BaseNodelet::getStreamData(rs_stream stream_index)
+  void BaseNodelet::getStreamData(rs_stream stream_index, rs::frame & frame)
   {
     if (stream_index == RS_STREAM_DEPTH)
     {
       // fill depth buffer
-      image_depth16_ = reinterpret_cast<const uint16_t *>(rs_get_frame_data(rs_device_, stream_index, 0));
-      if (depth_scale_meters_ == MILLIMETER_METERS)
+      image_depth16_ = reinterpret_cast<const uint16_t *>(frame.get_data());
+      float depth_scale_meters = rs_get_device_depth_scale(rs_device_, &rs_error_);
```
mdhorn, line 735: Should this function be called in here? getStreamData is called the most frequently in the code.
Can we initialize this something else?  Or make it a local static and initialize it once.

At a minimum we need to add a check error for the rs_error_

realsense_camera/src/base_nodelet.cpp
```diff
@@ -724,81 +726,94 @@ namespace realsense_camera
...
+  /*
+   * Publish topic.
+   */
+  void BaseNodelet::publishTopic(rs_stream stream_index, rs::frame &frame) try
+  {
+    // mutex to ensure only one frame per stream is processed at a time
+    boost::mutex::scoped_lock lock(frame_mutex_[stream_index]);
 
-        publishTopics();
+    if (rs_is_stream_enabled(rs_device_, (rs_stream) stream_index, 0) == 1)
```
mdhorn: Since we are called by the callback, we have to be enabled. We should remove this check.

realsense_camera/include/realsense_camera/base_nodelet.h
```diff
@@ -164,14 +160,15 @@ namespace realsense_camera
     virtual void disableStream(rs_stream stream_index);
     virtual std::string startCamera();
     virtual std::string stopCamera();
-    virtual void publishTopics();
-    virtual void publishTopic(rs_stream stream_index);
-    virtual void getStreamData(rs_stream stream_index);
+    virtual void publishTopic(rs_stream stream_index, rs::frame &  frame);
+    virtual void getStreamData(rs_stream stream_index, rs::frame &  frame);
```
mdhorn: Rename to setImageData

realsense_camera/src/base_nodelet.cpp
```diff
@@ -724,81 +726,94 @@ namespace realsense_camera
   /*
    * Copy frame data from realsense to member cv images.
    */
-  void BaseNodelet::getStreamData(rs_stream stream_index)
+  void BaseNodelet::getStreamData(rs_stream stream_index, rs::frame & frame)
```
mdhorn: Rename to setImageData

realsense_camera/include/realsense_camera/base_nodelet.h
```diff
@@ -131,13 +131,9 @@ namespace realsense_camera
...
     bool enable_pointcloud_;
     bool enable_tf_;
-    bool duplicate_depth_color_;
     const uint16_t *image_depth16_;
-    boost::shared_ptr<boost::thread> topic_thread_;
-    float depth_scale_meters_;
     cv::Mat cvWrapper_;
-    bool start_camera_ = true;
-    bool start_stop_srv_called_ = false;
+    boost::mutex frame_mutex_[STREAM_COUNT];
```
reaganlo: Where is this "frame_mutex_" used?

realsense_camera/src/base_nodelet.cpp
mkhansenbot: The value rs_get_device_depth_scale returns is device.config.depth_scale a member variable in the rs_device_ object. Do you still think it's an issue?

realsense_camera/src/base_nodelet.cpp
mkhansenbot: Since STREAM_COUNT is an int, I can't change stream to an rs_stream type. Instead, I'm using static_cast<rs_stream>(stream) to pass the stream var.

mdhorn: Why not cast STREAM_COUNT once instead of stream twice?

mdhorn: Or can we change STREAM_COUNT to a rs_stream type?

mkhansenbot: No ++ operator is defined for the enum type rs_stream, so you get this compiler error:
error: no ‘operator++(int)’ declared for postfix ‘++’ [-fpermissive]
     for (rs_stream stream; stream < (rs_stream) STREAM_COUNT; stream++)

mkhansenbot: We cant change STREAM_COUNT to an rs_stream type, since we are setting the value to 5 in constants.h. I would like us to actually remove the constant in a future patch and do this a different way, as I agree this is non-ideal. I would prefer a member variable num_streams_ that isn't hard-coded and varies depending on the camera type, set in the onInit() function for each type to the correct value for that camera. But that's another patch for another day.

mdhorn (approved):
Previous discussion items implemented.

LGTM 👍

reaganlo (approved):
Approved

PR #163, opened 2016-12-14
Make system wrapper function generic

mdhorn:
- Mark error message more generic for systemWrapper
- Add previously missing ROS dependency of boost


reaganlo (approved):
Approved

PR #164, opened 2016-12-15
Change color stream default to 30fps

mdhorn:
Improve reliability of the camera by reducing the color stream
to 30fps by default. This will be the highest frame rate which will
be validate with VGA or higher resolutions for use with ROS.

Default files will no longer use the librealsense preset modes
as they will be depreciated in the future.

### DIFF
--- a/realsense_camera/launch/f200_nodelet_default.launch
+++ b/realsense_camera/launch/f200_nodelet_default.launch
@@ -4,7 +4,20 @@
   <arg name="camera_type"  default="F200" /> <!-- Type of camera -->
   <arg name="serial_no"    default="" />
   <arg name="usb_port_id"  default="" /> <!-- USB "Bus#-Port#" -->
-  <arg name="manager"      value="$(arg camera)_nodelet_manager" />
+  <arg name="manager"      default="nodelet_manager" />
+
+  <!-- These 'arg' tags are just place-holders for passing values from test files.
+  The recommended way is to pass the values directly into the 'param' tags. -->
+  <arg name="mode"              default="manual" />
+  <arg name="depth_width"       default="640" />
+  <arg name="depth_height"      default="480" />
+  <arg name="color_fps"         default="30" />
+
+  <param name="$(arg camera)/driver/mode"              type="str"  value="$(arg mode)" />
+  <param name="$(arg camera)/driver/depth_width"       type="int"  value="$(arg depth_width)" />
+  <param name="$(arg camera)/driver/depth_height"      type="int"  value="$(arg depth_height)" />
+  <param name="$(arg camera)/driver/color_fps"         type="int"  value="$(arg color_fps)" />
+  <!-- Refer to the Wiki http://wiki.ros.org/realsense_camera for list of supported parameters -->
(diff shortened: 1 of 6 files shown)


realsense_camera/launch/r200_nodelet_default.launch
```diff
@@ -4,7 +4,36 @@
   <arg name="camera_type"  default="R200" /> <!-- Type of camera -->
   <arg name="serial_no"    default="" />
   <arg name="usb_port_id"  default="" /> <!-- USB "Bus#-Port#" -->
-  <arg name="manager"      value="$(arg camera)_nodelet_manager" />
+  <arg name="manager"      value="nodelet_manager" />
+
+  <!-- These 'arg' tags are just place-holders for passing values from test files.
+  The recommended way is to pass the values directly into the 'param' tags. -->
+  <arg name="enable_depth"      default="true" />
+  <arg name="enable_ir"         default="true" />
```
mkhansenbot: enable_ir and enable_ir2 defaults should be 'false'

realsense_camera/launch/f200_nodelet_default.launch
```diff
@@ -4,7 +4,34 @@
   <arg name="camera_type"  default="F200" /> <!-- Type of camera -->
   <arg name="serial_no"    default="" />
   <arg name="usb_port_id"  default="" /> <!-- USB "Bus#-Port#" -->
-  <arg name="manager"      value="$(arg camera)_nodelet_manager" />
+  <arg name="manager"      value="nodelet_manager" />
```
mkhansenbot: I realize you didn't change this, but should this be default= instead of value=? If it is default, it can be overridden by the command line if it is value, it is essentially hard-coded to "nodelet_manager". See this: http://wiki.ros.org/roslaunch/XML/arg. Same would apply for all the camera types.

mdhorn: @reaganlo I thought we had made this a value instead of default to prevent the user from changing the name of the nodelet manager?

reaganlo: Yes, we have defaults only for those fields that are likely to be changed by users.

reaganlo: @mdhorn On second thoughts I think lets keep it consistent and have everything as "defaults". I feel there is no harm in doing so.

reaganlo (commented):
With these changes, do we even need to have the "modify_params" launch files for all the cameras? Because the "default" launch files itself shows how params can be set in our cameras.

realsense_camera/launch/f200_nodelet_default.launch
mdhorn: @reaganlo I'm reformatting the contents of the default files. I've tested, and I can switch to manual mode and ONLY set manual mode and the color FPS -- much simpler default file.
I'll be resubmitting this PR shortly.

reaganlo (approved):
LGTM

mkhansenbot (approved):
Approved 👍

PR #168, opened 2016-12-16
Fix F200/SR300 tests for new default 30fps

mdhorn:
The F200 and SR300 tests were attempting too much code reuse
by include the r200_nodelet_default.launch file.
When the Default launch files were changed to set 30fps as the new
default for the color stream, F200/SR300 also had to set new defaults
for the Depth stream to prevent using the invalid R200 depth default.
The R200 default file does NOT pass the depth_width and depth_height
to the nodelet so the override values were lost.

### DIFF
--- a/realsense_camera/test/f200_nodelet_disable_color.test
+++ b/realsense_camera/test/f200_nodelet_disable_color.test
@@ -1,6 +1,16 @@
 <launch>
-  <arg name="camera_type"  value="F200" />
-  <include file="$(find realsense_camera)/test/r200_nodelet_disable_color.test">
+  <arg name="camera"       value="camera" />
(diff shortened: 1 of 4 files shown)


mkhansenbot (approved):
Approved 👍

kevincwells (approved):
It's always bothered me that the tests include the R200 tests like this, because it was only a matter of time until the cameras diverged.  The way it used to be (and still is in many of the other tests) takes advantage of a weird quirk that, say, "r200_nodelet_default.launch" still actually has the camera field parameterized and can thus be used to launch an F200. Technically works, but is dangerous.

I would almost argue that we change all of the tests over to be unique, but for maintenance purposes I suppose it's simpler to just change them as needed when issues like this arise.

Issue #120, opened 2016-09-20
rosrun tf view_frames shows no tf being published


sinhau:
### System Configuration

Please complete Your Configuration detail below. Refer to the BKC at [Configuration](../realsense_camera/README.md#configuration).

| Version | Your Configuration |
| :-- | :-- |
| Operating System | Ubuntu 16.04.1 LTS |
| Kernel | 4.4.0-38-generic |
| Backend | Video4Linux |
| ROS | ROS Kinetic |
| ROS RealSense | 1.4.0 |
| librealsense | 0.9.1 (although I used git checkout tags/stable so I would think this would say 0.9.2) |
| Camera Type-Firmware | R200 1.0.72.06 |

I'm launching two R200 cameras using the r200_nodelet_multiple_cameras.launch.  I modified the launch file as shown below to change the `enable_tf` parameter to true.  When I run the launch file, both cameras launch and I am getting all the data.  The ROS logs even shows that cameras should be publishing transforms (see terminal output below)  However, when I run `rosrun tf view_frames`, the output frames.pdf file says no tf data received. For some reason no tf is being published. Moreover, when I run `rostopic echo /tf_static`, I only see camera2 tf's.  
### r200_nodelet_multiple_cameras.launch

```
<!-- Sample launch file for using multiple RealSense R200 cameras -->
<launch>
  <arg name="manager" value="nodelet_manager" /> <!-- Single nodelet manager for all cameras -->
  <node pkg="nodelet" type="nodelet" name="$(arg manager)" args="manager" output="screen"/>

  <!-- MY MODS -->
  <param name="camera1/enable_pointcloud" type="bool" value="false" />
  <param name="camera1/enable_tf"         type="bool" value="true" />
  <param name="camera2/enable_pointcloud" type="bool" value="false" />
  <param name="camera2/enable_tf"         type="bool" value="true" />
  <!-- ======= -->

  <!-- User must include the .xml file for each camera with unique values for "camera" and 
       either the "serial_no" or "usb_port_id" of the camera.
       "camera" should be a user friendly string that follows the ROS Names convention. -->
  <include file="$(find realsense_camera)/launch/includes/nodelet.launch.xml">
    <arg name="manager"      value="$(arg manager)" />
    <arg name="camera"       value="camera1" />
    <arg name="camera_type"  value="R200" /> <!-- Type of camera -->
    <arg name="serial_no"    value="2431002398" /> <!-- Note: Replace with actual serial number -->
[...]

PR #169, opened 2016-12-17
Add Dynamic Transforms support -- multi-cam

mdhorn:
Fixes Issue: #120 

Due to a ROS bug which prevents publishing more than one static
transform in separate processes, enable the use of dynamic
camera transforms when needed for multiple camera support.

Static transforms are still the default.

See:
  ros/ros_comm#146
  ros/geometry2#181

### DIFF
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -41,7 +41,7 @@ namespace realsense_camera
    */
   BaseNodelet::~BaseNodelet()
   {
-    if (enable_tf_ == true && enable_tf_dynamic_ == false)
+    if (enable_tf_ == true && enable_tf_dynamic_ == true)
     {
       transform_thread_->join();
     }
@@ -96,7 +96,7 @@ namespace realsense_camera
     setStreams();
     startCamera();
 
-    // Start tranforms thread
+    // Start transforms thread
     if (enable_tf_ == true)
     {
       getCameraExtrinsics();
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -41,6 +41,11 @@ namespace realsense_camera
    */
   BaseNodelet::~BaseNodelet()
   {
+    if (enable_tf_ == true && enable_tf_dynamic_ == false)
+    {
+      transform_thread_->join();
+    }
+
     stopCamera();
 
     if (rs_context_)
@@ -91,10 +96,20 @@ namespace realsense_camera
     setStreams();
     startCamera();
 
-    // Start publishing tranforms
+    // Start tranforms thread
     if (enable_tf_ == true)
     {
-      publishStaticTransforms();
+      getCameraExtrinsics();
+
+      if (enable_tf_dynamic_ == true)
+      {
+        transform_thread_ =
+          boost::shared_ptr<boost::thread>(new boost::thread (boost::bind(&BaseNodelet::prepareTransforms, this)));
+      }
+      else
+      {
+        publishStaticTransforms();
+      }
     }
 
     // Start dynamic reconfigure callback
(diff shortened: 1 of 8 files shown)


realsense_camera/src/base_nodelet.cpp
```diff
@@ -41,6 +41,11 @@ namespace realsense_camera
    */
   BaseNodelet::~BaseNodelet()
   {
+    if (enable_tf_ == true && enable_tf_dynamic_ == false)
```
mkhansenbot: I think this should be enable_tf_dynamic_ == true

Shouldn't it?

realsense_camera/src/base_nodelet.cpp
```diff
@@ -91,10 +96,20 @@ namespace realsense_camera
     setStreams();
     startCamera();
 
-    // Start publishing tranforms
+    // Start tranforms thread
```
mkhansenbot: Can we fix the spelling of transforms?

mkhansenbot (changes requested):
One minor functional change, and one cosmetic change requested, see comments inline. I tested on R200 and ZR300, seems to work. Tested both together with separate launch files and that worked too when setting the enable_tf_dynamic=true.

Also, can we update the PR title and the description comments to reflect that static is still the default, we're just adding dynamic as an option.

kevincwells (approved):
LGTM from code perspective.  Only thing that stood out was the true/false change Matt saw and you already fixed.

Agreed that the PR title and subject needs to be changed if static is still the actual default as it is here.

mkhansenbot (approved):

realsense_camera/src/base_nodelet.cpp
```diff
@@ -1066,6 +1087,82 @@ namespace realsense_camera
...
+
+  /*
+   * Prepare transforms.
+   */
+  void BaseNodelet::prepareTransforms()
+  {
+    // Publish transforms for the cameras
+    ROS_INFO_STREAM(nodelet_name_ << " - Publishing camera transforms (/tf)");
+
+    ros::Rate loop_rate(1);
```
severin-lemaignan, line 1152: Any reason for such a slow loop? tf transforms are traditionally published at ~50Hz.

Additional parameter proposed in #221

mdhorn: The data is not changing, hence the slow loop rate.

The default for transforms is to publish them with a static transform, but due to a ROS bug (and a camera driver bug for F200/SR300) the dynamic transform option was added.
See http://wiki.ros.org/realsense_camera#Transform_Restrictions

doronhi:
@pavloblindnology , Somehow, until very recently I remained unaware of your work.
Would you like to open a new PR so it can be merged?
I would also like to add the `publish_tf` flag again, as was implemented [here](https://github.com/mdhorn/realsense/blob/cb6e27badbf7a962de94d76b89a1a84b9f2da1e2/realsense_camera/src/base_nodelet.cpp#L100)
The issue was referenced again at #9. 
What do you think?

pavloblindnology:
@doronhi Ok, I'll open PR.
AFAIU, `enable_tf` is intended to turn off TF generation at all (either static or dynamic), while `tf_dynamic` is used to switch between static and dynamic. Correct?
May I suggest to introduce a single variable, like `tf_generation` which can have 3 values: off, static, dynamic. This would get rid of all the mess with having 2 variables with unclear meaning (`enable_tf`, `publish_tf`).

doronhi:
Thanks!
I think `enable_tf` was already deleted. The last state was to have `publish_tf` which meant, well, to publish tf, and `tf_dynamic` which, if set to true, would cause the sending the dynamic_tf instead of the static_tf.

Maybe you are right and a 3 values variable is clearer. Since I don't know how to conduct a survey, and we need to update the README.md anyway, it's totally fine by me :)

And if you discover along the way that that solution has some unforeseen complications, the other solution is fine too.

PR #170, opened 2016-12-17
Roslint

mdhorn:
Fix errors reported by roslint in code base.

Some C++11 items suppressed for now.

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -27,6 +27,7 @@ find_package(catkin REQUIRED COMPONENTS
   std_msgs
   sensor_msgs
   pcl_ros
+  roslint
 )
 
 add_message_files(
@@ -55,6 +56,10 @@ generate_dynamic_reconfigure_options(
   cfg/sr300_params.cfg
   cfg/zr300_params.cfg
 )
+
+# ROS Lint the code
+roslint_cpp()
+roslint_add_test()
 
 #################################
 # catkin specific configuration #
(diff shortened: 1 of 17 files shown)


kevincwells:
I assume you rebased this on top of your dynamic transform commits in order to roslint them too. I also assume you'll do some ugly rebasery to swap in the squashed commit post merge before you try to merge this one?

mdhorn:
@kevincwells yes, rebased on the transform PR. once the other PR is merged only the two commits for roslint will get merged.

kevincwells (commented):
I'm nothing if not a pedant, so I highly approve of all these nitpicky changes linting enforces.

Appreciate moving the imports out of base_nodelet.h and down into where they're actually used.  (Trying to trace an include back to some base class that doesn't even use it is always one of my least favorite tasks.)

Thank you for getting rid of the C-style casts.  Is there a reason you had to use reinterpret_cast for the uint16_t pointer in camera_core.cpp test file?

I can't actually build this from home so I'm hesitant to put an "Approve" stamp on it, but assuming it builds, I'm all for cleaner code.

mdhorn:
@kevincwells If you are referring to the cast on line 168 (was 133), it was because roslint said:
`realsense/realsense_camera/test/camera_core.cpp:168:  Using C-style cast.  Use reinterpret_cast<uint16_t *>(...) instead  [readability/casting] [4]`

Without the cast we pass lint, but not the compile:
error: cannot convert ‘uchar* {aka unsigned char*}’ to uint16_t* {aka short unsigned int*}’

So leaving the cast in place.

mdhorn:
@kevincwells @mkhansen-intel 
Rebase with the new indigo-devel and now this PR Is just the two commits related to roslint.
Verified after rebase the code is both lint-free and builds; basic functionality with R200 passing.

mkhansenbot (approved):
Approved 👍

PR #173, opened 2016-12-20
Fix Multi-cam example launch file

mdhorn:
Leading '/' required to ensure both cameras use the same
nodelet manager due to group name space.


reaganlo (approved):
LGTM :)

PR #174, opened 2016-12-20
Do not register IR callback if stream disabled

mdhorn:
Work around bug in librealsense where the callback for
IR is called even if the stream is disable if IR2 stream
is enabled.
See https://github.com/IntelRealSense/librealsense/issues/393


reaganlo (approved):
Typo in comment.. "disable" = "disabled"

mkhansenbot (commented):
Do we know if the same thing can occur on other streams? Should we check the enable for every stream before setting up the callback for that stream?

mdhorn:
@mkhansen-intel We did check color, depth, and both IR streams, but the bug appears to just be mistaking calling the IR callback when IR2 is enabled.

PR #175, opened 2016-12-20
Start IMU in startCamera

mkhansenbot:
**Please ensure the above *guidelines for contributing* are met.**

Fixes Issue: #

Changes proposed in this pull request:
- This fixes the try_for_lock timeout error that sometimes occurs for ZR300 on some systems
- It changes the way the IMU is started for ZR300 to start IMU and streams using device->start(rs_source::RS_SOURCE_ALL)
- This removes the call to rs_start_source(...) for the IMU
- Doing so also aligns the camera and IMU timestamps

PR #176, opened 2016-12-21
Allow Disabling of IMU after PR#175

mdhorn:
Refactoring PR #175 broke the logic for disabling IMU


mkhansenbot (approved):
Good catch. Approved 👍

PR #180, opened 2017-01-03
Added retry in tests to avoid random failures

nwang2:
Some random failures  were observed during unit tests in CI, so added retry="3" in tests to avoid these failures. Triggered 3 rounds of CI testing with this change, and got 100% passed.


mdhorn (approved):
Looks Good To Me 👍 
How many times did you repeat the testing with the retry set to 3?

mkhansenbot (approved):
Approved :+1:

PR #182, opened 2017-01-06
Generate Warning for non-validated camera firmware

arbrindle:

Fixes Issue: #

- Checks camera firmware version against the validated firmware version. 
- Prints warning if the camera firmware version is different from validated firmware version.
- For ZR300, also checks the adapter firmware and motion module firmware.

### DIFF
--- a/realsense_camera/include/realsense_camera/base_nodelet.h
+++ b/realsense_camera/include/realsense_camera/base_nodelet.h
@@ -176,6 +176,8 @@ protected:
   virtual bool checkForSubscriber();
   virtual void wrappedSystem(std::vector<std::string> string_argv);
   virtual void setFrameCallbacks();
+  virtual std::string checkFirmwareValidation(std::string fw_type, std::string current_fw, std::string camera_name,
+        std::string camera_serial_number);
   std::function<void(rs::frame f)> depth_frame_handler_, color_frame_handler_, ir_frame_handler_;
 };
 }  // namespace realsense_camera
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -304,7 +304,7 @@ namespace realsense_camera
       }
       // print camera details
       detected_camera_msg = detected_camera_msg +
-            "\n\t\t\t\t - Serial No: " + camera_serial_number + ", USB Port ID: " +
+            "\n\t\t\t\t- Serial No: " + camera_serial_number + ", USB Port ID: " +
             rs_get_device_usb_port_id(rs_detected_device, &rs_error_) +
             ", Name: " + camera_name +
             ", Camera FW: " + camera_fw;
@@ -314,7 +314,7 @@ namespace realsense_camera
 
       if (!camera_warning_msg.empty())
       {
-        warning_msg = warning_msg + "\n\t\t\t\t-" + camera_warning_msg;
+        warning_msg = warning_msg + "\n\t\t\t\t- " + camera_warning_msg;
       }
 
       if (rs_supports(rs_detected_device, RS_CAPABILITIES_ADAPTER_BOARD, &rs_error_))
@@ -327,7 +327,7 @@ namespace realsense_camera
               camera_serial_number);
         if (!adapter_warning_msg.empty())
         {
-          warning_msg = warning_msg + "\n\t\t\t\t-" + adapter_warning_msg;
+          warning_msg = warning_msg + "\n\t\t\t\t- " + adapter_warning_msg;
         }
       }
 
@@ -341,7 +341,7 @@ namespace realsense_camera
               camera_serial_number);
         if (!motion_module_warning_msg.empty())
         {
-          warning_msg = warning_msg + "\n\t\t\t\t-" + motion_module_warning_msg;
+          warning_msg = warning_msg + "\n\t\t\t\t- " + motion_module_warning_msg;
         }
       }
       ROS_INFO_STREAM(nodelet_name_ + detected_camera_msg);
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -1299,7 +1299,7 @@ namespace realsense_camera
     if (current_fw != validated_firmware)
     {
       warning_msg = camera_serial_number + "'s current " + fw_type + " firmware is " + current_fw +
-            ". Validated " + fw_type + " firmware is " + validated_firmware;
+            ", Validated " + fw_type + " firmware is " + validated_firmware;
     }
     return warning_msg;
   }
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -31,6 +31,7 @@
 #include <string>
 #include <algorithm>
 #include <vector>
+#include <map>
 #include <realsense_camera/base_nodelet.h>
 
 using cv::Mat;
@@ -41,6 +42,8 @@ using std::vector;
 PLUGINLIB_EXPORT_CLASS(realsense_camera::BaseNodelet, nodelet::Nodelet)
 namespace realsense_camera
 {
+  const std::map<std::string, std::string> CAMERA_NAME_TO_VALIDATED_FIRMWARE
+        (MAP_START_VALUES, MAP_START_VALUES + MAP_START_VALUES_SIZE);
   /*
    * Nodelet Destructor.
    */
@@ -275,14 +278,24 @@ namespace realsense_camera
   {
     // print list of detected cameras
     std::vector<int> camera_type_index;
-    std::string detected_camera_msg = " - Detected the following cameras:";
+
     for (int i = 0; i < num_of_cameras; i++)
     {
+      std::string detected_camera_msg = " - Detected the following camera:";
+      std::string warning_msg = " - Detected unvalidated firmware:";
       // get device
       rs_device* rs_detected_device = rs_get_device(rs_context_, i, &rs_error_);
 
+      // get camera serial number
+      std::string camera_serial_number = rs_get_device_serial(rs_detected_device, &rs_error_);
+      checkError();
+
       // get camera name
       std::string camera_name = rs_get_device_name(rs_detected_device, &rs_error_);
+      checkError();
+
+      // get camera firmware
+      std::string camera_fw = rs_get_device_firmware_version(rs_detected_device, &rs_error_);
       checkError();
 
       if (camera_name.find(camera_type_) != std::string::npos)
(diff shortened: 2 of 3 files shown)


realsense_camera/src/base_nodelet.cpp
```diff
@@ -1206,6 +1208,40 @@ namespace realsense_camera
...
+    if (typeOfFirmware == "MOTION_MODULE")
+    {
+      currentFirmware = rs_get_device_info(rs_detected_device,
+            RS_CAMERA_INFO_MOTION_MODULE_FIRMWARE_VERSION, &rs_error_);
+      checkError();
+    }
+
+    if (currentFirmware != validatedFirmware)
+    {
+      ROS_WARN_STREAM(typeOfFirmware << " firmware is " << currentFirmware <<
```
mkhansenbot: Please use the nodelet_name_ member variable in the warning message so that the user knows which nodelet is printing the warning. This is a standard we follow in our nodelet. See other ROS_INFO_STREAM() statements for examples.

realsense_camera/include/realsense_camera/base_nodelet.h
```diff
@@ -173,6 +173,8 @@ class BaseNodelet: public nodelet::Nodelet
   virtual void publishDynamicTransforms();
   virtual void prepareTransforms();
   virtual void checkError();
+  virtual void checkFirmwareValidation(std::string validatedFirmware,
```
mkhansenbot: Please change all the inputs to const to ensure they can't be changed.

realsense_camera/include/realsense_camera/base_nodelet.h
```diff
@@ -173,6 +173,8 @@ class BaseNodelet: public nodelet::Nodelet
...
+        rs_device* rs_detected_device, std::string typeOfFirmware);
```
mkhansenbot: Please change rs_detected_device to const rs_device& instead of pointer. See this article for explanation: http://www.cplusplus.com/articles/z6vU7k9E/

realsense_camera/src/base_nodelet.cpp
```diff
@@ -312,6 +312,8 @@ namespace realsense_camera
         checkError();
         detected_camera_msg = detected_camera_msg + ", Motion Module FW: " + motion_module_fw;
       }
+
+      checkFirmwareValidation("validatedFirmware", rs_detected_device, "typeOfFirmware");
```
mkhansenbot: If you re-order the arguments and use default arguments in the declaration, you don't have to pass these dummy strings. See this: http://en.cppreference.com/w/cpp/language/default_arguments

mkhansenbot (changes requested):

mdhorn:
Please remove the swap file: 
realsense_camera/include/realsense_camera/.r200_nodelet.h.swp

mdhorn (changes requested):
If we aren't using the checkFirmwareValidation function, can you please remove it from all of the header and Cpp files.

Also, please run roslint and make updates before resubmitting.

realsense_camera/src/base_nodelet.cpp
```diff
@@ -275,34 +276,63 @@ namespace realsense_camera
...
+          {"Intel RealSense F200", F200_CAMERA_FW_VERSION},
+          {"Intel RealSense SR300", SR300_CAMERA_FW_VERSION},
+          {"Intel RealSense ZR300", ZR300_CAMERA_FW_VERSION},
+          {"Intel RealSense ZR300_ADAPTER", ZR300_ADAPTER_FW_VERSION},
+          {"Intel RealSense ZR300_MOTION_MODULE", ZR300_MOTION_MODULE_FW_VERSION}
+    };
+
     for (int i = 0; i < num_of_cameras; i++)
     {
+      std::string detected_camera_msg = " - Detected the following camera:";
```
mdhorn: Does this not result in printing the message for each camera found instead of once for all cameras?

arbrindle:
Here is sample output of when both a ZR300 and a F200 are detected. The ZR300 has unvalidated camera firmware and adapter firmware. 

```
[ INFO] [1484087578.718521071]: /camera/driver - Detected the following camera:
				 - Serial No: 3311803158, USB Port ID: 2-2-2, Name: Intel RealSense ZR300, Camera FW: 2.0.71.28, Adapter FW: 1.29.0.0, Motion Module FW: 1.25.0.0
[ WARN] [1484087578.718588513]: /camera/driver - Detected unvalidated firmware:
				 -3311803158's current camera firmware is 2.0.71.28. Validated camera firmware is 2.0.71.26. 
				 -3311803158's current adapter firmware is 1.29.0.0. Validated adapter firmware is 1.28.0.0. 
[ INFO] [1484087578.718651851]: /camera/driver - Detected the following camera:
				 - Serial No: 042140050404, USB Port ID: 2-1-1, Name: Intel RealSense F200, Camera FW: 2.60.0.0
``

```

mdhorn (approved):
LGTM 👍

mdhorn:
retest this please

realsense_camera/src/base_nodelet.cpp
```diff
@@ -275,34 +276,67 @@ namespace realsense_camera
   {
     // print list of detected cameras
     std::vector<int> camera_type_index;
-    std::string detected_camera_msg = " - Detected the following cameras:";
+
+    // map camera names to validated firmware
+    std::map<std::string, std::string> camera_name_to_validated_firmware =
```
mkhansenbot: Can this map definition be const and moved into constants.h? We've tried to avoid putting any hard-coded constants in the cpp files. Also, we've tried to avoid having any camera type specific code in the base nodelet, to keep it generic to all camera types.

realsense_camera/src/base_nodelet.cpp
```diff
@@ -275,34 +276,67 @@ namespace realsense_camera
...
-            "\n\t\t\t\t- Serial No: " + rs_get_device_serial(rs_detected_device, &rs_error_) +
-            ", USB Port ID: " + rs_get_device_usb_port_id(rs_detected_device, &rs_error_) +
+            "\n\t\t\t\t - Serial No: " + camera_serial_number + ", USB Port ID: " +
+            rs_get_device_usb_port_id(rs_detected_device, &rs_error_) +
             ", Name: " + camera_name +
-            ", Camera FW: " + rs_get_device_firmware_version(rs_detected_device, &rs_error_);
+            ", Camera FW: " + camera_fw;
       checkError();
 
+      if (camera_fw != camera_name_to_validated_firmware[camera_name])
```
mkhansenbot: There's a lot of code duplication here, can the checks of the fw and printing warnings be refactored into a common method that is called for each fw?

mkhansenbot (dismissed):

realsense_camera/src/base_nodelet.cpp
mdhorn: I would not make it constant as we will likely need to change/override these map values if we enable a negative-testing method for QA.

arbrindle: When I put the map declaration/initialization into constants.h, I get the following error:

```
CMakeFiles/realsense_camera_nodelet.dir/src/r200_nodelet.cpp.o:(.bss+0x0): multiple definition of `realsense_camera::camera_name_to_validated_firmware[abi:cxx11]'
CMakeFiles/realsense_camera_nodelet.dir/src/base_nodelet.cpp.o:(.bss+0x0): first defined here
CMakeFiles/realsense_camera_nodelet.dir/src/f200_nodelet.cpp.o:(.bss+0x0): multiple definition of `realsense_camera::camera_name_to_validated_firmware[abi:cxx11]'
CMakeFiles/realsense_camera_nodelet.dir/src/base_nodelet.cpp.o:(.bss+0x0): first defined here
CMakeFiles/realsense_camera_nodelet.dir/src/sr300_nodelet.cpp.o:(.bss+0x0): multiple definition of `realsense_camera::camera_name_to_validated_firmware[abi:cxx11]'
CMakeFiles/realsense_camera_nodelet.dir/src/base_nodelet.cpp.o:(.bss+0x0): first defined here
CMakeFiles/realsense_camera_nodelet.dir/src/zr300_nodelet.cpp.o:(.bss+0x0): multiple definition of `realsense_camera::camera_name_to_validated_firmware[abi:cxx11]'
CMakeFiles/realsense_camera_nodelet.dir/src/base_nodelet.cpp.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
realsense/realsense_camera/CMakeFiles/realsense_camera_nodelet.dir/build.make:465: recipe for target '/home/arbrindl/catkin_ws/devel/lib/librealsense_camera_nodelet.so' failed
make[2]: *** [/home/arbrindl/catkin_ws/devel/lib/librealsense_camera_nodelet.so] Error 1
CMakeFiles/Makefile2:2383: recipe for target 'realsense/realsense_camera/CMakeFiles/realsense_camera_nodelet.dir/all' failed
make[1]: *** [realsense/realsense_camera/CMakeFiles/realsense_camera_nodelet.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2
Invoking "make -j8 -l8" failed
```

I'm not sure why it complains about multiple definitions of the map when I don't use it in anything other than base_nodelet.cpp. I was only able to get rid of these errors when I put the map in base_nodelet.cpp instead of constants.h. Do you have any idea why these errors would pop up when I move the map to constants.h?

ghindman:
On further consideration I still have significant concern about this commit.  

1) We don't have insight into production firmware versions/updates, so we may end up throwing a warning against a factory-burned version of FW, which is not good.
2) We're building a list of specific firmware that we must then maintain.

If we're going to do this then it should start with a minimum version set at the first official production version, and then check for that version or newer.

mdhorn:
@ghindman Each release of realsense_camera is run through QA validation is only ONE firmware version per camera type. This will be the version of firmware that does not generate a warning -- all other versions (older or newer) will generate a warning to the customer. This is intended to communicate that the version of realsense_camera they are using was not validated with that firmware.

Yes, we will have to update the code with new firmware releases, but this is why QA will have a validation test which ensures that there are no warnings during test.

This is only generating a Warning, not an error.

arbrindle:
@mdhorn I know you considered sometimes showing a warning and sometimes showing info; perhaps if the firmware is older, we give a warning, and if the firmware is newer, we mention that in an info stream.

mdhorn (changes requested):
please change the checkFirmwareValidation function to return the warning string instead of passing in a pointer to the warning message string.

Let the calling function handle appending the returned string into one longer warning message.

mdhorn:
retest this please

mdhorn (approved):
LGTM 👍

PR #189, opened 2017-01-12
Increase retry count for rgbd test

nwang2:
During one week observation on CI, rgdb test still failed randomly. It's about a 10% fail rate, not always failed at the same topic, but always in rgbd test. So increase the retry count for rgbd test to improve the stability.


nwang2:
retest this please

nwang2:
retest this please

mdhorn:
As discussed, we will not be making this change as it no longer seem necessary.

PR #191, opened 2017-01-13
Use shared timestamp for SR300/F200 cameras

mdhorn:
Due to camera hardware issues on the SR300 and F200, reverted back
to using a common timestamp updated by the fastest stream. This was the
old behaviour when rs_wait_for_frames was used prior to release 1.7.0.


mkhansenbot (approved):
Looks fine to me. At a future date we should discuss how to refactor the duplicated code from F200 & SR300. For now though I understand why it is done this way. Approved

PR #204, opened 2017-02-15
Add option to link against non-catkin librealsense

rojkov:
In OpenEmbedded setups where both `meta-ros` and `meta-intel-realsense` layers are used it's problematic to avoid two copies of librealsense installed onto an embedded target: one catkit-based installed under `/opt/ros` and one other installed to a standard system location.

This patch adds an option making realsense_camera link against librealsense provided by the host system.

By default the currently existing behavior is preserved.

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -14,8 +14,9 @@ set(CMAKE_EXE_LINKER_FLAGS "-pie -z noexecstack -z relro -z now")
 # Flags shared libraries
 set(CMAKE_SHARED_LINKER_FLAGS "-z noexecstack -z relro -z now ${CMAKE_SHARED_LINKER_FLAGS}")
 
-find_package(catkin REQUIRED COMPONENTS
-  librealsense
+option(USE_SYSTEM_LIBREALSENSE "Build realsense_camera against system librealsense, not the one built with catkin" OFF)
+
+set(REALSENSE_CATKIN_BASED_DEPS
   dynamic_reconfigure
   roscpp
   nodelet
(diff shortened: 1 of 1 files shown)


rojkov:
ping.
The corresponding meta-ros PR  is here https://github.com/bmwcarit/meta-ros/pull/457. And it has been tested already.

mdhorn (changes requested):
Sorry, we were in the middle of a regression which needed to be completed before merging new code.

I also want to try a few test builds to approve possible fails in the ROS Build farm.

The code looks okay to me, but please rebase this branch as there have been minor updates.

rojkov:
* rebased to the current indigo-devel
* made running roslint tasks conditional since they require `CATKIN_ENABLE_TESTING=ON`
* force pushed.

mdhorn (approved):

fmessmer:
Somehow https://github.com/intel-ros/realsense/commit/701f9393bf102c5d130fbe80b434b4060c405c31 lets our Travis CI fail due to 
```
Errors     << realsense_camera:cmake /root/catkin_ws/logs/realsense_camera/build.cmake.000.log

CMake Error at /root/catkin_ws/src/realsense/realsense_camera/CMakeLists.txt:122 (roslint_cpp):

  Unknown CMake command "roslint_cpp".
```
see https://travis-ci.org/ipa320/cob_robots/jobs/212986198#L6578

our travis.yml can be seen here: https://github.com/ipa320/cob_robots/blob/indigo_dev/.travis.yml

---
Submitting an issue seems to be blocked somehow...

mdhorn:
There appears to be some race condition in the change.
`catkin_make_isolated` also has the same fail signature, where `catkin_make_isolated -j1` passes.
I'll be submitting a PR with a minor change to fix this.

PR #221, opened 2017-03-24
Enable configuration of the TF publication rate when using tf_dynamic

severin-lemaignan:
The parameter is called `tf_publication_rate`

Related to review comment on issue #169

### DIFF
--- a/realsense_camera/launch/r200_nodelet_modify_params.launch
+++ b/realsense_camera/launch/r200_nodelet_modify_params.launch
@@ -8,35 +8,37 @@
 
   <!-- These 'arg' tags are just place-holders for passing values from test files.
   The recommended way is to pass the values directly into the 'param' tags. -->
-  <arg name="enable_depth"      default="true" />
-  <arg name="enable_ir"         default="false" />
-  <arg name="enable_ir2"        default="false" />
-  <arg name="enable_color"      default="true" />
-  <arg name="enable_pointcloud" default="false" />
-  <arg name="enable_tf"         default="true" />
-  <arg name="enable_tf_dynamic" default="false" />
-  <arg name="mode"              default="manual" />
-  <arg name="depth_width"       default="640" />
-  <arg name="depth_height"      default="480" />
-  <arg name="color_width"       default="1920" />
-  <arg name="color_height"      default="1080" />
-  <arg name="depth_fps"         default="30" />
-  <arg name="color_fps"         default="30" />
+  <arg name="enable_depth"        default="true" />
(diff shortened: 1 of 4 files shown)


mdhorn:
Add my same comment here for clarity:

The data is not changing, hence the slow loop rate.

The default for transforms is to publish them with a static transform, but due to a ROS bug (and a camera driver bug for F200/SR300) the dynamic transform option was added.
See http://wiki.ros.org/realsense_camera#Transform_Restrictions

severin-lemaignan:
@mdhorn I'm using `enable_tf_dynamic` exactly for the reason described in #169 (which is indeed even more acute for F200/SR300 as we do not have the option to use 2 launch files).

That said, dynamically publishing the camera transforms at 1Hz is essentially useless: TF will be usually unable to compute transforms between the camera frames and other frames in the system as the camera transforms will be too old.

For instance, in the old [tf#static_transform_publisher](http://wiki.ros.org/tf#static_transform_publisher) (ie, before tf2 and the latched `/tf_static` topic), they recommend 10Hz.

realsense_camera/launch/r200_nodelet_modify_params.launch
```diff
@@ -8,20 +8,21 @@
...
-  <arg name="depth_fps"         default="30" />
-  <arg name="color_fps"         default="30" />
+  <arg name="enable_depth"        default="true" />
+  <arg name="enable_ir"           default="false" />
+  <arg name="enable_ir2"          default="false" />
+  <arg name="enable_color"        default="true" />
+  <arg name="enable_pointcloud"   default="false" />
+  <arg name="enable_tf"           default="true" />
+  <arg name="enable_tf_dynamic"   default="false" />
+  <arg name="tf_publication_rate" default="1.0" />
```
mdhorn, line 18: This file is missing setting the parameter with the argument value.
Around line 34, the following should be added:
```
  <param name="$(arg camera)/driver/tf_publication_rate" type="double" value="$(arg tf_publication_rate)" />
```

mdhorn (changes requested):

realsense_camera/launch/r200_nodelet_modify_params.launch
severin-lemaignan: My bad. Fixed.

mdhorn (approved):

mdhorn:
@severin-lemaignan We just started a validation cycle for the release of 1.8.0.
I'll check with our QA team, and if they to adding this change and still keep our scheduled I'll merge this on Monday; else it may be a week before merging.

Issue #205, opened 2017-02-22
Add a script for restarting the USB device


AndyZe:
We are using our RealSense on a mobile robot platform, which means long cables and lots of motion. The cameras often crash, so we found it convenient to write a C program that finds and restarts all R200 cameras without power cycling them. (It might be more convenient if it were a Python script, but I was in a hurry to get this done and found a decent C template online.) Let me know if this is of interest and where to put it and I'll do a PR. Otherwise, hopefully it inspires something similar.

[reset_r200.c.txt](https://github.com/intel-ros/realsense/files/795138/reset_r200.c.txt)

AndyZe:
To test it, compile:
$ cc reset_r200.c -o reset_r200

Make it executable:
$ chmod +x reset_r200

AndyZe:
Adding a line to kill all zombie R200 processes, based on the process name "nodelet."
[reset_r200.c.txt](https://github.com/intel-ros/realsense/files/815434/reset_r200.c.txt)

mattcurfman:
Hi @AndyZe, it is not normal for R200 cameras to crash, but acknowledge that USB3 devices are more sensitive to long cables, etc.  I'd recommend including your reset utility as a .cpp file in the 'src' directory, and adding the required commands to the CMakeLists.txt file to build it as a standalone reset utility.

PR #239, opened 2017-05-08
Adding a program to find and reset all RealSense cameras.

AndyZe:
Fixes Issue: #205 

-Automatically find and reset all types of RealSense camera.

You might want to check that I found all possible device #'s for RealSense cameras:

```
reset_all_devices("0a80");

// SR300's

reset_all_devices("0aa5");

reset_all_devices("0ab3");

// ZR300's

reset_all_devices("0ad0");

reset_all_devices("0ACB");
```

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -141,6 +141,8 @@ if (CATKIN_ENABLE_TESTING)
     )
   add_dependencies(tests_rgbd_topics ${PROJECT_NAME}_generate_messages_cpp)
   add_dependencies(tests_rgbd_topics ${catkin_EXPORTED_TARGETS})
+
+  add_executable(reset_realsense src/reset_realsense.cpp)
 endif()
 
 # Install nodelet library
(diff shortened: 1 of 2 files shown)


mattcurfman (changes requested):
Is there a way to do this without killing all nodelets?  Would typically expect many nodelets in running system, beyond what are servicing camera.  Wouldn't want to kill all of them?

AndyZe:
Here's a new way to do it. It would be easier if you could give the process a unique name, though, like **realsense_nodelet**. I'm not sure how easy that would be.

doronhi:
I believe that the with the `initial_reset` parameter, the automatic reset following the automatic detection of a problem and the orderly ctrl-C exit, this PR is not necessary anymore.
Would you agree?

AndyZe:
I haven't tried initial_reset, but feel free to close the issue if that's true.

joekeo:
I am currently using a T265 camera and when launching the node with ros 3 out of 5 times the T265 node will fail to start.
I have in the launch file the <arg name="initial_reset"       default="true"/> and <arg name="initial_reset"            value="$(arg initial_reset)"/>

And the camera still does not initialise quite often.
Has anyone been able to solve this and have the camera reliably 'turning on'?

amilcarlucas:
Any progress on this ?

AndyZe:
Three years later, some clients are still having trouble with zombie Realsense ROS nodes. This script was based on `rosnode kill ...`, but that doesn't even work anymore to kill the nodes.

Issue #258, opened 2017-07-26
[zr300] frame_id prefixing not correct for ir, ir2, imu and fisheye


fmessmer:
### System Configuration
Please complete Your Configuration detail below. Refer to the BKC at [Configuration](../realsense_camera/README.md#configuration).

| Version               | Your Configuration   |
|:--------------------- |:---------------------|
| Operating System      | Ubuntu 14.04.5               |
| Kernel                | ?-?-?                |
| ROS                   | ROS Indigo                    |
| ROS RealSense         | from source https://github.com/intel-ros/realsense/commit/9211801edeaecac1a7ebfa94c6517fd9d95fbb14              |
| librealsense          | current release                |
| Camera Type-Firmware  | zr300            |

---

### Expected Behavior
all frames broadcasted by realsense_camera driver are prefixed using the `camera` launch file argument

### Actual Behavior
frame prefixing works for `depth` and `rgbd` frames, but does not work for `ir`, `ir2`, `imu` and `fisheye`.
for those they seem to be (hardcoded/always) prefixed with "camera", i.e. the argument is not considered

topic prefixing works correctly for all image topics

### Steps to Reproduce
 - `roslaunch realsense_camera zr300_nodelet_rgdb.launch camera:=oink`

A `rostopic echo /oink/depth/camera_info/header -n1` (rgbd respectively) will show:
```
header:
  seq: 61
  stamp:
    secs: 1501081015
    nsecs: 824442857
  frame_id: oink_depth_optical_frame
```
while `rostopic echo /oink/fisheye/camera_info/header -n1` (ir, ir2, imu respectively) will show:
```
header:
  seq: 0
  stamp:
    secs: 1501081106
    nsecs: 903379461
  frame_id: camera_fisheye_optical_frame
```
the wrong prefixing can also be seen when useing `rosrun tf2_tools view_frames.py`!

[frames.pdf](https://github.com/intel-ros/realsense/files/1177194/frames.pdf)

fmessmer:
fix is proposed in #259 
@dorodnic @icarpis

icarpis:
PR #259 is merged

PR #259, opened 2017-07-27
Added missing zr300 launch file arguments

benmaidel:
Fixes Issue: #258 

- added missing zr300 arguments to launch file


fmessmer (approved):
:+1:

fmessmer:
can this be merged please?

icarpis (commented):
Hi @ipa-bnm, thank you for your helping!
There is more work to be done regarding this change. Currently the new parameters appears only in the Launch file and not referenced/published from the source code. As a result, no data is published under the new messages. for instance, there is no data published under /camera/fisheye/camera_info .
Please take a look how we publish the other ROS messages:  depth, rgb, ir, depth_optical_frame_id and etc.

benmaidel:
Hello @icarpis. I think there is a little misunderstanding. This PR is just adding missing launch file params and topic remaps for the zr300 camera. The driver is already reading this params from the parameter server. Take a look here: https://github.com/intel-ros/realsense/blob/indigo-devel/realsense_camera/src/zr300_nodelet.cpp#L98-L112 The zr300 driver is also already publishing imu and fisheye data to the default topic names. To be able to change the topic names I've introduced these two remappings. Also, if we would like the use the `nodelet_rgbd.launch.xml` together with the zr300 camera we need to pass the frame_id params. Otherwise we will not be able to set the frame_ids for `fisheye, fisheye_optical, ir_optical, ir2_optical, imu and imu_optical`. See this Issue https://github.com/intel-ros/realsense/issues/258

For cameras without imu and/or fisheye (r200 for example) this PR should be fine as well, because other realsense camera will just ignore this remaps and params.

icarpis (approved):
@ipa-bnm, thank you for the clarification.
I ran this PR on ZR300, R200 and SR300 to make sure nothing was broken.

Issue #30, opened 2016-04-21
Registered depth and color


senia907:
| Version | Best Known | Your Configuration |
| :-- | :-- | :-- |
| Operating System | Ubuntu 14.04.4 LTS | Ubuntu 14.04.4 LTS |
| Kernel | 4.4.0-040400-generic | 4.4.0-040400-generic |
| ROS | indigo | indigo |
| librealsense | 0.9.1 | 0.9.1 |
| R200 Firmware | 1.0.72.06 | 1.0.72.06 |
### Expected Behavior

Registered(matched) depth and color image at 640x480.
I'm trying to use rtabmap(mapping package, not so relevant to the problem) in ros, and I want to get a color and a matching depth image at 640x480 resolution.
### Actual Behavior

Using "realsense_r200_nodelet_standalone_manual.launch" (with color&depth width and height set to 640x480), and using the topics 
/camera/color/image_raw
/camera/depth/image_raw
I get the following results:
![both](https://cloud.githubusercontent.com/assets/12409556/14709292/95b2c00c-07d8-11e6-9edd-fa6ec37713e7.png)
![depth](https://cloud.githubusercontent.com/assets/12409556/14709291/95b10ec4-07d8-11e6-8299-6826d50e5f91.png)
![color](https://cloud.githubusercontent.com/assets/12409556/14709293/95b4401c-07d8-11e6-9e47-e464b1113c17.png)

where the color and depth doesn't match.
Using "realsense_r200_rgbd.launch", which should register(match?) the color&depth gives me a 480x360 resolution at "/camera/depth/image_raw". other topics didn't give me an input, or bad(480x360) resolution.
Only "/camera/depth_registered/sw_registered/image_rect" and "#same#/image_rect_raw" gave me results at the desired resolution(640x480) but at some wired format(not as "/camera/depth/image_raw")

trying to change the default resolution at [realsense_camera_nodelet.h](https://github.com/intel-ros/realsense/blob/indigo-devel/camera/src/realsense_camera_nodelet.h) resulted in a flickering(bad) depth image. Couldn't find other direction to try.

I saw in the [limitation](https://github.com/intel-ros/realsense/tree/indigo-devel/camera#Limitations) section you mention that there isn't a hardware based registartion, but the software registration doesn't suite me either, is there a way to fix it? Or something else to try to get a matching depth and color image at 640x480.
### Steps to Reproduce

Mentioned launch files and "rosrun image_view image_view image:=#topic#"

[...]

Issue #233, opened 2017-04-12
Vertical misalignment of registered(+rectified) depth image compared to rectified RGB image


mindThomas:
### System Configuration
| Version               | Your Configuration   |
|:--------------------- |:---------------------|
| Operating System      | Yocto               |
| Kernel                | 4.4.3-yocto-standard                |
| ROS                   | indigo                    |
| ROS RealSense         | 1.7.2                |
| librealsense          | 1.12.1   (system lib)                |
| Camera Type-Firmware  | R200-1.0.71.06            |

### Expected Behavior
I would expect /camera/rgb/image_rect_color and /camera/depth_registered/sw_registered/image_rect_raw to be aligned.

### Actual Behavior
Please download the pictures below and try to match them (with transparancy).
![rgb-image_rect_color](https://cloud.githubusercontent.com/assets/1436465/24958742/b1d8931c-1f90-11e7-8efc-028a585f8fe5.png)
![depth_registered-image_rect_raw](https://cloud.githubusercontent.com/assets/1436465/24958745/b4678e12-1f90-11e7-891e-28de04278b5e.png)

Doing a transparent matching of the above test images results in the following horizontally mismatched picture.
![horizontal-mismatch](https://cloud.githubusercontent.com/assets/1436465/24958974/7fb5ff5e-1f91-11e7-8d40-d0b05635dfa3.png)
Notice how the registered depth image has a 9 pixel tall black area in the top and a 20 pixel black bar in the right side. If the registered depth image is just moved 9 pixels upwards the color and depth image seem to match much better.
![improved-matching](https://cloud.githubusercontent.com/assets/1436465/24959025/b4f284ee-1f91-11e7-8c36-e063b95a40ae.png)

### Steps to Reproduce
Launch the rgbd nodelet.
`roslaunch realsense_camera r200_nodelet_rgbd.launch`
Capture images with rqt or similar

### Further Questions
Can someone elaborate on the difference between all the topics being published by the RGBD nodelet?
```
/camera/depth/camera_info
/camera/depth/disparity
/camera/depth/image
/camera/depth/image_raw
/camera/depth/image_rect
/camera/depth/image_rect_raw
/camera/depth/points
/camera/depth_registered/disparity
/camera/depth_registered/points
/camera/depth_registered/sw_registered/camera_info
/camera/depth_registered/sw_registered/image_rect
/camera/depth_registered/sw_registered/image_rect_raw
/camera/ir/camera_info
/camera/ir/image_raw
[...]

Issue #201, opened 2017-01-27
ROS published registered point cloud has significantly higher registration error than ccp-example cloud


skohlbr:
### System Configuration

| Version               | Your Configuration   |
|:--------------------- |:---------------------| 
| Operating System      | "Ubuntu 16.04.1 LTS"          |
| Kernel                | 4.4.0-59-generic              |
| ROS                   |  kinetic      |
| ROS RealSense         | 1.6.1               |
| librealsense          | 1.11.1                |
| Camera Type-Firmware  | Name: Intel RealSense R200, Camera FW: 1.0.72.06            |

---

### Expected Behavior

The ROS published point cloud on the `/camera/depth_registered/points` topic should have comparable registration error to the one that is shown in the `cpp-pointcloud` example

### Actual Behavior

The cloud published on the `/camera/depth_registered/points` has visibly higher registration error than the one provided by the example code. See provided example images below. The ones with blue background are from the cpp-pointcloud example, the ones with grey background from rviz. Fingers are colored correctly for the example, while they are mostly white when holding them horizontal for the ROS cloud. This indicates a registration offset along the vertical axis between depth and rgb image.

![cpp_pointcloud_fingers_horizontal](https://cloud.githubusercontent.com/assets/205963/22368768/9538021c-e488-11e6-9d58-e516ece9de03.png)
![cpp_pointcloud_fingers_vertical](https://cloud.githubusercontent.com/assets/205963/22368767/9537c7b6-e488-11e6-9795-e37ead959b72.png)

![ros_pointcloud_fingers_horizontal](https://cloud.githubusercontent.com/assets/205963/22368777/a8ca3570-e488-11e6-9310-393161ab86f8.png)
![ros_pointcloud_fingers_vertical](https://cloud.githubusercontent.com/assets/205963/22368778/a8cc23ee-e488-11e6-934e-045a2084708b.png)

#### ROS cloud:
Install realsense camera from .debs
<pre>
sudo apt-get install ros-kinetic-realsense-camera
</pre>
Run launch file
<pre>
roslaunch realsense_camera r200_nodelet_rgbd.launch 
</pre>
Look at point cloud in rviz

#### cpp-poincloud example cloud

Clone librealsense repo and build with example as per https://github.com/IntelRealSense/librealsense/blob/master/doc/installation.md

Navigate to `build/devel/lib/librealsense` and start `./cpp-pointcloud`

skohlbr:
Looks like #184 might be the culprit here.

[...]

Issue #184, opened 2017-01-07
The rotation between Depth and color image coordinate system


robolabsjtu:
### System Configuration
Please complete Your Configuration detail below. Refer to the BKC at [Configuration](../realsense_camera/README.md#configuration).

| Version               | Your Configuration   |
|:--------------------- |:---------------------|
| Operating System      | Ubuntu 16.04.1 LTS                |
| Kernel                | 4.4.0-31-generic               |
| ROS                   | kinetic                  |
| ROS RealSense         | 1.6.0               |
| librealsense          | 1.12              |
| Camera Type-Firmware  | Serial No: 2391016074, USB Port ID: 2-4, Name: Intel RealSense R200, Firmware: 1.0.72.06          |

---

### Expected Behavior

Last time I got a small registered error in /camera/depth_registered/points
Issue Link:  https://github.com/intel-ros/realsense/issues/115

I found maybe the problem is about the  rotation between Depth and Color image.

In the base_nodelet.cpp, the  rotation between Depth and Color image coordinate system was set to identity matrix. as follows
///////
    b2d_msg.header.stamp = transform_ts_;
    b2d_msg.header.frame_id = base_frame_id_;
    b2d_msg.child_frame_id = frame_id_[RS_STREAM_DEPTH];
    b2d_msg.transform.translation.x =  color2depth_extrinsic_.translation[2];
    b2d_msg.transform.translation.y = -color2depth_extrinsic_.translation[0];
    b2d_msg.transform.translation.z = -color2depth_extrinsic_.translation[1];
    b2d_msg.transform.rotation.x = 0;
    b2d_msg.transform.rotation.y = 0;
    b2d_msg.transform.rotation.z = 0;
    b2d_msg.transform.rotation.w = 1;
static_tf_broadcaster_.sendTransform(b2d_msg);
///////
However in my r200 device the really rotation matrix is not a identity matrix. as follows
//////
  depth_to_rgb(0,0) = 0.999996;
  depth_to_rgb(1,0) = 0.002805;
  depth_to_rgb(2,0) = -0.000393;
  depth_to_rgb(0,1) = -0.002796;
  depth_to_rgb(1,1) = 0.999757;
  depth_to_rgb(2,1) = 0.021851;
  depth_to_rgb(0,2) = 0.000454;
  depth_to_rgb(1,2) = -0.021850;
  depth_to_rgb(2,2) = 0.999761;
//////  
[...]

PR #261, opened 2017-08-12
Fix R200 color cloud misalignment by using rectified color calibration

skohlbr:
This PR fixes the cloud/color camera alignment issues reported in issues stated below.
 
Fixes Issue: #30,  #184, #201, #233

- Use rectified camera stream calibration instead of unrectified. It has to be noted that even just performing this change (and leaving color image retrieval as-is) vastly improves point cloud alignment (reducing error from ~9 pixels to maybe ~1 pixel in most image regions).
- Use the rectified color stream consistently for the R200. As a lot of the stream handling is done in base_nodelet for all camera types, performing the necessary changes without breaking other cameras (SR300 etc.) is not trivial. As the calibrated color image is published now, it has been remapped in the [nodelet_rgbd.launch.xml](https://github.com/skohlbr/realsense/blob/8d008af73bc2a5d25263ba11bdf61b5ce729370b/realsense_camera/launch/includes/nodelet_rgbd.launch.xml#L67) file. As a side effect, the rectification plugin loaded as part of the rgbd launch setup now throws ROS_WARNs about missing the image_color topic.

Looking for feedback by the maintainers, as this PR obviously cannot be merged right now without breaking other Realsense camera (SR300 etc.) use with the driver. See #201 for details of how this PR came about (screenshots etc.).


skohlbr:
Updated PR and initial description.

dorodnic:
Hi @skohlbr 
Thank you for taking the time to dig into this! We will test and see how to integrate with other devices.

compiaffe:
@dorodnic @icarpis Any news on this one?

icarpis (approved):

fmessmer:
@skohlbr @icarpis 
To be straight, I'm non perception expert...I just experienced that https://github.com/intel-ros/realsense/pull/261/files#diff-bc72946ee2b0a49ae6c992ef05d4f704R75 breaks the perception topic pipeline for our zr300 cameras...
As stated in the PR description, this has been warned about...however, I'm missing a solution how users for other cameras could guarantee their previous behavior?

Does the `else if (stream_index == RS_STREAM_COLOR)` part affect the zr300, too?
Should/Could the `rgb_processing` be turned off for zr300, too?
If the remap is kind of r200-specific, why was it not done in the specific launch file?
Could the remap be toggled with an `unless` argument - which defaults to false?

PR #289, opened 2018-01-02
Added dynamic params, multiple cameras and JSON file features

icarpis:
- Support working with multiple cameras
- Added an ability to load Advanced Mode JSON files
- Added dynamic params for SR300, RS415 and RS435


shaun-edwards:
Thanks for this PR!....exactly what we were looking for.

PR #300, opened 2018-01-28
Disable ZR300 rgb_processing

icarpis:
Disable ZR300 rgb_processing because a rectified color stream is already provided by [Librealsense](https://github.com/intel-ros/realsense/blob/dadd3f95405ca5af2694488681231fbaf84de419/realsense_camera/src/sync_nodelet.cpp#L193).


ev-mp (approved):
Looks good

dorodnic (approved):

PR #324, opened 2018-02-27
Renaming ROS package from realsense_ros_camera to realsense2_camera 

icarpis:


icarpis:
@sharronliu

sharronliu:
Noted, thanks!

Issue #335, opened 2018-03-05
float comparison in BaseNodelet::publishPCTopic() causes out of bound access and segmentation fault


lsolanka:
With my R200, for some reason this code:
```c++
 if (color_pixel[1] < 0.0f || color_pixel[1] > image_color.rows
     || color_pixel[0] < 0.0f || color_pixel[0] > image_color.cols)
{
   // For out of bounds color data, default to a shade of blue in order to visually distinguish holes.
   // This color value is same as the librealsense out of bounds color value.
   *iter_r = static_cast<uint8_t>(96);
   *iter_g = static_cast<uint8_t>(157);
   *iter_b = static_cast<uint8_t>(198);
}
else
{
  int i = static_cast<int>(color_pixel[0]);
  int j = static_cast<int>(color_pixel[1]);

  *iter_r = static_cast<uint8_t>(color_data[i * 3 + j * image_color.cols * 3]);
  *iter_g = static_cast<uint8_t>(color_data[i * 3 + j * image_color.cols * 3 + 1]);
  *iter_b = static_cast<uint8_t>(color_data[i * 3 + j * image_color.cols * 3 + 2]);
}


```
Causes a segmentation fault because from time to time color_pixel[1] == image_color.rows or color_pixel[0] == image_color.cols.

The image access in the else clause is then out of bounds by 1 pixel, and librealsense_camera_nodelet.so segfaults as a result.

lsolanka:
Hi,
I noticed that you closed this issue without merging the pull request. Is there any issue with the patch? Is it not relevant?

PR #336, opened 2018-03-05
fix off-by-one error in comparison of color_pixel and image_color sizes

lsolanka:
This fixes an issue in #335.

The mapping of points to pixels can fall exactly on the boundary of the image size. In that case there will be out of bounds access to the color_data array which results in segmentation fault.

This patch makes sure this does not happen.

### DIFF
--- a/realsense_camera/src/base_nodelet.cpp
+++ b/realsense_camera/src/base_nodelet.cpp
@@ -1012,8 +1012,8 @@ namespace realsense_camera
             rs_transform_point_to_point(color_point, &z_extrinsic, depth_point);
             rs_project_point_to_pixel(color_pixel, &color_intrinsic, color_point);
 
-            if (color_pixel[1] < 0.0f || color_pixel[1] > image_color.rows
-                || color_pixel[0] < 0.0f || color_pixel[0] > image_color.cols)
+            if (color_pixel[1] < 0.0f || color_pixel[1] >= image_color.rows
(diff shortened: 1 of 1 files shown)


doronhi (approved):
looks good.

doronhi:
@lsolanka Very important. I'll be happy if you could make the same bug fix in Realsense2 Ros wrapper (branch development). Please let me know if you plan to do it or if you prefer me to put it in.

lsolanka:
> @lsolanka Very important. I'll be happy if you could make the same bug fix in Realsense2 Ros wrapper (branch development). Please let me know if you plan to do it or if you prefer me to put it in.

Unfortunately I am swamped with some other projects so I don't think I'll be able to prepare the fix. More so that the realsense2 doesn't support the camera I have so I wouldn't be able to test it. Might be best if you prepare the fix yourself.

doronhi:
> @lsolanka Very important. I'll be happy if you could make the same bug fix in Realsense2 Ros wrapper (branch development). Please let me know if you plan to do it or if you prefer me to put it in.
> 
> Unfortunately I am swamped with some other projects so I don't think I'll be able to prepare the fix. More so that the realsense2 doesn't support the camera I have so I wouldn't be able to test it. Might be best if you prepare the fix yourself.

Absolutely. I'll put it in there. Thanks for noticing and sorry for our very late response.

PR #352, opened 2018-03-29
Feature/diagnostics

reinzor:
Diagnostics for frequency of the image publishers.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -303,7 +303,8 @@ void BaseRealSenseNode::setupPublishers()
             image_raw << _stream_name[stream] << "/image_" << ((rectified_image)?"rect_":"") << "raw";
             camera_info << _stream_name[stream] << "/camera_info";
 
-            _image_publishers[stream] = image_transport.advertise(image_raw.str(), 1);
+            std::shared_ptr<FrequencyDiagnostics> frequency_diagnostics(new FrequencyDiagnostics(_fps[stream], _stream_name[stream], _serial_no));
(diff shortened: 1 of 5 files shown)


icarpis:
@reinzor, thank you for this!
I tested it and it looks very good.
There is one little change that need to be done. I will write it in the review summary.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -312,7 +313,8 @@ void BaseRealSenseNode::setupPublishers()
                 aligned_image_raw << "aligned_depth_to_" << _stream_name[stream] << "/image_raw";
                 aligned_camera_info << "aligned_depth_to_" << _stream_name[stream] << "/camera_info";
 
-                _depth_aligned_image_publishers[stream] = image_transport.advertise(aligned_image_raw.str(), 1);
+                std::shared_ptr<FrequencyDiagnostics> frequency_diagnostics(new FrequencyDiagnostics(_fps[stream], _stream_name[stream], _serial_no));
```
icarpis: Please add the prefix of the aligned frames to the stream name and pass it to `frequency_diagnostics` ctor.
```
std::string aligned_stream_name = "aligned_depth_to_" + _stream_name[stream];
std::shared_ptr<FrequencyDiagnostics> frequency_diagnostics(new FrequencyDiagnostics(_fps[stream], aligned_stream_name, _serial_no));
```

reinzor: Thanks, you are right! Just updated the PR.

icarpis (approved):

PR #362, opened 2018-04-06
Added /depth/points topic+publisher for publishing an xyz PointCloud2

josephduchesne:
Added /depth/points topic, publisher, to allow for publishing an xyz PointCloud2 in addition to the existing /depth/color/points topic.

### CPU load comparison
hardware: 1x Intel D435 camera, i7-4790K 4.0Ghz
config: 6fps depth, 15fps color. 640x480 depth, 1280x720 color
1 subscriber to depth/points: 14-15% 1 core cpu usage, 29.73MB/s BW on rostopic
1 subscriber to depth/color/points: 21-22% 1 core cpu usage, 59.37MB/s BW on rostopic

This also allows for a PointCloud2 topic to be published with the color camera disabled:
config: 6fps depth, 15fps color. 640x480 depth, color disabled
1 subscriber to depth/points: 11-12% 1 core cpu usage, 29.73MB/s BW on rostopic

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -321,7 +321,8 @@ void BaseRealSenseNode::setupPublishers()
 
             if (stream == DEPTH && _pointcloud)
             {
-                _pointcloud_publisher = _node_handle.advertise<sensor_msgs::PointCloud2>("depth/color/points", 1);
+                _pointcloud_xyz_publisher = _node_handle.advertise<sensor_msgs::PointCloud2>("depth/points", 1);
+                _pointcloud_xyzrgb_publisher = _node_handle.advertise<sensor_msgs::PointCloud2>("depth/color/points", 1);
             }
         }
     }
@@ -616,10 +617,15 @@ void BaseRealSenseNode::setupStreams()
                                  _encoding);
                 }
 
-                if(_pointcloud && (0 != _pointcloud_publisher.getNumSubscribers()))
+                if(_pointcloud && (0 != _pointcloud_xyzrgb_publisher.getNumSubscribers()))
                 {
-                    ROS_DEBUG("publishPCTopic(...)");
+                    ROS_DEBUG("publishRgbToDepthPCTopic(...)");
                     publishRgbToDepthPCTopic(t, is_frame_arrived);
+                }
+                if(_pointcloud && (0 != _pointcloud_xyz_publisher.getNumSubscribers()))
+                {
+                    ROS_DEBUG("publishDepthPCTopic(...)");
+                    publishDepthPCTopic(t, is_frame_arrived);
                 }
             }
             catch(const std::exception& ex)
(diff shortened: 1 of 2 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -1048,6 +1054,75 @@ void BaseRealSenseNode::publishStaticTransforms()
...
+    // Fill the PointCloud2 fields
+    for (int y = 0; y < depth_intrinsics.height; ++y)
+    {
+        for (int x = 0; x < depth_intrinsics.width; ++x)
+        {
+            scaled_depth = static_cast<float>(*image_depth16) * _depth_scale_meters;
+            float depth_pixel[2] = {static_cast<float>(x), static_cast<float>(y)};
+            rs2_deproject_pixel_to_point(depth_point, &depth_intrinsics, depth_pixel, scaled_depth);
+
+            if (depth_point[2] <= 0.f || depth_point[2] > 5.f)
```
dorodnic: I would avoid cropping the depth values, since the same wrapper is likely to be used with cameras of larger baseline. Perhaps this can be configurable?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -616,11 +617,16 @@ void BaseRealSenseNode::setupStreams()
                                  _encoding);
                 }
 
-                if(_pointcloud && (0 != _pointcloud_publisher.getNumSubscribers()))
+                if(_pointcloud && (0 != _pointcloud_xyzrgb_publisher.getNumSubscribers()))
```
icarpis, line 620: I suggest adding another `_pointcloud` variable for the Depth Point Cloud or changing the existing parameter to `_rgbd_pointcloud` and connect it to a new [ROS parameter](https://github.com/intel-ros/realsense/blob/91e10b7e051b46ea6edebc56215f483a3b78a0f5/realsense2_camera/src/base_realsense_node.cpp#L104) in order to enable Depth PC and RGBD PC separately.

In addition, it would be nice to get another [Launch file](https://github.com/intel-ros/realsense/blob/development/realsense2_camera/launch/demo_pointcloud.launch) for the Depth PC.

josephduchesne: Is there a good reason to even have the _pointcloud flag? Now that both topics publish only when subscribed to, I can't think of a reason to enable/disable them using rosparam (since AFIK that's a static config).

icarpis: Seperating the topics by rosparams prevents advertising and allocating other unwanted topics and ROS resources.

baumanta:
Is this PR still active? I would also have a great interest in getting an xyz pointcloud. Right now I have a work around using a launch file which also starts the depth_image_proc node. But it would be nice to have a clean upstream solution

doronhi:
Is the following running method answers to the demand in this PR?
`roslaunch realsense2_camera rs_camera.launch filters:=pointcloud pointcloud_texture_stream:=RS2_STREAM_ANY`

PR #367, opened 2018-04-19
fix the aligned depth frame unit conversion issue

AlanBB277:
Greeting maintainers,

Thanks for maintaining this repo. It really helps my project in a big way.

The thing is, the readings of the aligned depth frames on SR300 are incorrect currently, so I've sent a PR to fix it.

The issue was caused by converting the depth frame unit byte by byte, but the frame is encoded in 16UC1.

Since there is no need to do unit conversion on  DS400 series(unit = 0.001), so this issue only appears on SR300 camera(unit = 0.000124987).

This PR has been verified on my side with SR300. If you agree with this modification, please help me to verify it on DS400 series.

If there is any suggestion for this PR, please feel free to advise me.

Thanks again! Hope this helps :)

Alan

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -372,8 +372,6 @@ void BaseRealSenseNode::alignFrame(const rs2_intrinsics& from_intrin,
 {
     static const auto meter_to_mm = 0.001f;
     uint8_t* p_out_frame = out_vec.data();
-    auto from_vid_frame = from_image.as<rs2::video_frame>();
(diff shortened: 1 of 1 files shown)


dorodnic:
Hi @AlanBB277 
Thank you for the pull-request! 
This looks correct to me, but it might take a bit more time to properly test it and publish into the next release.

oka1125:
Hi @AlanBB277 
I found the same issue with SR300 camera.
Thank you for the pull-request.

lz89:
I tested with D435 and it seems working fine. It also fix the issue that if the param-depthunits is changed from 1000 to 100, the reconstructed results would be wrong.

PR #376, opened 2018-04-27
remove librealsense2 depend as unused

yechun1:
Ref to commit-id (785ee8) Align ROS wrapper to the new Librealsense API.
librealsense2 depend should be removed, it is not required and no librealsense2 ros package exist now.
cmake checking will fail for some builds.

Signed-off-by: Chris Ye <chris.ye@intel.com>

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -68,7 +68,7 @@ generate_dynamic_reconfigure_options(
 # RealSense ROS Node
 catkin_package(
     LIBRARIES ${PROJECT_NAME}
-    CATKIN_DEPENDS message_runtime roscpp sensor_msgs std_msgs librealsense2
+    CATKIN_DEPENDS message_runtime roscpp sensor_msgs std_msgs
     nodelet
     cv_bridge
     image_transport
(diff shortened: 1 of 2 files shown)


yechun1:
@icarpis would you please review this patch? 
when we build ros1_bridge it met errors while dependence checking for realsense package. The dependence librealsense2 ros package is not required and should be removed. The dependent libraries should have already be aligned to the new system native librealsense.

doronhi:
@yechun1 
Can you describe the compilation process you took to provoke this error? I assume you followed a different path than the one described under "Installation Instructions": https://github.com/intel-ros/realsense

yechun1:
@doronhi 
Checking with the code(CMakeLists.txt and package.xml),  "librealsense2" is redundant, it is not used anymore after "commit-id (785ee8) Align ROS wrapper to the new Librealsense API." It's better for us to remove "librealsense2" depend, to avoid building issues.

On my work, it will build fail while build ros1_bridge:
1. build ros1 with realsense, (succeed)
2. source ros1 setup.sh (succeed)
3. build ros2 (ros1_bridge) (build fail)

SteveMacenski:
Pinging back here. Mostly we just need librealsense to be librealsense and not librealsense2 since there is no rosdep definition for the package.xml for librealsense2 which causes alot of havoc.

doronhi:
Merged under PR #440 together with PR #380 by @rhaschke

PR #377, opened 2018-04-27
Make generated headers includable by other packages

comkieffer:
Other packages may need to use headers generated by this package. For
example, to use the dynamic_reconfigure::Client to modify dynamic
reconfigure parameters exposed by the realsense camera manager we need
the `rsXXX_paramsConfig.h` files.

This change updates the `catkin_package` call to better specify exported
libraris and header file requirements as well as adding an install rule
for the headers.

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -68,11 +68,13 @@ generate_dynamic_reconfigure_options(
 # RealSense ROS Node
 catkin_package(
     LIBRARIES ${PROJECT_NAME}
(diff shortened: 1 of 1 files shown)


doronhi:
Hi @comkieffer
Somehow I managed to neglect this PR of yours. No excuses.
The code was changed a lot since this PR was submitted. `dynamic_reconfigure` is no longer in use and no header files are generated by it. There are 2 header files, Extrinsics.h, IMUInfo.h, that are being generated but I see they are already being copied to install/include/realsense2_camera.
So, it seems to me that there is no longer a need for the content in this PR.
Can you confirm and close? (Sorry it took 3 years...)

PR #383, opened 2018-05-22
Fix coordinate system transforms so that the pointcloud aligns with camera view

mikolajz:
Currently, the coordinate systems are off, which one can observe by doing:
- launch the RealSense node.
- open rviz.
- in rviz, add the PointCloud2 and a Camera for color/image_raw. If testing with objects close enough, reduce the "Size" in PointCloud2 or change "Style" to point.
- In the Camera window, we will see the pointcloud from the camera viewpoint. Since point colors are computed from color/image_row, the images should align up to floating point rounding errors. Before these patches, they clearly don't align. After them, they align.

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -87,7 +87,7 @@ namespace realsense2_camera
         void setupPublishers();
         void setupStreams();
         void updateStreamCalibData(const rs2::video_stream_profile& video_profile);
-        Eigen::Quaternionf rotationMatrixToQuaternion(const float rotation[3]) const;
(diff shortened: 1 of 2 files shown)


dorodnic:
Hi @mikolajz 
Sorry for the delay. We will do some tests and merge it as soon as possible.

mikolajz:
No problem, thanks for looking into it.

On Tue, Jun 26, 2018 at 8:17 PM, Sergey Dorodnicov <notifications@github.com

doronhi:
I check and found that it does make things much better.
Image before:
before_fix_crop

Image After fix:
fixed_crop

Issue #318, opened 2018-02-22
Auto-Exposure Not Consistently Working at System Init on D435


dpipemazo:
Running v2.0.2 of this package with v2.10.0 of `librealsense` and camera firmware `5.9.2.0`. Ubuntu 16.04 with 4.10.0-28 kernel.

I noticed quite poor depth performance when viewing the Point Cloud in RVIZ after power cycling the D435 camera but no issue when using `realsense-viewer`. This caused the data in RVIZ to appear splotchy and inconsistent.

I then launched `rosrun rqt_reconfigure rqt_reconfigure` and noticed that the auto-exposure for the depth stream (first checkbox at the top of the `realsense` config menu) was indeed turned on (as indicated by the box being checked). However, unchecking this box and then re-checking it consistently fixed the issue indicating that auto-exposure at nodelet boot is not working. This might be an issue on the `librealsense` side, but I'm not too sure since it looked OK in `realsense-viewer`.

Mitchell-Hampton:
I am experiencing the same thing.

icarpis:
@Mitchell-Hampton, @dpipemazo, please post the output of the RealSense ROS node.

icarpis:
I managed to reproduce this issue and it looks like a firmware bug.
As a workaround you can physically replug the camera or call to [Hardware Reset](https://github.com/IntelRealSense/librealsense/blob/51e766389f75731be6d37a4f435b5c6b172609bc/tools/data-collect/rs-data-collect.cpp#L249) method.

dpipemazo:
@icarpis Thanks for looking into it! Sorry I didn't post the output of the node; got pulled away from working on it for a few days. Are you aware of a firmware release schedule or when we might be abe to expect an update?

ammarhusain:
I just ran into this issue as well. The auto exposure works well after toggling it off & on in `rqt_reconfigure`. I am going to try calling `hardware_reset` and see if that helps. 

@icarpis I would be really interested in a firmware fix for this as well.

agrunnet:
Can you confirm that you are only seeing this on Linux AND ROS? I dont see this issue on Windows, but maybe I am not doing the right thing to reproduce. We need to figure out whether it is FW or SW.

reinzor:
We have experienced the same issue, this is the script we use to solve it now (as a workaround):
```
#!/bin/bash
[...]

PR #392, opened 2018-06-07
Workaround for autoexposure not working on startup

lorenwel:
Workaround for #318, implementing changes as proposed there.


karyuv:
when is this going to merged with the default?. I really need this fix.

doronhi:
Sorry for the very long delay.
The usual behavior of the camera is such that when you change the exposure manually it automatically set auto_exposure to off (even if it is not reflected in the rqt_reconfigure yet, which is another issue...)
That's why I had a problem accepting your PR.
I add a fix in my last commit and it will be available in the upcoming release, hopefully tomorrow..
Thank you so much for bringing this issue on.

Issue #400, opened 2018-06-26
URDF and meshes for D435?


Affonso-Gui:
Hello,

Are we getting URDF and mesh files for D415 and D435 camera under `realsense2_camera/`?

The tf links provided by `_r430.urdf.xacro` do not really fit the D435 model, and I am also having some work trying to convert the official [STEP files](https://www.intel.com/content/www/us/en/support/articles/000026841/emerging-technologies/intel-realsense-technology.html) into collada using free software.

Am doing somehow with custom files, but would be glad to have some official ones.

Best regards.

PR #404, opened 2018-07-03
Add mesh and urdf for D435

Affonso-Gui:
Collada file is the conversion of the [official STEP file](https://www.intel.com/content/www/us/en/support/articles/000026841/emerging-technologies/intel-realsense-technology.html).
Other files use the same template as `r430` and `r415`, with measures from the data sheet.

Can be tested with: `roslaunch realsense2_camera view_d435_model.launch`

Solves #400.

### DIFF
--- a/realsense2_camera/launch/view_d435_model.launch
+++ b/realsense2_camera/launch/view_d435_model.launch
@@ -1,0 +1,8 @@
+<launch>
+    <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find realsense2_camera)/urdf/test_d435_camera.urdf.xacro'" />
+    <node name="robot_state_publisher" pkg="robot_state_publisher" type="state_publisher" />
+
+    <arg name="gui" default="True" />
+    <param name="use_gui" value="$(arg gui)"/>
+    <node name="rviz" pkg="rviz" type="rviz" args="-d $(find realsense2_camera)/rviz/urdf.rviz" required="true" />
+</launch>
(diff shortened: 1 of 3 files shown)


mattcurf:
Would love to see this merged 👍

felixvd:
Works great, tested it in our simulation.

d435_2
d435_1
(The material doesn't appear in Gazebo)

Would propose moving all the properties and material definitions inside the macro, because they can pollute the global namespace otherwise. I would also add a "prefix" parameter so that multiple cameras can be used, but that goes for the other cameras in this package as well and is a separate issue.

dorodnic:
@Affonso-Gui this is great, thank you very much.
Will be merged soon, we might look into implementing the suggestion from @felixvd before merging.

waiyc:
Hi, thanks for adding the d435 model into gazebo ! However, is there any way that I can publish the image data captured from the simulated camera in gazebo?

The closest reference code for this purpose that I manage to find is [R200](https://github.com/SyrianSpock/realsense_gazebo_plugin/blob/master/urdf/realsense-RS200.macro.xacro)

The part that confused me most is the Gazebo reference parameters. 
Is the parameters used by D435 in librealsense2.so same as the parameters used by R200 in [librealsense_gazebo_plugin.so parameters](https://github.com/SyrianSpock/realsense_gazebo_plugin/blob/043beada1e89a54e431b33c1ba819bf485d09dcd/urdf/realsense-RS200.macro.xacro#L79).

felixvd:
That is outside the scope of this issue, but we also use [the same repo](https://github.com/SyrianSpock/realsense_gazebo_plugin) for our gazebo simulation, so we would appreciate an update with newer realsense models. The camera parameters of the R200 are absolutely not the same as for the D435.

PR #405, opened 2018-07-08
Fixing the length of an array argument in rotationMatrixToQuaternion

rsinnet:
The incoming argument for the coefficients of the rotation matrix was set to a dimension three float array but nine coefficients are used in the function.


doronhi (approved):
True. Thanks.

PR #425, opened 2018-07-24
Overwrite all frame ids 

cagbal:
Issue: 
We don't want to use camera_frame_x convention, because it causes lots of problems with multiple cameras and it reduces the flexibility. 

Solution: 
Adding a camera argument and replacing "camera" with camera argument.


doronhi:
I merged PR#426. It seems to address this issue as well.

cagbal:
Thanks!

PR #426, opened 2018-07-24
Making camera namespace configurable -- necessesity to launch multiple cameras

SteveMacenski:
This is needed when you want to launch multiple cameras with multiple launch files which is a common usecase.

This allows to specify the serial and camera ID for the serial to bringup a number of cameras.


SteveMacenski:
PR https://github.com/intel-ros/realsense/pull/425 also adds to this effect and does this for TF needs which is also really necessary.

PR #435, opened 2018-08-06
missing rosdep key

fmessmer:
cannot execute `setup_workspace` because `rosdep install` fails...


SteveMacenski:
its missing keys, but you should switch to `librealsense` not `librealsense2`

fmessmer:
never mind, I wanted to push to our own fork...

PR #438, opened 2018-08-12
line up <group ns> parameter in all launch files.

doronhi:
In addressing Pull requests #425, #426 a parameter <camera> was added to several launch files to describe the namespace. That is, without realizing that a similar parameter named <namespace> was already introduced in rs_rgbd.launch.


SteveMacenski:
Hi,

For my PR I did recognize that it existed in rgbd launch per the standards that openni has for the rgbd launch infrastructure, which intel doesnt use so that launch file naming is a bit of a misnomer

However, the intel realsense ros has always used camera as the namespace parameter even back to the r200's and this PR may break the expected interface some of us are used to

PR #439, opened 2018-08-12
namespace argument renamed "camera".

doronhi:
Agreed, in reference to comment made by @SteveMacenski in PR #438


doronhi:
Agreed, in reference to comment made by @SteveMacenski in PR #438

PR #440, opened 2018-08-12
merge PR regarding CMakefile and package.xml

doronhi:
merge PR #380 by @rhaschke
and PR #376 by @yechun1


SteveMacenski:
Hi,

`librealsense2` has problems because there is no rosdep definition for it, its still under `librealsense`. Deleting the dependency altogether will cause issues with isolated builds since its still required. What this should be is changed from `librealsense2` -> `librealsense`.

PR #455, opened 2018-08-30
PR for hardware_reset during initialization

yycho0108:
Merging [icarpis's fix](https://github.com/icarpis/realsense/tree/reset_dev) as mentioned in https://github.com/intel-ros/realsense/issues/314#issuecomment-365196850 so that it works with the current development branch.
Unless the firmware issue will be fundamentally solved, this workaround makes the whole pipeline significantly more robust for the time being.

Tested with `catkin build` on Ubuntu 16.04 / ROS Kinetic Kame with `librealsense 2.16`.

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(realsense2_camera)
+add_compile_options(-std=c++11)
 
 option(BUILD_WITH_OPENMP "Use OpenMP" OFF)
 
--- a/realsense2_camera/include/realsense_node_factory.h
+++ b/realsense2_camera/include/realsense_node_factory.h
@@ -26,6 +26,9 @@
 #include <eigen3/Eigen/Geometry>
 #include <fstream>
 
+#include <mutex>
+#include <condition_variable>
+
 
 namespace realsense2_camera
 {
--- a/realsense2_camera/src/realsense_node_factory.cpp
+++ b/realsense2_camera/src/realsense_node_factory.cpp
@@ -109,42 +109,6 @@ void RealSenseNodeFactory::onInit()
             auto _device = pipe->get_active_profile().get_device();
             _realSenseNode = std::unique_ptr<BaseRealSenseNode>(new BaseRealSenseNode(nh, privateNh, _device, serial_no));
         }else{
-
-            auto list = _ctx.query_devices();
-            if (0 == list.size())
-            {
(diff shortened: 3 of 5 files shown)


rdaems:
Hi yycho0108

I've been using your fork to fix this problem, thanks. But now I'm working with two camera's and it doesn't seem to work. Only one of the two camera's works.

yycho0108:
@rdaems I suspect you might need to set the `serial_no` in order to specify each device during startup (otherwise the first device to be queried will be used in the code, which may not give consistent results).
It should be able to be passed in as a ROS parameter.

rdaems:
Hi @yycho0108 
I did so according to https://github.com/intel-ros/realsense/blob/development/realsense2_camera/launch/rs_multiple_devices.launch but it didn't work.

yycho0108:
@rdaems I tested the current configuration with two D415 cameras today, and it worked without much issue. I'm not the maintainer of the package, but it would help them look into the issue more deeply if you post the error logs for them.

For reference, my configurations are as follows:
```
Ubuntu Xenial 16.04
ROS : Kinetic Kame
librealsense : 2.16.1
camera firmware : 5.10.3
```

peci1:
I tried this PR with a D435 connected to USB 2, and it did not work. I saw the USB device being reset, but then, during the driver initialization, the USB device disappeared at all. This behavior was consistent regardless of USB port on the computer.

I think the resetting behavior should be optional - e.g. a ROS parameter should define if you want to reset the device or not.

madsherlock:
This PR fixed my issue: Reconnecting to a D415 and/or D435 was impossible without unplugging the sensor(s) first.
I agree that the resetting behaviour should be optional, as it adds a delay to the initialization.
However, my understanding is that this is a temporary workaround until the actual bug is fixed in firmware, which of course makes it less of a deal in the long term.

RachitB11:
Hi @yycho0108 ,
I tried this PR with 2 D415 s and its not working for me. The callback in both instances gets triggered when one of the cameras comes back causing only the first camera to get reset while the other one never comes back.

doronhi:
@yycho0108 ,
Did you check it with the last version on development branch?
I added initial_reset option to rs_multiple_devices.launch (check out the latest commit) and ran the command:
`roslaunch realsense2_camera rs_multiple_devices.launch serial_no_camera1:=725112060411 serial_no_camera2:=727612060220 initial_reset:=True`

Using 2 D415 cameras. Firmware of one is: 05.11.02.00 and the other: 05.10.11.90

Everything seems to work fine.

PR #458, opened 2018-09-05
Add ROS parameter for the required firmware version.

shuntaraw:
When the system is involved with many cameras that are partially replaced/added over time, synchronizing the firmware of the all connected cameras is not a trivial task. It would be nice if the sensors with outdated firmware could automatically be detected. With this PR, a ROS parameter is added to verify that the current firmware is same as or newer than the known stable version. This PR does not change the default behavior.

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -103,6 +103,8 @@ namespace realsense2_camera
         };
 
         static std::string getNamespaceStr();
+        static bool isFirmwareLatest(const std::string& current_version,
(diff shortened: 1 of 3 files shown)


130s:
@shuntaraw or anyone, any reason why this PR was closed w/o even a discussion? I opened a ticket to discuss an issue related to wrong firmware version https://github.com/IntelRealSense/realsense-ros/issues/1681, and this PR seems like an interesting idea.

shuntaraw:
I don't remember why this PR was closed. Please feel free to reuse it if it makes sense. I no longer work on this project and cannot verify the code.

Issue #419, opened 2018-07-19
Possible time drift


pavloblindnology:
According to next lines, if device time drifts away from host time, then ROS frame timestamps will also drift, because software synchronization is done only once after the 1st frame retrieval.
https://github.com/intel-ros/realsense/blob/5210e673ec0d266c62590b433c8b568954f4fdfd/realsense2_camera/src/base_realsense_node.cpp#L543
https://github.com/intel-ros/realsense/blob/5210e673ec0d266c62590b433c8b568954f4fdfd/realsense2_camera/src/base_realsense_node.cpp#L550
The drift could be avoided if we had possibility to get current camera time.
Then we could compare it to frame timestamp and convert the latter to host time with the next formula:
`frame_timestamp_host = current_timestamp_host - (current_timestamp_camera - frame_timestamp_camera)`

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

PR #461, opened 2018-09-07
Estimate clock offset by low-pass filter.

shuntaraw:
The clock on realsense camera is significantly skewed, and the message timestamp diverges from ROS/system clock if a constant offset is added as currently implemented. This PR add the low-pass filter to estimate the offset that can gradually change overtime. The filter is 1-tap IIR, and its computational cost is negligible. 

Here is the comparison of measured offset, filtered offset (this PR), and constant offset (current implementation). The offset is the difference between device timestamp and ros::Time::now() in callback. The clock is clearly skewed i.e. the offset is not constant, and the filtered result fits to the measurement well.    
https://paste.pics/263f5be4d111c3a36279d8389fdf5b1c

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -147,6 +147,8 @@ namespace realsense2_camera
                         const rs2_extrinsics& from_to_other,
                         std::vector<uint8_t>& out_vec);
 
+        void updateClockOffset(const ros::Time& ros_time, const rs2::frame& frame);
+
         std::string _json_file_path;
         std::string _serial_no;
         float _depth_scale_meters;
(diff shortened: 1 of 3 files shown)


shuntaraw:
A better solution is proposed in #419

shuntaraw:
I found #419 does not solve the problem because RS2_FRAME_METADATA_BACKEND_TIMESTAMP also jitters. Since I've deleted the branch let me create a new one.

shuntaraw:
I found #419 does not solve the problem addressed in this PR because RS2_FRAME_METADATA_BACKEND_TIMESTAMP also jitters.

pavloblindnology:
@shuntaraw Solution proposed in #419 has nothing to do with RS2_FRAME_METADATA_BACKEND_TIMESTAMP.
#419 proposes to compute frame's host timestamp with the next formula
`frame_timestamp_host = current_timestamp_host - (current_timestamp_camera - frame_timestamp_camera)`. I.e. we synchronize host and device time by current time.

shuntaraw:
@pavloblindnology current_timestamp_host is obtained via RS2_FRAME_METADATA_BACKEND_TIMESTAMP. I'm not sure if it it jitters, because the metadata is not included (apparently removed after postprocessing) with my D435.

pavloblindnology:
@shuntaraw In #419 `current_timestamp_host` is meant to give host time at the moment of function call, and `current_timestamp_camera` to give camera time at the moment of function call. They have nothing to do with frame metadata.
You probably don't have `RS2_FRAME_METADATA_BACKEND_TIMESTAMP` because you haven't patched v4l on Linux following the installation instructions.

Issue #429, opened 2018-08-01
Given depth profile is not supported even when depth is turned off.


alspitz:
It seems that the ROS node checks if the depth profile is enabled even if no depth stream is requested. This results in the error:

`Given depth profile is not supported by current device!`

I would think the check at https://github.com/intel-ros/realsense/blob/5210e673ec0d266c62590b433c8b568954f4fdfd/realsense2_camera/src/base_realsense_node.cpp#L839 should only happen if the user requests a depth stream.

Is only streaming color images not supported or am I doing something wrong?

J-Rojas:
Realsense viewer allows the depth projector to be entirely disabled so this should be supported through some configuration.

doronhi:
I added PR#601 to fix this issue while preserving static transforms.
It's been a long while but, if it's not too much trouble, I would appreciate your checking this out before I merge. 
Thanks and sorry for the long delay.

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

RealSenseCustomerSupport:
----------------------------------------------

Close this issue for lack of activity. Please feel free to reopen if the issue is still relevant.

PR #462, opened 2018-09-09
Fix crashes related to disabling the depth sensors at launch.

J-Rojas:


J-Rojas:
Fix for issue #429

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -87,7 +87,9 @@ void BaseRealSenseNode::publishTopics()
     setupDevice();
     setupPublishers();
     setupStreams();
-    publishStaticTransforms();
+    if (_enable[DEPTH]) {
+      publishStaticTransforms();
```
shuntaraw, line 91: We want tf for enabled streams to be published, such as color_frame to camera_link even if depth is disabled.

J-Rojas: Okay, the publishStaticTransforms() function would have to be refactored so the other tf frames are independent of the depth frame_id. They are currently coupled to the extrinsics of the depth camera, which IIRC, are not available if the depth camera is disabled, causing a crash.

doronhi:
Added ability in PR#601. Thanks for the code tips and insights.

Issue #448, opened 2018-08-22
Tf problem when running multiple devices


bsaendig:
Hello,
rs_multiple_devices.launch creates different namespaces for the topics, but the tfs for the cameras are all named the same (camera_link).
I think it would be very useful if the camera manager had a parameter to set the name for the specific camera.

Thanks in advance
Bertram

rupak-d:
Hi bsaendig,
Have a look at https://github.com/intel-ros/realsense/issues/346

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

RealSenseCustomerSupport:
----------------------------------------------

Close this issue for lack of activity. Please feel free to reopen if the issue is still relevant.

Issue #354, opened 2018-03-30
How to change the tf name for D435 cameras?


tcjcxy30:
Hi ：
    I use a D435 to replace kinect v1 on my robot. It's hard to me because there are three problem.
    1. the tf link of the color frame is called `camera_color_frame`, but the kinect named `camera_rgb_frame`. So I change the str in the file `include/constrants.h` from `camera_color_frame` to `camera_rgb_frame`. Also i change the rgb value in the launch file. By this way, the rbg image `camera_info` can be related to `camera_rgb_frame`, but the point cloud can't be seen in rviz now, it so wired. So how can i do to change the color frame to camera_rgb_frame?
    2. After I change the code by 1, the D435 can't be reconnected again. I need to pull out the usb and pick again, the D435 can be reconnected. I get the erroe warning  `(backend-v4l2.cpp 808) empty frame has arrived` So what wrong with me?
    3. The tf name is hard to change, in openni, i can name the frame like `<arg name="camera" default="my_camera" />` and the link can be seen as `my_camera_link`, but here I can't do like this. So how to change the link name easily?

neomanic:
1. Don't go changing the realsense source, that'll just end up being a pain. Rather, you should use the [tf_remap](http://wiki.ros.org/tf#tf_remap) node to remap /camera_color_frame to /camera_rgb_frame

2. As I commented in issue #346, this is a known issue. I have to re-plug the cameras every time I close the ROS node.

3. Use tf_remap node

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

Issue #346, opened 2018-03-19
A new tf attachment is not created for multiple D4XX cameras


AndyZe:
With the previous RealSense cameras, the tf link that the camera would attach to could be namespaced like this:

`  <arg name="camera"          default="camera_ee" />`

Where this camera would now attach to camera_ee_link. I haven't found a way to do this with the new D4XX. They seem to always attach to "camera_link" despite a namespace:

`  <arg name="namespace"           default="camera_ee"/>`  <== does not work

neomanic:
I'm just dealing with this exact problem, having dual D435s... and found a bit of a hack workaround. I have dual point-clouds in Rviz as of 5 minutes ago.

The parameter for the base `camera_link` is there in the source (see include/constants.h), but it's not exposed in the launch files. So I made my own launch file by copying `nodelet.launch.xml`, and altered it to expose the `base_frame_id` as a parameter, see  [r435.launch.txt](https://github.com/intel-ros/realsense/files/1828090/r435.launch.txt) which you can use as-is.

I also added a `tf_prefix` parameter that gets prefixed to all the frame ids so I could easily create a parent launch file as follows:
```
<launch>
  <arg name="rs_left_serial"    default="xxxxxx_redacted"/>
  <arg name="rs_right_serial"    default="xxxxxx_redacted"/>
  
  <group ns="rs_left">
    <include file="$(find my_robot_src)/launch/realsense_d435.launch"> 
      <arg name="serial_no"         value="$(arg rs_left_serial)"/>
      <arg name="tf_prefix"           default="rs_left_"/>
      <arg name="enable_pointcloud" value="true"/>
      <arg name="enable_sync"       value="false"/>
      <arg name="align_depth"       value="false"/>
      
      <arg name="depth_fps"                value="6"/>
      <arg name="infra1_fps"               value="6"/>
      <arg name="infra2_fps"               value="6"/>
      <arg name="color_fps"                value="6"/>
    </include>
  </group>

... do other camera here in group rs_right ...
</launch>
```

AndyZe:
Thank you!

tcjcxy30:
Hello @neomanic 
        I try to use you r435.launch.txt to change the published  tf name like `camera_depth_frame` to `xxx_depth_frame`. But I still get the error info `(backend-v4l2.cpp 808) empth frame has arrived`. 
        I only change your code at line 68, 
[...]

PR #471, opened 2018-09-18
new launch parameter for frame distinction in multi camera use

baumanta:
Added a new parameter "tf_prefix" in launch file to be able to use multiple cameras. Before, when multiple cameras were started, all published their topics in frames with the same name (eg. camera_link). Now all cameras can publish to distinct frames (eg. right_camera_link, left_camera_link)

### DIFF
--- a/realsense2_camera/launch/includes/nodelet.launch.xml
+++ b/realsense2_camera/launch/includes/nodelet.launch.xml
@@ -1,6 +1,7 @@
 <launch>
   <arg name="manager"             default="realsense2_camera_manager"/>
   <arg name="serial_no"           default=""/>
+  <arg name="tf_prefix"           default=""/>
   <arg name="json_file_path"      default=""/>
   <arg name="rosbag_filename"     default=""/>
   <arg name="depth"               default="depth"/>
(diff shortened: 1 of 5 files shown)


doronhi:
Sorry for the long delay. It has nothing to do with contributor agreement, just my availability.
Could anyone please explain why other topics (i.e. camera/image_raw) use group's namespace and the static broadcaster for the tf topics (i.e. camera_link) fail to do it?
Anyway, if it is the case then the solution seems ok. I would like to suggest that the default value for tf_prefix will be $(arg camera). Any thoughts?

neomanic:
Commenting here because @baumanta has fixed the issue raised in issues #346, #354, and #448. Yay and thanks!

I was perplexed at first too and it took me more than a few hours to resolve it as per the launch file I posted in #346.

To answer the question @doronhi AFAICT it's because the tf tree is entirely separate from the topics namespacing, which makes a bit of sense if you think of how retrieving sensor data is a separate issue altogether from the sensor's physical location on the robot. 

The following image shows how I have it set up on my robot: the tfs follow the physical build, whereas the topics are set up quite differently: all the sensors are at the root of the tree with a namespace each (so I have /rs_right/*, /rs_left/*, /lidar/*, etc).

tf_tree

I support @doronhi to have the default value `$(arg camera)`. Disclaimer: that's not from knowing whether it's the standard pattern or not, it's just because I don't like it when nodes dump a bajillion tf nodes right into the root /tf namespace and clutter it up.

baumanta:
@doronhi @neomanic thanks for reviewing! I agree, that it would be nice to have the parameters "camera" and "tf_prefix" coupled by default. But if we set the default for "tf_prefix" to camera, then the frames would be named:

camera_camera_link
camera_camera_depth_frame
....

by default. Which I think is not so nice. So one option would be to change the naming convention, such that the frames without prefix are only:  _link, _depth_frame ....
In this case if we default "tf_prefix" to "camera"  we would have:

camera_link
camera_depth_frame

But if the "tf_prefix" is set differently, it would allow for names like:

rs_link
rs_depth_frame ...

If that is ok, I would gladly do those changes!

doronhi:
Thank you both, so much, for your comments. I feel like your second option, @baumanta, of naming the frames without prefix to _link, _depth_frame etc. is very appropriate. As the default namespace is "camera" everything would stay unchanged for those users who didn't face the problem so far. In addition, when the namespace name is changed I don't see a problem with the tf nodes' names change as well.
For users for whom it is important to keep the namespace and the tf names apart it's still an option to bypass the default.
To conclude, I'll be happy if you could change the default value for tf_prefix to be $(arg camera) and also, as you suggested, change "the naming convention, such that the frames without prefix are only: _link, _depth_frame ...."

doronhi:
merged. Thank you guys!

PR #475, opened 2018-09-19
Add librealsense2 dependency to package.xml

shuntaraw:
The dependency on librealsense2 is missing in package.xml. This causes build error in building using catkin tools.   

[repro]
mkdir -p test/src
git clone https://github.com/intel-ros/realsense test/src/librealsense
git clone https://github.com/intel-ros/realsense test/src/realsense
catkin build --workspace test

[expected]
build suceeds

[actual]
CMake Error at .../test/src/realsense/realsense2_camera/CMakeLists.txt:37 (message):
   Intel RealSense SDK 2.0 is missing, please install it from https://github.com/IntelRealSense/librealsense/releases

### DIFF
--- a/realsense2_camera/package.xml
+++ b/realsense2_camera/package.xml
@@ -13,6 +13,7 @@
   <author email="sergey.dorodnicov@intel.com">Sergey Dorodnicov</author>
   <author email="doron.hirshberg@intel.com">Doron Hirshberg</author>
   <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>librealsense2</build_depend>
   <build_depend>message_generation</build_depend>
   <build_depend>roscpp</build_depend>
   <build_depend>sensor_msgs</build_depend>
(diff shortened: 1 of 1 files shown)


AustinDeric:
-1 on this unfortunately.

to install items using the package.xml you will need to do a rosdep install.  However, librealsense2 is not a package that's able to be installed using rosdep according to the list of packages:  https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml

In order to test this you can add librealsense2 to the package.xml, as you have done, and run the following:
```
mkdir -p catkin_ws/src
cd catkin_ws/src
git clone https://github.com/intel-ros/realsense
rosdep install --from-paths src --ignore-src -r -y 
cd ..
catkin build 
```

I believe your commit will still fail.  

The solution is to add librealsesne2 to the rosdep keys file referenced above and then this PR would be +1.  Without this the error you are seeing is the correct way to go because it prompts the user to install the library manually.

shuntaraw:
The issue I want to address here is that the build of librealsense is not triggered by the build of realsense2_camera when both codes are in the workspace. In this case, we need to run 'catkin build' twice to propagate the change in librealsense to realsense2_camera, which is clearly error prone.  I am aware that librealsense is not a catkin package and rosdep install will not work properly. There may be a better way to define the source-code dependency on non-catkin packages in cmake.

PR #481, opened 2018-09-26
Initialize the reconfigure server with the private nodehandle

Rayman:
In a nodelet the default nodehandle will be the manager's handle, but
all the parameters are loaded in the private nodehandle. This patch
initialzes all the reconfigure servers such that the parameters are
properly loaded from the private nodehandle.


shaun-edwards:
+1 for this change, although it will break anyone setting parameters on launch.

Rayman:
I'm not working with the realsense anymore so I'll revisit this issue at a later moment

Issue #477, opened 2018-09-20
D400 series: a method to pause streaming


AndyZe:
Similar to 'force_power' that the previous generation had.

This is important to me so I can switch between multiple cameras when my bandwidth is limited. Basically, I would like to put all cameras except one or two in sleep mode.

AndyZe:
My PR solves the network bandwidth issue by pausing ROS publication. It would be better to pause the RealSense camera at the hardware level, though, to save USB3 bandwidth and power, too. I didn't see an easy way to do that.

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

RealSenseCustomerSupport:
----------------------------------------------

Close this issue for lack of activity. Please feel free to reopen if the issue is still relevant.

PR #487, opened 2018-10-02
Add dynamic_reconfigure option to pause a sensor (issue #477)

AndyZe:
I hope you guys can see the benefits of being able to pause RealSense sensors, especially on mobile robots that have limited network bandwidth or low-power CPUs. This PR adds a check box in dynamic_reconfigure that allows any D415/D435 to be paused. Using `top`, I see the CPU load drop from ~10% to ~1% when a sensor is paused. I'm sure network data rate drops, too.

The sensor can also be paused programmatically with the dynamic_reconfigure API.

I will mention, I do not see the USB traffic drop when a sensor is paused. I would like to fix that, too, if you have any suggestions on how to do that.


AndyZe:
@doronhi @mattcurfman  Hello, is anybody there? I would like to get this merged for a NASA-related project.

doronhi:
Hi @AndyZe,
Sorry about my lack of response. I am actually looking on and off at your proposal for couple of days now.
I feel like the correct solution should be that if there are no subscribers there should be no CPU usage. That the additional parameter is not necessary.
I didn’t managed to check it through yet and hence my lack of response. Again, my apologies.
What do you think?
In any case, I hear your urgency and I’ll try to get to that really ASAP.

AndyZe:
Good to hear, thanks.

ROS typically doesn't use USB for communications, so I don't think that's a ROS issue. My guess is, CPU data rates would remain high if you used the .stop() method even outside of ROS.

Anyway, higher CPU data rates don't affect my current application. I'm more concerned with ethernet bandwidth and CPU load. I could see it being a problem if somebody were to use, like, 4 RealSense devices at once.

Edit: I'm testing that USB traffic drops with realsense-viewer. It looks like it does, as it should. I'm checking with `sudo usbtop`.

doronhi:
Hi @AndyZe ,
I merged your pull request.
My thought were that I would like to disable the sensors, the same way you did, but according to whether there are subscribers for them or not. i.e. I think it is a good idea to disable unused sensors and it should be an inner logic of the wrapper.
I realized, however, that your use case may require you to disable the sensor for saving resources without unsubscribing from the topic.

PR #490, opened 2018-10-11
Fixed missing Dependencies

ipa-fog:
Updated CMakeLists.txt and Package.xml to fix missing dependency for pkg "rgbd_launch".


doronhi:
I reconsidered and decided it's best to avoid adding rgbd_launch as a dependency.
rs_rgbd.launch is an example of using the rgbd module but no need to make everyone need to install an extra package.

PR #509, opened 2018-11-04
Add missing pcl library to tests_camera_core

jerry73204:
The PR is due to the compile error on Arch Linux.
```
/usr/bin/ld: CMakeFiles/tests_camera_core.dir/test/camera_core.cpp.o: undefined reference to symbol '_ZN3pcl7console5printENS0_15VERBOSITY_LEVELEPKcz'
/usr/bin/ld: /usr/lib/libpcl_common.so.1.8: error adding symbols: DSO missing from command line
```
The cause is that the tests_camera_core module requires `pcl::console::print`, which library `-lpcl_common` is missing in command line. The patch fixes it by adding PCL libraries in cmake.

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -40,6 +40,8 @@ find_package(catkin REQUIRED COMPONENTS
   ${REALSENSE_CATKIN_BASED_DEPS}
 )
 
+find_package(PCL REQUIRED)
+
 add_message_files(
   FILES
   IMUInfo.msg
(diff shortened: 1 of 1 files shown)


doronhi:
I don't understand what is it that you fix.
tests_camera_core and camera_core.cpp are not part of realsense2_camera.

jerry73204:
I see. It is a compatibility fix for indigo release. You may see the PR is merged to indigo-devel. The fix is due to the fact that the modern CMake on Arch Linux does not link correct pcl library.

PR #520, opened 2018-11-26
Add external time reference restamping [enable VIO with RS435 IR frames]

m0oz:
Adds class ExternalTimestamping that allows to subscribe to a topic with timestamps that were captured using an external micro controller. Incoming frames are matched to hw timestamps, re-stamped and then published.

The matching works both ways: 
An incoming timestamp will be buffered if no frame was yet received or vice versa. We wait until we receive frame and timestamp within a specified time interval. Then, we use seq IDs if frame and stamps to determine an offset between the two streams. Any subsequent stamp/frame pairs will be correlated using this seq offset.
If we drop a frame or stamp or if the two for some reason do not arrive within the specified time interval we reset the sync mode and wait to re-initialize based on the next incoming pair.

The interval within that we expect stamp and frame to arrive can be specified using the variables kMaxStampAge and kMaxFrameAge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,35 @@ to activate the filters, use the argument "filters" and deperate them with a com
 roslaunch realsense2_camera rs_camera.launch filters:=temporal,spatial,pointcloud
 ```
 
+### Restamp depth and infra-red images using an external time reference (VIO setup)
+By default, depth and IR frames are stamped using frame metadata. The metadata timestamps are based on the RealSense's on-board clock. For applications like VIO that require time-synchronization of multiple sensors such as GPS, IMU and cameras it is required to read all of them based on the same time reference. The RealSense D435i will allow you to get time-synchronized image and IMU streams without any further setup. However, if you intend to use more sensor than just IMU and camera or if you are using D435/D415 without and onborad IMU you can use an external time reference and restamp frames before publishing.
+
+The realsense will try to match incoming depth and IR frames to stamps that it receives on topic /hw_stamp.
+As soon as it correlates frame and stamp the frame is re-stamped and published. Note that depth and IR and depth frames will not be published until the node receives timestamps to restamp them.
+
+Requirements:
+* RealSense D435: FW 5.10.3 or newer, D415: 5.9.14 or newer
+* Seperate microcontroller (MC) that allows to capture 1.8V pulses.
+* ROS wrapper to connect to the MC and publish messages with the captured timestamps
+
+Setup:
+* Prepare the microcontroller to read camera output pulses. This could be an arduino or a UAV's flight controller. All the other relevant sensors should be read by this board so that they are based on the same time reference. Setup the relevant digital input pin and make sure that you can capture pulses of 100us length. Note: You might need to convert the signal to a higher voltage if you board cannot caputure pulses at 1.8V
+* Setup the microcontroller's ROS node to publish a message of type std_msgs/Header to rostopic /hw_stamp for each measured pulse.
+* Connect the cameras digital output pin the MCs input pin (Check the multi-cam [whitepaper](https://realsense.intel.com/wp-content/uploads/sites/63/Multiple_Camera_WhitePaper_rev1.1.pdf) for the correct pin mapping).
+* Set the cameras inter_cam_sync mode to 1 to run the device as master (If you use multiple cameras: Set one camera as master and the rest as slave).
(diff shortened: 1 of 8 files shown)


realsense2_camera/include/constants.h
```diff
@@ -35,6 +35,7 @@ namespace realsense2_camera
     const bool ALIGN_DEPTH    = false;
     const bool POINTCLOUD     = false;
     const bool SYNC_FRAMES    = false;
+    const bool MAVROS_TRIGGERING = false;
```
m0oz, line 38: Not needed here, to be removed

README.md
```diff
@@ -109,6 +109,35 @@ to activate the filters, use the argument "filters" and deperate them with a com
...
+As soon as it correlates frame and stamp the frame is re-stamped and published. Note that depth and IR and depth frames will not be published until the node receives timestamps to restamp them.
+
+Requirements:
+* RealSense D435: FW 5.10.3 or newer, D415: 5.9.14 or newer
+* Seperate microcontroller (MC) that allows to capture 1.8V pulses.
+* ROS wrapper to connect to the MC and publish messages with the captured timestamps
+
+Setup:
+* Prepare the microcontroller to read camera output pulses. This could be an arduino or a UAV's flight controller. All the other relevant sensors should be read by this board so that they are based on the same time reference. Setup the relevant digital input pin and make sure that you can capture pulses of 100us length. Note: You might need to convert the signal to a higher voltage if you board cannot caputure pulses at 1.8V
+* Setup the microcontroller's ROS node to publish a message of type std_msgs/Header to rostopic /hw_stamp for each measured pulse.
```
mhkabir, line 125: A word of caution here : I think that this will not actually work as intended since ROS overwrites the `header.seq` field when publishing a topic.

andre-nguyen:
@m0oz Is this functional or was this a proof of concept to start a discussion?

m0oz:
@andre-nguyen I used the feature branch for testing and had it running with the occasional quirk.
However, I did not rebase or maintain the feature as it does not seem like it would get merged any time soon as there is no feedback at all from the intel team on this PR

Honestly, I would recommend to get yourself one of the new D435i cameras, which have an onboard IMU with hw time synchronization and spare yourself from setting it up and testing it yourself ;)

andre-nguyen:
@m0oz Actually I rebased this on top of master and I'm currently using it. Curiously enough I sometimes have dropped frames where I have a time stamp but no frame to go with it. @doronhi I think this would be a very interesting feature to merge in provided we rebase it and parameterize the streams in the launch file which should be processed by the timestamp.

SayyidSyed (approved):

doronhi:
@m0oz
I want to apologize for not providing a feedback much sooner.
The truth of the matter is that your original PR seemed very difficult to check without an appropriate external micro controller. It also seemed very difficult to validate in terms of accuracy of synchronization and frame drops. As it didn’t seem like a common request it got pushed down in priority.
From what I read in the later posts, it seems that the D435i answers the original problem. Is that true?
Also, not long ago, a global time stamp domain was added. It stamps the frames with a system time which can then be synced with any other device connected to the system. Is this a way to tackle the original issue?

mhkabir:
@doronhi Are there any docs on the global timestamp domain stuff? How is the camera's time synchronised to the host system time?

mhkabir:
The original intent of this PR was to allow timestamping of image frames with sub-millisecond accuracy against the host PC clock.

doronhi:
The global time algorithm assumes a linear equation between the device’s HW clock and system clock.
Each frame’s HW timestamp is converted to system timestamp using this equation.
A separate thread occasionally samples the device for its clock reading, taking the reply time into consideration, and the algorithm updates the equation’s parameters accordingly.
I measured the reported time difference between 2 frames taken simultaneously by 2 different cameras. Using this method I measured  less then 1 millisecond difference in 99.9% of frames and less then 2 milliseconds in the rest of the frames.
In the current librealsense release, 2.24.0, the algorithm still suffer from noise in the first couple of seconds, until it gathered enough samples and stabilized. This also will be corrected shortly [#4396](https://github.com/IntelRealSense/librealsense/pull/4396)

mhkabir:
Unless you're doing a two-way message exchange with the host computer, the one-way translation you describe can only yield timestamps which are accurate with respect to each other (i.e accurate delta times). If I'm not making a mistake, there will still be a (generally unoberservable) static offset with respect to the host timebase. This makes that method unsuitable for use where there are other sensors (e.g an IMU) correctly timestamped against the host timebase, unless you estimate the static offset by e.g a visual-inertial calibration.

PR #537, opened 2018-12-14
Fixed: invalid module name format for ROS

carlos-wong:


doronhi:
Hi @carlos-wong ,
Can you please explain why this change is needed? Is it customary ? Is there a protocol definition?
Another question, which sensor has a name with '-' in it so that your change applies to it?
Thanks,

carlos-wong:
Hi @dorodnic , when I use the sr300. it make the module name "Coded-Light Depth Sensor". It will cause invalid module name error on ROS kinetic.

Issue #547, opened 2018-12-20
Compilation error


atyshka:
I've just pulled the latest changes to the repo (excited about the reset pr!) and I'm unable to build. Here's the error I'm getting:

`/usr/include/c++/7/bits/stl_iterator_base_types.h:205:5: error: no type named ‘iterator_category’ in ‘struct std::iterator_traits<rs2::frameset::iterator>’`

I'm using the latest realsense drivers. Was there a recent API change that broke the code? Can anyone else reproduce this issue?

doronhi:
Sorry about that. You need to take the latest development branch from librealsense.
It’s been update a couple of hours ago.

atyshka:
Aww... another thing to compile from source? I guess if I have to I'll do so but I'd prefer to have the ros drivers work with the general releases.

doronhi:
You are absolutely right. The next release will be coordinated with librealsense’s. Hopefully in a matter of days.

mattcorsaro1:
I'm still getting this error. I'm using librealsense-2.17.0 and the development branch of this package. ROS Kinetic on 16.04, kernel 4.4.0-142-generic.

`In file included from /usr/include/c++/5/algorithm:62:0,
                 from /usr/include/boost/smart_ptr/shared_ptr.hpp:37,
                 from /usr/include/boost/shared_ptr.hpp:17,
                 from /opt/ros/kinetic/include/class_loader/./class_loader.hpp:36,
                 from /opt/ros/kinetic/include/class_loader/class_loader.h:35,
                 from /opt/ros/kinetic/include/pluginlib/./class_list_macros.hpp:40,
                 from /opt/ros/kinetic/include/pluginlib/class_list_macros.h:35,
                 from /home/mcorsaro/kuka_ws/src/realsense/realsense2_camera/src/../include/../include/realsense_node_factory.h:6,
                 from /home/mcorsaro/kuka_ws/src/realsense/realsense2_camera/src/../include/base_realsense_node.h:6,
                 from /home/mcorsaro/kuka_ws/src/realsense/realsense2_camera/src/base_realsense_node.cpp:1:
[...]

Issue #554, opened 2018-12-24
Compilation error


AdamPengG:
I got the D435I camera. Then I install the sdk by "sudo apt-get install librealsense2-dev" and git clone the ROS wrapper.But the terminal shows a compilation error :
"/usr/include/c++/5/bits/stl_algo.h:162:34: error: no matching function for call to ‘__iterator_category(rs2::frameset::iterator&)’"
I have read the issue #547.Shall I reinstall the newest Realsense sdk or ROS wrapper.Another way ,maybe wait your new Ros wrapper.

doronhi:
Fixed. Development branch should now work with librealsense 2.17.0 (latest release).

Issue #552, opened 2018-12-22
catkin_make error with the latest librealsense


koji:
`catkin_make -DCATKIN_ENABLE_TESTING=False -DCMAKE_BUILD_TYPE=Release`

Got the following errors. Did anyone face this?

```
[ 96%] Building CXX object realsense/realsense2_camera/CMakeFiles/realsense2_camera.dir/src/realsense_node_factory.cpp.o
In file included from /home/parallels/catkin_ws/src/realsense/realsense2_camera/src/../include/base_realsense_node.h:7:0,
                 from /home/parallels/catkin_ws/src/realsense/realsense2_camera/src/base_realsense_node.cpp:1:
/home/parallels/catkin_ws/src/realsense/ddynamic_reconfigure/include/ddynamic_reconfigure/ddynamic_reconfigure.h:213:0: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
           #pragma clang diagnostic push // CLion suppressor
 ^
/home/parallels/catkin_ws/src/realsense/ddynamic_reconfigure/include/ddynamic_reconfigure/ddynamic_reconfigure.h:214:0: warning: ignoring #pragma ide diagnostic [-Wunknown-pragmas]
           #pragma ide diagnostic ignored "OCUnusedGlobalDeclarationInspection"
 ^
In file included from /home/parallels/catkin_ws/src/realsense/realsense2_camera/src/../include/base_realsense_node.h:7:0,
                 from /home/parallels/catkin_ws/src/realsense/realsense2_camera/src/realsense_node_factory.cpp:5:
/home/parallels/catkin_ws/src/realsense/ddynamic_reconfigure/include/ddynamic_reconfigure/ddynamic_reconfigure.h:213:0: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
           #pragma clang diagnostic push // CLion suppressor
 ^
/home/parallels/catkin_ws/src/realsense/ddynamic_reconfigure/include/ddynamic_reconfigure/ddynamic_reconfigure.h:220:0: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
           #pragma clang diagnostic pop
 ^
/home/parallels/catkin_ws/src/realsense/ddynamic_reconfigure/include/ddynamic_reconfigure/ddynamic_reconfigure.h:214:0: warning: ignoring #pragma ide diagnostic [-Wunknown-pragmas]
[...]

PR #555, opened 2018-12-24
fix to work with librealsense v2.17.0

doronhi:
Allow compiling branch development with librealsense release v2.17.0.


doronhi:
Fix issues #547, #554, #552

Issue #497, opened 2018-10-19
Incorrect TF frame `camera_link` when using `view_d435_model.launch`


peci1:
When you are viewing the model using `view_d435_model.launch`, this is the TF you get:

```
rosrun tf tf_echo camera_link camera_depth_frame
At time 1539958677.743
- Translation: [0.015, 0.018, 0.013]
- Rotation: in Quaternion [0.000, 0.000, 0.000, 1.000]
            in RPY (radian) [0.000, -0.000, 0.000]
            in RPY (degree) [0.000, -0.000, 0.000]
```

Whereas with the live camera, you get: 

```
rosrun tf tf_echo camera_link camera_depth_frame
At time 0.000
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [0.000, 0.000, 0.000, 1.000]
            in RPY (radian) [0.000, -0.000, 0.000]
            in RPY (degree) [0.000, -0.000, 0.000]
```

Indeed, in the xacro file used in viewing, there is a [non-zero transform between camera_link and camera_depth_frame](https://github.com/intel-ros/realsense/blob/a63471f6b3885047878d623308ca4c92ac368a85/realsense2_camera/urdf/_d435.urdf.xacro#L74), whereas [in the driver source, camera_link and camera_depth_frame coincide](https://github.com/intel-ros/realsense/blob/a63471f6b3885047878d623308ca4c92ac368a85/realsense2_camera/src/base_realsense_node.cpp#L1104).

I assume the XACRO file should be fixed.

atyshka:
@doronhi Here is a detailed video of the issue. I started the model which had the correct transforms but then everything is offset when the driver is started.
![peek 2018-11-13 10-36](https://user-images.githubusercontent.com/19317207/48425147-33e7a400-e732-11e8-9dc7-6ed4073c7eaa.gif)

When the driver is started the frames no longer line up with the geometry.
@peci1 recommends changing the xacro, but I like the origin where it is. The mounting point is a lot more logical than the 0 point corner of the model. I'd propose either changing the transforms in the driver to match the xacro or adding a launch file param to disable tf publishing if static transforms are already being published by a robot state publisher. I'd opt for the second option, many other sensors use this parameter and it won't change behavior unless used.

atyshka:
@doronhi Any suggestions on how you'd like to handle this issue?

PR #567, opened 2019-01-08
Fixed different transforms between xacro and driver

### LINKED ISSUE #469: What's the serial_no for d435 camera should I input in the launch file?
It seems that the TF frames were not same with the real d435 camera model after launched the file "rs_camera.launch". 
I guess the reason was miss of the serial_no in the launch file, so the default model was not d435 but other type.
Dose anyone know the serial_no for d435?
Thanks!

atyshka:
This fixes #469 and #497 

The d435 URDF had an error where there was a non-zero transform for camera_link but the driver output a zero origin for this link. This caused the transforms to shift between starting a robot description and starting the driver. This commit brings the transforms in line with the driver while keeping the visual and collision geometry the same.

### DIFF
--- a/realsense2_camera/launch/rs_multiple_devices.launch
+++ b/realsense2_camera/launch/rs_multiple_devices.launch
@@ -5,11 +5,13 @@
   <arg name="camera2"              			default="camera2"/>		<!-- Note: Replace with camera name -->
   <arg name="tf_prefix_camera1"           	default="$(arg camera1)"/>
   <arg name="tf_prefix_camera2"           	default="$(arg camera2)"/>
+  <arg name="initial_reset"         default="false"/>
 
   <group ns="$(arg camera1)">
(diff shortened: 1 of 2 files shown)


doronhi (commented):
The problem needs to be solved and I agree that camera_link is the basis for all the other optical nodes and therefor it should have a (0,0,0) transformation between camera_link and camera_depth_frame.
I feel there should be another joint between the physical mounting point and camera_link.
Would you check out my suggestion in:
https://github.com/doronhi/realsense/tree/atyshka-development (last commit)

If you agree with the concept, please pay attention to the collision section. I didn't attend to it.

atyshka:
@doronhi I rebased your branch onto mine and fixed the collision coordinates. All that needed to be done was shifting with the same units as you did for the mesh. This seems like a good solution, thanks for the input!

doronhi:
Thank you, for your patience and persistence.

Issue #583, opened 2019-01-22
Decimation filter does not adapt CameraInfo


stwirth:
Thanks for adding the support for the decimation filter!
I have one issue using it:
When running
```
roslaunch realsense2_camera rs_camera.launch depth_width:=1280 depth_height:=720 filters:=decimation
```
I see the decimation filter parameters in dynamic reconfigure and can manipulate them. When I change the filter magnitude, I see that the image size changes accordingly:
```
$ rosrun dynamic_reconfigure dynparam set camera/decimation 'Filter Magnitude' 1
stwirth@x:~$ rostopic echo -n 1 /camera/depth/image_rect_raw --noarr
header: 
  seq: 21027
  stamp: 
    secs: 1548150346
    nsecs: 234946562
  frame_id: "camera_depth_optical_frame"
height: 720
width: 1280
encoding: "16UC1"
is_bigendian: 0
step: 2560
data: "<array type: uint8, length: 1843200>"
---
$ rosrun dynamic_reconfigure dynparam set camera/decimation 'Filter Magnitude' 4
$ rostopic echo -n 1 /camera/depth/image_rect_raw --noarr
header: 
  seq: 21271
  stamp: 
    secs: 1548150354
    nsecs: 365142743
  frame_id: "camera_depth_optical_frame"
height: 180
width: 320
encoding: "16UC1"
is_bigendian: 0
step: 640
data: "<array type: uint8, length: 115200>"
---
```
However, the CameraInfo message content does not change:
```
$ rosrun dynamic_reconfigure dynparam set camera/decimation 'Filter Magnitude' 1
$ rostopic echo -n 1 /camera/depth/camera_info
header: 
  seq: 26057
  stamp: 
    secs: 1548150513
    nsecs: 842895075
  frame_id: "camera_depth_optical_frame"
height: 720
width: 1280
distortion_model: "plumb_bob"
D: [0.0, 0.0, 0.0, 0.0, 0.0]
K: [933.902587890625, 0.0, 649.7789306640625, 0.0, 933.902587890625, 359.9447326660156, 0.0, 0.0, 1.0]
R: [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
P: [933.902587890625, 0.0, 649.7789306640625, 0.0, 0.0, 933.902587890625, 359.9447326660156, 0.0, 0.0, 0.0, 1.0, 0.0]
binning_x: 0
binning_y: 0
roi: 
[...]

PR #587, opened 2019-01-23
fix bug: update camera_info if image size changes.

doronhi:
fix bug in issue #583.


stwirth:
Works well for me, thanks for the quick fix!

Issue #332, opened 2018-03-04
depth_registered pointcloud has incorrect depth scaling


r0b0cop:
If I run roslaunch realsense2_camera rs_rgbd.launch and examine the pointcloud published on /camera/depth_registered/points in RViz, the points appear to be projected incorrectly. For example, a .3m long object in the scene is rescaled to around 2.2m (as indicated by rviz's measure tool). Changing the fixed frame does not solve this.

Running roslaunch realsense2_camera demo_pointcloud.launch works however. Objects published to the topic /camera/depth/color/points are correctly shown as .3m long. Unfortunately, this topic doesn't align points with the color frame. 

The problem could be related to the _depth_scale_meters parameter used in base_realsense_node.cpp, but I haven't been able to find any documentation on that parameter. Am I misunderstanding something about the registered depth cloud, or any suggestions on how to fix this?

Environment info:
SR300 with firmware 3.21.0.0
ROS kinetic

oka1125:
By solution #350, I found another problem has occurred:

    [problem]
    - Striped pattern occurs in the published aligned depth image.
    - Jagged pattern occurs in the published point cloud 2.

I think that "depth_units / meter_to_mm (decimal value)" multiplied to each (upper/lower) byte is not good,
because decimal point values will be disappeared after calculation in each byte.

So I propose new solution like following:

```
                        int out_pixel_index = y * other_intrin.width + x;
+                       uint16_t value = depth / meter_to_mm;
                        //Tranfer n-bit pixel to n-bit pixel
                        for (int i = 0; i < from_bytes_per_pixel; i++)
                        {
                            const auto out_offset = out_pixel_index * output_image_bytes_per_pixel + i;
-                           const auto from_offset = from_pixel_index * output_image_bytes_per_pixel + i;
-                           p_out_frame[out_offset] = p_from_frame[from_offset] * (depth_units / meter_to_mm);
+                           p_out_frame[out_offset] = (value >> (8 * i)) & 0x00FF;
                        }
```

oka1125:
I noticed that the same problem has been already written in the following topic:
    "fix the aligned depth frame unit conversion issue #367 "

jcgarciaca:
I am facing the same issue, do you have any idea how to solve it?
Just for reference, in my case the scale difference is for XYZ.
Thank you.

oka1125:
In the latest version (https://github.com/intel-ros/realsense/tree/2.0.4), this issue seems to be solved.

[...]

PR #605, opened 2019-02-08
fix the aligned depth frame unit conversion issue again

oka1125:
When I run roslaunch realsense2_camera rs_rgbd.launch and examine the pointcloud published on in RViz, the points appear to be projected incorrectly (about 8 times larger than the correct size).

* Environment info:
    - Camera: SR300
    - ROS: kinetic
    - librealsense: v2.18.1
    - realsense_ros: 2.1.4

This issue was reported in following issues and solved in tag 2.0.4 of realsense(ros):
* depth_registered pointcloud has incorrect depth scaling #332
* fix the aligned depth frame unit conversion issue #367

But the same problem occurred again from tag 2.1.1 of realsense(ros).
I think refactoring in tag 2.1.0 is the cause of this problem.

I made a pull request to fix this problem, so please approve it.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -653,6 +653,19 @@ void BaseRealSenseNode::publishAlignedDepthToOthers(rs2::frameset frames, const 
             rs2::frameset processed = frames.apply_filter(align);
             rs2::depth_frame aligned_depth_frame = processed.get_depth_frame();
 
+            static const auto meter_to_mm = 0.001f;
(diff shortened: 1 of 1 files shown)


doronhi:
This is definitely an issue. 
It concerns the depth image, in addition to the align depth image so the correction may be needed to be there.
It reveals another issue that even with D400 devices, changes with dynamic calibration to the depth scale parameters are not reflected in the depth image. 
These issues need to address as well. 
I thank you for the PR for bringing these issues to my attention. I'll attend to them as soon as time permits.

doronhi:
merged into PR#682.

PR #614, opened 2019-02-13
TM265

doronhi:
Add basic support for TM265 Tracking Module.
Include some interface changes: 
- new topics for tracking.
- add new parameters in launch files for fisheye1 and fisheye2 cameras, enable_tm2
- removed duplicate parameters in launch files: parameters for both infra1 and infra2 reduced to infra, as in the devices themselves.
- instead of enable_imu there are the more basic enable_gyro and enable_accel.

Some inner code modifications in order to generalize it a little.
Fix issue with align filter - was initiated every frame.


stwirth:
This looks great! I would suggest using the `tf2_ros` interface instead of `tf` as the latter has been deprecated.

lucasdai (approved):

PR #615, opened 2019-02-13
Added URDF and mesh files to install target

CameronDevine:
I Modified the CMake file so that URDF and mesh files will be installed.

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -122,3 +122,13 @@ install(DIRECTORY rviz/
 install(FILES nodelet_plugins.xml
     DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
     )
+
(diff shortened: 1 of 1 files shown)


doronhi:
Although I don't see a problem, I wander what doesn't work without this change?
Can you give me an example?

CameronDevine:
I realize I should have explained this in the initial pull request.

In its current state, when building `realsense2_camera` in your catkin workspace if you source `catkin_ws/devel/setup.bash` I believe the URDF files can be found by ROS. However if you source `catkin_ws/install/setup.bash` the URDF files will not be in your path, and ROS will not be able to find them.

SteveMacenski:
Moreover, the install space is what is used when releasing a debian package, or when installing software from catkin on another machine 

(also had to fix this on a fork)

doronhi:
Thank you guys. It’ll most likely be in the next release.

PR #631, opened 2019-02-22
T265

schmidtp1:
@doronhi, I noticed that I get high latency in the odometry messages after ~1s and they arrive in batches (see screenshot). This seems to be related to TCP batching and can be avoided by using the no TCP latency option in the subscriber or UDP but unfortunately it seems there is no option for the publisher. I didn't see this issue with other messages like PoseStamped and Vector3Stamped and would prefer to add them to avoid running into this issue for applications that require low latency. (Maybe the odometry could be published at a lower rate then?)

odom_gyro_screenshot from 2019-02-21 14-06-21

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -14,6 +14,7 @@
 #include <sensor_msgs/point_cloud2_iterator.h>
 #include <sensor_msgs/Imu.h>
 #include <nav_msgs/Odometry.h>
+#include <geometry_msgs/PoseStamped.h>
 
 #include <queue>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


schmidtp1:
Please feel free to change the organization of the publishers (map). If this makes sense to you, I would also like to add the velocity twist the same way and we need acceleration (estimate).

doronhi:
As the problem of bursts seems to be related to the TCP protocol and can be bypassed, if needed, by subscribing with UDP protocol, I don't think that, for the time being, we have to solve it here. 
Moreover, I know of usages for odometry messages but I am not familiar with a need for pose message that don't include covariance. Therefor, I think that for now we will not publish another sub-topic for pose only.

PR #639, opened 2019-03-01
Estimate accurate timestamps by calculating clock offset

shuntaraw:
The device clock can advance at the different speed than the system clock, which causes the systematic drifts of timestamps. This change allows to estimate the clock offset using a low pass filter, and compensate for the drift. Experimental results show that the typical drift is about 0.000075 sec/s, equivalent to 1 frame difference in 6 minutes at 30Hz. The drift can be successfully corrected by this patch. 

The typical clock drifts are plotted as follows. 
before: https://prnt.sc/mrmqqb
after: https://prnt.sc/mrmr74

The sporadic jumps in the plots are due to frame latency, and the timestamp after the offset correction is steady and monotonic (as before). This is essentially same PR as #461 and rebased to current HEAD.


doronhi:
librealsense v2.22.0 default time stamps include synchronization between device clock and system clock, based on a linear equation estimation.
I commented in [here](https://github.com/IntelRealSense/realsense-ros/pull/710#issuecomment-494372845)

shuntaraw:
Thanks for the update. I think this PR is no longer necessary and will close it.

PR #643, opened 2019-03-02
T265 coord rviz

schmidtp1:
This PR proposes to revert to the original coordinates that are documented and consistent with librealsense API. For visualization purpose a reference frame is used with z pointing up, i.e. gravity-aligned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,11 @@ Once running it will publish, among others, the following topics:
 - /camera/fisheye1/image_raw
 - /camera/fisheye2/image_raw
 
+The pose (estimate) can be visualized in RViz by starting up the camera with
+```bash
+roslaunch realsense2_camera demo_t265.launch
+```
+
 
 ## Packages using RealSense ROS Camera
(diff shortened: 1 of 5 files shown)


ibaranov-cp:
Are you sure this is consistent with the ROS standard? http://www.ros.org/reps/rep-0103.html
Camera frame has z forwards.. Just double checking :)

schmidtp1:
Hi @ibaranov-cp, thanks for the reference!
In the comment above, I was referring to the "world" reference frame with z pointing up.
The T265 body frame is different from the camera frame (in the sense of imagers) and centered between the two fisheye. Below is the figure from the data sheet (https://www.intel.com/content/dam/support/us/en/documents/emerging-technologies/intel-realsense-technology/IntelRealSenseTrackingT265Datasheet.pdf).
screen shot 2019-03-04 at 10 57 06 pm
I agree that the ROS package/wrappers should follow the ROS convention. Thanks again for the reference!

Issue #609, opened 2019-02-10
DDynamic Reconfigure - Default Values


RhysMcK:
Is there simple way to set the **default** values for the dynamic reconfigurable variables?

stwirth:
I'm running into the same problem. The "traditional" ROS dynamic reconfigure server reads all parameter values form the ROS parameter server on startup and sets the parameters to those values. When a dynamic reconfigure call is made, it also updates the values on the parameter server. This way, running a `rosparam get /` always gets the current configuration of the system, including dynamically configurable parameters. It would be nice if the camera driver behaved the same to be able to save and restore the parameter configuration state.

stwirth:
A workaround for this problem is to include a dynamic reconfigure call in the launch file:
```
  <node name="$(anon param_setter)" pkg="dynamic_reconfigure" type="dynparam"
        args="set /camera/Stereo_Module 'Visual Preset' 3" />
```

stwirth:
Another option is probably to use the JSON config file.

stwirth:
One problem with communicating the realsense ROS node parameters with the ROS parameter server is that in the current implementation some parameters have spaces in them (e.g. "Emitter Enabled") which is not a valid ROS parameter name (see http://wiki.ros.org/Names).

doronhi:
@stwirth 
I don't really understand what is it that you suggest the driver (realsense2_wrapper ) should do?
Can you walk me through your case?

doronhi:
> I'm running into the same problem. The "traditional" ROS dynamic reconfigure server reads all parameter values form the ROS parameter server on startup and sets the parameters to those values. When a dynamic reconfigure call is made, it also updates the values on the parameter server. This way, running a `rosparam get /` always gets the current configuration of the system, including dynamically configurable parameters. It would be nice if the camera driver behaved the same to be able to save and restore the parameter configuration state.

@stwirth 
I don't really understand what is it that you suggest the driver (realsense2_wrapper ) should do?
Can you walk me through your case?

stwirth:
@doronhi this is actually a problem of **d**dynamic_reconfigure.
In dynamic_reconfigure, the initialization sequence is as follows:
* On construction, `init()` is called which pulls values for all parameters from the ROS parameter server (if they are set): https://github.com/ros/dynamic_reconfigure/blob/4f320fcdc3d6cd6fbf7a7f9a969af81ca521b761/include/dynamic_reconfigure/server.h#L190-L192
[...]

PR #667, opened 2019-03-11
Set defaults from param server

efernandez:
This provides a fix for #609 that works as follows for each option:

- It sets the `option_name` to `create_graph_resource_name(rs2_option_to_string(option)))`. Same as before, which is used for both the `ddynamic_reconfigure` and ROS (static) param.
- It takes the default value from `sensor.get_option(option)`, which was passed to the `DD*` param constructors before. Now it's assigned to `option_value`.
- Then it tries to load a param from the `nh1` node handle namespace for the `option_name`, using `option_value` as default. If the param was set in the ROS param server, `nh1.param` returns `true`, in which case we set the option to that value in the sensor with `sensor.set_option(option, option_value)`. Without this the params are set but the sensor still runs with its own defaults.
- For options with `min` and `max`, if the value in the ROS param server is out of the `[min, max]` range, a `ROS_WARN` log message is reported and the default value from the sensor is used, as before.
- For enum options, if the value is greater than the number of enums supported, a `ROS_WARN` log message is reported and the default value from the sensor is used, as before.
- Finally, the `DD*` param is constructed and added to the `ddynamic_reconfigure`.

The fix is provided in this commit: https://github.com/intel-ros/realsense/commit/97dd66ee05326bb9b5b9070035bc437d906c3f73

A better fix would be to make `ddynamic_reconfigure` work as `dynamic_reconfigure`, implementing the behavior explained by @stwirth in https://github.com/intel-ros/realsense/issues/609#issuecomment-464836205. However, this repo has a copy of `ddynamic_reconfigure`, so I'm reluctant of making any changes to that package. I've already contacted @awesomebytes to see if we could implement some of those changes in https://github.com/awesomebytes/ddynamic_reconfigure, and more importantly, if we could add it to the official ROS distro https://github.com/ros/rosdistro/blob/master/melodic/distribution.yaml, so you don't have to copy it here, which makes it very hard to maintain and extend.

The two other commits are optional:

1. https://github.com/intel-ros/realsense/commit/1840901de6944a9c12aaebce8bf5dae8e4b459a4 provides a small change to follow the same code used in `ros::names`

2. https://github.com/intel-ros/realsense/commit/8866e8ac503fb9538c3d29d5f14d433951c081e2 transforms to lower case, which IMHO it's more correct because param names are usually lower case. This will require people to update the names of their params to lower case, but since this was already broken, this is probably a good time to get this done properly. Either way, I made this change in a separate commit so I can easily drop it from this PR if you don't like it.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -235,23 +235,44 @@ void BaseRealSenseNode::registerDynamicOption(ros::NodeHandle& nh, rs2::options 
     for (auto i = 0; i < RS2_OPTION_COUNT; i++)
     {
         rs2_option option = static_cast<rs2_option>(i);
-        std::string option_name = create_graph_resource_name(rs2_option_to_string(option));
+        const std::string option_name(create_graph_resource_name(rs2_option_to_string(option)));
         if (!sensor.supports(option) || sensor.is_option_read_only(option))
         {
             continue;
         }
         if (is_checkbox(sensor, option))
         {
-            ddynrec->add(new DDBool(option_name, i, sensor.get_option_description(option), bool(sensor.get_option(option))));
+            auto option_value = bool(sensor.get_option(option));
+            if (nh1.param(option_name, option_value, option_value))
+            {
+                sensor.set_option(option, option_value);
+            }
+            ddynrec->add(new DDBool(option_name, i, sensor.get_option_description(option), option_value));
             continue;
         }
         std::map<std::string, int> enum_dict = get_enum_method(sensor, option);
         if (enum_dict.empty())
         {
             rs2::option_range op_range = sensor.get_option_range(option);
+            const auto sensor_option_value = sensor.get_option(option);
+            auto option_value = sensor_option_value;
+            if (nh1.param(option_name, option_value, option_value))
+            {
+                if (option_value < op_range.min || op_range.max < option_value)
+                {
+                    ROS_WARN_STREAM("Param '" << nh1.resolveName(option_name) << "' has value " << option
+                            << " outside the range [" << op_range.min << ", " << op_range.max
+                            << "]. Using current sensor value " << sensor_option_value << " instead.");
+                    option_value = sensor_option_value;
+                }
+                else
+                {
+                    sensor.set_option(option, option_value);
+                }
+            }
             if (is_int_option(sensor, option))
             {
-                ddynrec->add(new DDInt(option_name, i, sensor.get_option_description(option), sensor.get_option(option), op_range.min, op_range.max));
+                ddynrec->add(new DDInt(option_name, i, sensor.get_option_description(option), option_value, op_range.min, op_range.max));
             }
             else
             {
@@ -268,24 +289,40 @@ void BaseRealSenseNode::registerDynamicOption(ros::NodeHandle& nh, rs2::options 
                 }
                 else
                 {
-                    ddynrec->add(new DDDouble(option_name, i, sensor.get_option_description(option), sensor.get_option(option), op_range.min, op_range.max));
+                    ddynrec->add(new DDDouble(option_name, i, sensor.get_option_description(option), option_value, op_range.min, op_range.max));
                 }
             }
         }
         else
         {
-            if (int(sensor.get_option(option)) > (int)enum_dict.size())
+            const auto sensor_option_value = sensor.get_option(option);
+            auto option_value = int(sensor_option_value);
+            if (nh1.param(option_name, option_value, option_value))
             {
-                ROS_WARN_STREAM("Option " << option_name << 
-                                " has a value: " << int(sensor.get_option(option)) << 
+                if (option_value > (int)enum_dict.size())
+                {
+                  ROS_WARN_STREAM("Param '" << nh1.resolveName(option_name) << "' has value " << option << " > "
+                                            << enum_dict.size() << ". Using current sensor value "
+                                            << sensor_option_value << " instead.");
+                  option_value = sensor_option_value;
+                }
+                else
+                {
+                    sensor.set_option(option, option_value);
+                }
+            }
+            if (option_value > (int)enum_dict.size())
+            {
+                ROS_WARN_STREAM("Option " << option_name <<
+                                " has a value: " << option_value <<
                                 " which is beyond it's scope: " << enum_dict.size());
-            ROS_DEBUG_STREAM("Add enum: " << rs2_option_to_string(option) << ". value=" << int(sensor.get_option(option)));
+                ROS_DEBUG_STREAM("Add enum: " << option_name << ". value=" << option_value);
                 for (auto item: enum_dict)
                 {
                     ROS_INFO_STREAM("Add item: " << item.first << ":" << item.second); // << ":" << sensor.get_option_description(static_cast<rs2_option>(item.second)));
                 }
             }
-            ddynrec->add(new DDEnum(option_name, i, sensor.get_option_description(option), int(sensor.get_option(option)), enum_dict));
+            ddynrec->add(new DDEnum(option_name, i, sensor.get_option_description(option), option_value, enum_dict));
         }
     }
     ddynrec->start(boost::bind(callback, _1, _2, sensor));
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -217,14 +217,22 @@ std::map<std::string, int> get_enum_method(rs2::options sensor, rs2_option optio
 }
 
 /**
+ * Same as ros::names::isValidCharInName, but re-implemented here because it's not exposed.
+ */
+bool isValidCharInName(char c)
+{
+    return std::isalnum(c) || c == '/' || c == '_';
+}
+
+/**
  * ROS Graph Resource names don't allow spaces and hyphens (see http://wiki.ros.org/Names),
  * so we replace them here with underscores.
  */
 std::string create_graph_resource_name(const std::string &original_name)
 {
   std::string fixed_name = original_name;
-  std::replace(fixed_name.begin(), fixed_name.end(), '-', '_');
-  std::replace(fixed_name.begin(), fixed_name.end(), ' ', '_');
+  std::replace_if(fixed_name.begin(), fixed_name.end(), [](const char c) { return !isValidCharInName(c); },
+                  '_');
   return fixed_name;
 }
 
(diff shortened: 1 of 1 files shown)


stwirth:
Nice! :+1:

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -235,43 +246,80 @@ void BaseRealSenseNode::registerDynamicOption(ros::NodeHandle& nh, rs2::options
...
             }
         }
         else
         {
-            if (int(sensor.get_option(option)) > (int)enum_dict.size())
+            const auto sensor_option_value = sensor.get_option(option);
+            auto option_value = int(sensor_option_value);
+            if (nh1.param(option_name, option_value, option_value))
+            {
+                if (option_value > (int)enum_dict.size())
```
stwirth: A check for `option_value < 0` is missing here.

stwirth: And I think it should be `>=`.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -235,43 +246,80 @@ void BaseRealSenseNode::registerDynamicOption(ros::NodeHandle& nh, rs2::options
...
+                                            << enum_dict.size() << ". Using current sensor value "
+                                            << sensor_option_value << " instead.");
+                  option_value = sensor_option_value;
+                }
+                else
+                {
+                    sensor.set_option(option, option_value);
+                }
+            }
+            if (option_value > (int)enum_dict.size())
```
stwirth: This check does not seem to be needed here (the only possibility of `option_value` to be out of range would be by reading it from the param server - and that already has the check above).

efernandez: @doronhi Added https://github.com/intel-ros/realsense/commit/ccaa46462a02e63033276586b20a71a9188f3fa0#diff-15761e8f7813869ae417a96398deda78R249 in ccaa46462a02e63033276586b20a71a9188f3fa0 , although the commit message doesn't say why. Could you clarify why that check is needed, please?

realsense2_camera/src/base_realsense_node.cpp
efernandez: I copied this check from the original one in https://github.com/intel-ros/realsense/commit/ccaa46462a02e63033276586b20a71a9188f3fa0#diff-15761e8f7813869ae417a96398deda78R249 (introduced in ccaa46462a02e63033276586b20a71a9188f3fa0), but I agree the check looks wrong if we assume the enum starts from `0`.

However, according to https://github.com/intel-ros/realsense/blob/9fe3a44fc5ac82f618f18ffd19e0a7c3c4c2ce97/realsense2_camera/src/base_realsense_node.cpp#L212 the enum values go from `min` to `max` with a certain `step`. I've updated this accordingly and remove the check below, so it's only done when there's a ROS (static) param set, assuming option value retrieved from the sensor are always correct; if they aren't, that should be fixed in the `librealsense2` repo IMHO.

realsense2_camera/src/base_realsense_node.cpp
efernandez: I've removed this validation check, since the option value retrieved from the sensor should always be correct IMHO.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -208,23 +209,53 @@ std::map<std::string, int> get_enum_method(rs2::options sensor, rs2_option optio
     if (is_enum_option(sensor, option))
     {
         rs2::option_range op_range = sensor.get_option_range(option);
-        for (auto val = op_range.min; val <= op_range.max; val += op_range.step)
+        const auto op_range_min = int(op_range.min);
```
efernandez, line 212: Note that `rs2::option_range` is a `struct` with `float` values, so we should cast the `min`, `step`, `max` fields to `int` before looping. Looping with `float` values could give unexpected results due to numerical precision, especially if we accumulate the `step` in the iterator counter.

@doronhi Maybe this is the reason you were checking for the option value for enums in https://github.com/intel-ros/realsense/commit/ccaa46462a02e63033276586b20a71a9188f3fa0#diff-15761e8f7813869ae417a96398deda78R249 :thinking:

RhysMcK:
This looks good! Thanks!

doronhi:
As this is a new interface, could you provide a usage example please?

efernandez:
> face, could you provide a usage example please?

@doronhi I don't understand your comment. I'm not providing a new interface AFAIK, if you mean a C++ interface.

This fixes an issue with the ROS wrapper not reading the ROS params from the ROS param server because `ddynamic_reconfigure` doesn't do that, which broke previous behavior when it switched from `dynamic_reconfigure` to `ddynamic_reconfigure`.

doronhi:
> > face, could you provide a usage example please?
> 
> @doronhi I don't understand your comment. I'm not providing a new interface AFAIK, if you mean a C++ interface.
> 
> This fixes an issue with the ROS wrapper not reading the ROS params from the ROS param server because `ddynamic_reconfigure` doesn't do that, which broke previous behavior when it switched from `dynamic_reconfigure` to `ddynamic_reconfigure`.

Sorry for being unclear.
I started roscore.
Then I tried several combinations of `rosparam set /stereo_module/emitter_enabled` and ended up with the following parameters in my parameter server:
```
doronhi@ubuntu:~$ rosparam list | grep emitter
/camera/realsense2_camera/stereo_module/emitter_enabled
/camera/stereo_module/emitter_enabled
/emitter_enabled
/stereo_module/emitter_enabled
```
Still, running rs_camera.launch after that, the `emitter_enabled` parameter was not changed.
So first, as I don't usually use the parameter server, what is my stupid mistake?
second, I would like to have an example of usage so others will have an easier life.

I think that in issue #609 , @stwirth mentioned that once the dynamic parameter is changed, its value is saved to the parameter server. Does that sound right? Could you add this behavior as well. It would also have made it clearer how to modify the parameters as they would have appeared by `rosparam list` command.

Thanks,

doronhi:
Take a look at [issue#689](https://github.com/intel-ros/realsense/issues/689)
Maybe the part of updating the parameter server is already going to be done there..

efernandez:
@doronhi If #689 gets resolved, then we don't need this PR, because that implementation of `ddynamic_reconfigure` already reads the default values from the ROS param server, e.g. https://github.com/pal-robotics/ddynamic_reconfigure/blob/kinetic-devel/src/ddynamic_reconfigure.cpp#L294

That implementation doesn't call `setParam` AFAIK, so it doesn't set the parameters into the ROS param server on start up or on change. However, IMHO this is good because:
1. Calling `setParam` isn't efficient because it creates a socket
2. You can still check the dynamic values with `rostopic echo` into the `parameter_updates` (see examples below)
3. If we set parameters into the server from this driver, then running it again will produce different initial behaviour. I mean that the driver will start using the last configuration set dynamically, that maybe it wasn't the intention of the user.

Not sure how are you planning to move forward from here, @doronhi , but I suggest the following:
1. Remove `ddynamic_reconfigure` from this repo, which was a bad thing in the first place.
2. Depend on the `ddynamic_reconfigure` officially released, which I guess it's already released after reading the last comments in #689 
3. Update the code to use the interface from the official `ddynamic_reconfigure` pkg.

All that can be done in a single PR. I can contribute it, if that sounds good to you.

Answering your question, this should work:
1. `roscore`
2. `rosparam set /camera/stereo_module/emitter_enabled false`
3. `roslaunch realsense2_camera rs_camera.launch`
4. `rostopic echo /camera/stereo_module/parameter_updates -n1 | grep -A 1 emitter_enabled | grep value | awk '{print $2}'` will show `False`

After this state, you will also have this behaviour:
1. Stop the camera driver
2. `rosparam delete /camera/stereo_module/emitter_enabled`
3. `roslaunch realsense2_camera rs_camera.launch`
4. `rostopic echo /camera/stereo_module/parameter_updates -n1 | grep -A 1 emitter_enabled | grep value | awk '{print $2}'` will show `False`, because now the default is the value of the option in the camera driver/sensor. If you unplug the camera and plug it in again, this might change.

And also:
1. Stop the camera driver
2. `rosparam set /camera/stereo_module/emitter_enabled true`
3. `roslaunch realsense2_camera rs_camera.launch`
4. `rostopic echo /camera/stereo_module/parameter_updates -n1 | grep -A 1 emitter_enabled | grep value | awk '{print $2}'` will show `True`

All that would be the expected behaviour of the official `ddynamic_reconfigure` as well, so I guess nothing should be documented here.

I confirm that behaviour with this PR. It should be the same if you switch to the official `ddynamic_reconfigure`, which long-term is the correct thing to do IMHO. However, I'd appreciate if you merge this PR first, so #609 gets fixed.

doronhi:
I merged the PR.
I understand what you said about the shortcomings of writing values back to the parameter server.
- @stwirth ,I would appreciate your opinion as it differs from what I understood in [#609](https://github.com/intel-ros/realsense/issues/609) to be your needs.
- @v-lopez, what do you think is going to be implemented inside ddynamic_reconfigure? Will it write values to the parameter server or not?

@efernandez , I would appreciate your help in updating to [ddynamic_reconfigure 0.1.7](https://github.com/pal-robotics/ddynamic_reconfigure/commit/08cf7eb6d4eca35c8f34ae95221f1d2b3b7e4c6b).
However, the reason I chose to copy the code was to avoid using sub-module which made life for realsense users harder.
I hope in the very near future to create a debian package for realsense2_camera. When it happens, it would be better to include ddynamic_reconfigure as sub-module as most users will not download the code anymore.
Until then, and until ddynamic-reconfigure itself is released as a debian, I would prefer to keep an updated copy.

v-lopez:
The current WIP implementation is here: https://github.com/pal-robotics/ddynamic_reconfigure/tree/refactor-functions

And the migrated code for realsense here: https://github.com/pal-robotics-forks/realsense

We'll validate today with our cameras and then we'll submit for your evaluation.
In the meantime I'll be extending the tests and adding DDynamic reconfigurable strings.

Also I've requested 0.1.7 to be released on kinetic, but you'll be using 0.2.0 or later, which will include the changes I'm making so you can provide callbacks instead of pointers to the variables you are reconfiguring.

efernandez:
Thanks @doronhi 

It looks like @v-lopez is already doing the hard work to bring us to a position where we can nuke `ddynamic_reconfigure` from this repo. :smiley: 

I'm going to wait for his tests and for things to be released (I think there's a release sync tomorrow EOD). After that, if there's anything missed, I can help.

Maybe we can continue the discussion on #689

stwirth:
@doronhi I'm happy with the way this is going.

PR #673, opened 2019-03-13
Export dynamic_reconfigure to dependent packages

efernandez:


doronhi:
Could you please elaborate why this is needed? What is the use case?

stwirth:
Here's an explanation of what `CATKIN_DEPENDS` does: https://answers.ros.org/question/58498/what-is-the-purpose-of-catkin_depends/

ddynamic_reconfigure does not encapsulate dynamic_reconfigure entirely, `#include <ddynamic_reconfigure/ddynamic_reconfigure.h>` pulls in `#include <dynamic_reconfigure/server.h>`.
The `CATKIN_DEPENDS` tells packages that `find_package(catkin COMPONENTS ddynamic_reconfigure)` to pull include dirs and libraries from dynamic_reconfigure as well.

efernandez:
@doronhi As @stwirth already said, we need this because `ddynamic_reconfigure` uses `dynamic_reconfigure` on its headers, so client libraries/pkgs require that dependency exported in order to build properly.

This is an issue if you do an isolated build.

Thanks for merging this!

PR #676, opened 2019-03-14
fix Twist in odometry problem

akirayou:
1.Covalience of Twist in odometry have inf. 
  Because pose.tracker_confidence is unsidned.  you have to cast signed. (L1190)

2.Coordinates of Twist of odometry must be based on chilid_frame 
  " The twist in this message corresponds to the robot's velocity in the child frame, normally the coordinate frame of the mobile base,"
  Please see  http://wiki.ros.org/navigation/Tutorials/RobotSetup/Odom

3.In correct covariance setting?
May be cov_twist means rotation, cov_pose means linear movement.
So, fllowing covariance is better.

But,there is another solution. Don't use cov_pose,and cov_twist. 
On my experience on T265, the real covariance is not on simple rule  such as pow(10),
I want to set manualy for each tracker_confidence . I dont't know which you like.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1187,18 +1187,27 @@ void BaseRealSenseNode::pose_callback(rs2::frame frame)
 
     if (0 != _imu_publishers[stream_index].getNumSubscribers())
     {
-        double cov_pose(_linear_accel_cov * pow(10, 3-pose.tracker_confidence));
-        double cov_twist(_angular_velocity_cov * pow(10, 1-pose.tracker_confidence));
+        double cov_pose(_linear_accel_cov * pow(10, 3-(int)pose.tracker_confidence));
+        double cov_twist(_angular_velocity_cov * pow(10, 1-(int)pose.tracker_confidence));
 
         geometry_msgs::Vector3Stamped v_msg;
         v_msg.vector.x = -pose.velocity.z;
         v_msg.vector.y = -pose.velocity.x;
         v_msg.vector.z = pose.velocity.y;
-
+	tf::Vector3 tfv;
+	tf::vector3MsgToTF(v_msg.vector,tfv);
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Its highly unlikely your covariances for all of these parameters will be the same, shouldnt the entire matrix be configurable?

akirayou:
I think user setting covariances (each element of diagonal vector)  which depends on tracker_confidence.
Because setting correct covariance is little bit difficult. So leaving it to user is easy workaround.

About pose covariance:
In case of tracker_condidence==3, T265 have very little covariance ( depends on distance of subject,Number of subjects...But I don't know how to get the information.)
In case of tracker_condidence!=2 , pose will have very very big covariance in sevral seconds.
After that covariance keeps high even if  tracker_confidence==3 again.(without relocalization)
It also depends on relocalization mode or not. After the relocalization,covariance gets very small again.
But, I have no idea to detect relocalization was done in successful.
Any way user can assume reasonable covariance experimentarly  for there site.

About  twist covariance:
In case of tracker_confidence==3,very little covariance,and there is no relocalization problem,
Consitering no feature image (such as white wall),tracker_condifence  will be 2or1.
IMU is only way to know the velocity. Twist covariance can be assumed by IMU covariance.
T265 have big linear drift,But have small rotation drift. 
So, at least covariance of rotation and linear motion must be set individually.

I think setting 4 covariance which is conbination of  (linear motion,rotation)*(pose,twist) for each tracker_confidence(4state)  [ i.e. 4*4=16 variable] is reasonable workaround.

But it just workaround. May be there is better way.

doronhi:
Thank you. First of all for the code corrections which I will merge promptly.
Regarding the values themselves, I agree that such simple rule such as pow(10) is not the solution and moreover, looking at your corrections now, it is clear to me that no way the cov for pose is the same as cov for linear velocity, so that also needs to be fixed.
I believe that the covariance should come from the device. I feel that it is not the user job to measure it nor does he have all the information. If these values will not become available in the near future it will inevitably will be added to the user parameters.

Issue #494, opened 2018-10-16
Split description from ROS driver


wxmerkt:
Hi,
We have many use cases where we'd just like to have the description (xacro/URDF/mesh) on a machine and not have the requirement to install the SDK and compile the driver. Thus it would be great to create (and binary-release) a ``realsense_description`` package which includes the xacro, URDF, visualisation-launch file, and the meshes. Plus, the current DAE is 15 MB (!) - a watertight and shrunk smaller mesh would be better.

On our end, we'd be happy to do the work towards creating a ``realsense_description`` package and binary-release it if there is interest from the community.

Thanks,
Wolfgang

gavanderhoorn:
> Plus, the current DAE is 15 MB (!) - a watertight and shrunk smaller mesh would be better.

Indeed.

I like a bit of detail as much as @wxmerkt probably, but 15MB for a camera mesh? :)

AustinDeric:
In addition, i have a few observations/requests that would be a part of this issue.
-  there is not ${prefix} in the xacro, which makes running 2 cameras in the same urdf impossible (http://wiki.ros.org/urdf/Tutorials/Using%20Xacro%20to%20Clean%20Up%20a%20URDF%20File)
- The xacros are not available to the users of the binary because the urdf directory is not installed: https://github.com/plusone-robotics/realsense/blob/development/realsense2_camera/CMakeLists.txt#L120.  Was there a reason for this?

I am going to take a shot at this, unless anyone else has made progress?

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

athackst:
I'm also interested in this.  We've internally forked this and made changes for our system (including adding gazebo tags).  @AustinDeric @wxmerkt  I'm happy to contribute if you've already started something

wxmerkt:
I'd like to see some input and feedback from the @intel-ros maintainers of this driver, e.g., @doronhi - we seem to have strong community interest in improvements to the description and splitting it out/releasing it to the ROS buildfarm. Can you please comment/assist? Thank you!

jarvisschultz:
I also just finished adding Gazebo tags and prefix support to the xacro files for an internal project here. Seems like there is interest in making these changes. Happy to support any PRs if they'd be accepted.

athackst:
Kk, I made a minimal change set PR so we can separate the reorg from the URDF updates.

PR #690, opened 2019-03-20
Feature/realsense2 description

athackst:
Minimal change set to move realsense urdfs and related files to their own package, realsense2_description. 

This PR is the first step to address issue #494


realsense2_description/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+# RealSense description
+catkin_package(
+    )
+
+# Install launch files
+install(DIRECTORY launch/
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/launch
+    )
```
gavanderhoorn: All of these directories could be installed in a single `install(..)` block:

```cmake
install(DIRECTORY launch meshes rviz urdf
  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
```

athackst: :+1: Done.

PR #691, opened 2019-03-21
wheel_odometry

doronhi:
Integrating robot's odometry into T265.
rs_camera.launch include 2 more parameters that needs to be filled to use this option:
topic_odom_in - which topic to listen to. Only using `twist.twist.linear`
calib_odom_file - calibration file. Format is described in [merged PR#3462](https://github.com/IntelRealSense/librealsense/pull/3462#issuecomment-472491730)

### DIFF
--- a/realsense2_camera/src/t265_realsense_node.cpp
+++ b/realsense2_camera/src/t265_realsense_node.cpp
@@ -6,7 +6,67 @@ T265RealsenseNode::T265RealsenseNode(ros::NodeHandle& nodeHandle,
                                      ros::NodeHandle& privateNodeHandle,
                                      rs2::device dev,
                                      const std::string& serial_no) : 
-                                     BaseRealSenseNode(nodeHandle, privateNodeHandle, dev, serial_no) {}
+                                     BaseRealSenseNode(nodeHandle, privateNodeHandle, dev, serial_no),
+                                     _wo_snr(dev.first<rs2::wheel_odometer>()),
+                                     _use_odom_in(false) 
+                                     {
+                                         initializeOdometryInput();
+                                     }
+
+void T265RealsenseNode::initializeOdometryInput()
+{
+    std::string calib_odom_file;
+    _pnh.param("calib_odom_file", calib_odom_file, std::string(""));
+    if (calib_odom_file.empty())
+    {
+        ROS_INFO("No calib_odom_file. No input odometry accepted.");
+        return;
+    }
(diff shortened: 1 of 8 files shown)


skohlbr:
Will odometry also work for robots that undergo significant motion in rough terrain and/or exhibit significant noise in their odometry estimate? Or is there an implicit assumption that the robot is moving on flat ground? I'm thinking about tracked vehicles moving outdoors here.

realsense2_camera/src/t265_realsense_node.cpp
```diff
@@ -6,7 +6,67 @@ T265RealsenseNode::T265RealsenseNode(ros::NodeHandle& nodeHandle,
...
+    _pnh.param("topic_odom_in", topic_odom_in, DEFAULT_TOPIC_ODOM_IN);
+    ROS_INFO_STREAM("Subscribing to in_odom topic: " << topic_odom_in);
+
+    _odom_subscriber = _node_handle.subscribe(topic_odom_in, 1, &T265RealsenseNode::odom_in_callback, this);
+}
+
+void T265RealsenseNode::odom_in_callback(const nav_msgs::Odometry::ConstPtr& msg)
+{
+    ROS_INFO("Got in_odom message");
+    rs2_vector velocity {-(float)(msg->twist.twist.linear.y),
```
schmidtp1, line 63: No need to apply a coordinate transformation here. The relative transformation between odometry frame and T265 body frame is handled in the calibration.

doronhi: I want to keep the calibration.json file in a ROS standard. That is, X-forward, Y-Left and Z-Up.
For that reason it is necessary to convert from ROS coordinate system to T265's.

sadransh:
@skohlbr I have the same question. Did you find out anything new?

schmidtp1:
@skohlbr, @sadransh, there is no implicit assumption of the robot moving on a flat ground YET if only one component of the 3D velocity measurement vector is provided and the other two are set to zero, these act as pseudo measurements and enforce a "soft" constraint on the instantaneous velocity in these directions to be zero, i.e. that robot is moving on a (smooth) surface and cannot go up/down or left/right. It doesn't necessarily have to be flat and would work when driving on ramps for example or for rotation when driving over bumps. (This is important to keep track of visual features.)
The measurement noise covariance parameter allows to account for (white) noise.

PR #696, opened 2019-03-25
add flag publish_odom_tf (default to true)

doronhi:


pavloblindnology:
Yes. This certainly helps.
To make it even better, a 2nd set of `odom_frame_id` and `pose_frame_id` is needed - smth like `tf_odom_frame_id` and `tf_pose_frame_id` (with default values identical to `odom_frame_id` and `pose_frame_id`). So, that the next TF links are published: `tf_odom_frame_id` -> `tf_pose_frame_id`, `pose_frame_id` -> `base_frame_id`. And odometry message published with `pose_frame_id` and `base_frame_id`.
This way we can use odometry message and `pose_frame_id` sub-tree for robot_localization, etc, and `tf_odom_frame_id` -> `tf_pose_frame_id` TF link for smth else (trajectory building, etc.).
Or, by setting `tf_pose_frame_id` = `pose_frame_id`, we can use it as before.

doronhi:
I merge this PR.
Add launch param: 
* publish_odom_tf : Defaults to True. If True publish TF from odom_frame to pose_frame.

About the second suggestion, I need more time to get to the bottom of the need and I don't want to postpone the current addition.

PR #702, opened 2019-03-26
Add option to use external nodelet manager, Fixes #697

### LINKED ISSUE #697: Not able to pass external nodelet manager
In my setup I have a central nodelet manager in which several nodes run in. I'd like to pass that nodelet manager to the realsense node. I tried using the manager argument in the launch file but it seems that that argument is not supposed to be used that way. The issue is that the realsense launch file always creates a new nodelet manager for the node to run in. (https://github.com/intel-ros/realsense/blob/development/realsense2_camera/launch/includes/nodelet.launch.xml#L82)

sundermann:
Right now a nodelet manager is always created when running the realsense
node. This is not always a wanted behaviour as a nodelet manager might
already be running somewhere else in the ROS system. This patch adds a
parameter to disable launching a new nodelet manager. In this case the
nodelet manager must be provided using the manager arg. The new
parameter is set to false by default, so existing behaviour is not
changed.

Fixes #697

### DIFF
--- a/realsense2_camera/launch/includes/nodelet.launch.xml
+++ b/realsense2_camera/launch/includes/nodelet.launch.xml
@@ -1,4 +1,5 @@
 <launch>
+  <arg name="external_manager"    default="false"/>
   <arg name="manager"             default="realsense2_camera_manager"/>
   <arg name="serial_no"           default=""/>
   <arg name="tf_prefix"           default=""/>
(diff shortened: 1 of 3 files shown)


sundermann:
Any updates on this?

doronhi:
I appreciate your effort, willingness to share and caring to improve. Unfortunately it might take a while to address this issue so I can't promise a time frame right now.
I can promise you that I will be in touch and let you know when things move and maybe ask you for further clarifications.
Sorry for now and thanks you.

sundermann:
No worries, I'm happy to answer any questions with this PR that might occur!

sundermann:
Any news on this?

doronhi:
Sorry. Not yet..

sundermann:
Any updates? This is a very simple change and does not touch any existing behaviour.

sundermann:
Thanks!

PR #710, opened 2019-04-02
Better timestamp correction

akirayou:
1. Recent librealsense(on T265) returns timestamp of system time. 
 This is the change of librealsense.
https://github.com/IntelRealSense/librealsense/pull/3609/commits/5a6e1c4b691e1e142932f475c40c40a64db7b5e0

   So I got the WRAN message "Frame metadata isn't available!".
   In librealsense,You can get frame time by RS2_FRAME_METADATA_FRAME_TIMESTAMP.
   In "frame_time()" , I try to get FREAME_TIME,and use get_timestamp() as backup method.
   latest lirealsense dose not have the metadata yet , but you can get it by small patch.
https://github.com/akirayou/librealsense/commit/ecbcb32358dda9535ccd9e72146ea7178311b1ed

2.In case of T265 device clock seems to be  already corrected by librealsense.
 Additional time correction is bad idea.it makes more system latency. So I want to skip the time correction. 
 Please see: time correction part of librealsense(for T265)
  https://github.com/IntelRealSense/librealsense/blob/1148c9b4d50a653119fbb666e52acb8b850bb18c/third-party/libtm/libtm/src/Device.cpp#L230-L274

3.For other device (which have there own device timestamp)
Device's clock have slightly different speed.So periodic time correction is needed.
But in original implementation ,setBaseTime is called only one time. 
So I wrote continuous time correction with IIR filter in  "timestampFromFrame()"

4.To make it stable.
Original implimentation of  realsense_node_factory.cpp
Change_device_callback is enabled before StartDevice().
Sometimes it makes twice calling of StartDevice().
It makes my code unstable.
I workaround this( change_device_callback is enabled after the StartDevice).

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -193,7 +193,7 @@ namespace realsense2_camera
         double timestampFromFrame(const rs2::frame &frame);
 	double elapsedTimeFromFrame(const rs2::frame &frame);
 	double frameTime(const rs2::frame &frame);
-        void fix_depth_scale(rs2::depth_frame depth_frame);
+        cv::Mat& fix_depth_scale(const cv::Mat& from_image, cv::Mat& to_image);
         void clip_depth(rs2::depth_frame depth_frame, float clipping_dist);
         void updateStreamCalibData(const rs2::video_stream_profile& video_profile);
         void publishStaticTransforms();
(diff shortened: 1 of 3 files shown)


doronhi:
@akirayou ,
Thank you so much for your time and work.
latest librealsense 2.22.0 include time correction for D400 series. It is based on estimation using a linear equation as it adapts better for cases where sometimes the device clock is faster and sometimes slower then the system time.
For running on a single machine, no further correction is needed I think. 
For running on a multiple machines environment, a similar correction to ROS time should be done in the wrapper but I think it's best to do it in a similar way, using linear equation rather then IIR filter.

Regarding the stability problem, I totally agree and incorporated your idea into #775.
Therefor, this PR can be closed.

PR #718, opened 2019-04-05
Added missing dependency on rgbd_launch

peci1:


doronhi:
rgbd_launch is not mandatory for running the package. It is merely needed for running `rs_rgbd.launch` which is an example launch file.
You can have the pointcloud directly using filters:=pointcloud.

peci1:
Then it'd maybe deserve at least a comment in the launch file?

doronhi:
Agreed.

doronhi:
Done directly to development branch.
Thank you for helping to clarify.

PR #759, opened 2019-05-07
TF Denormalized Quaternion Fix

manomitbal:
When transforming and utilizing poses from the T265 camera frame, a TF_DENORMALIZED_QUATERNION error shows up that ends up ignoring transforms based on TF_NAN_INPUT. This fix normalizes the quaternion before publishing.

### DIFF
--- a/realsense2_camera/src/t265_realsense_node.cpp
+++ b/realsense2_camera/src/t265_realsense_node.cpp
@@ -103,18 +103,18 @@ void T265RealsenseNode::calcAndPublishStaticTransform(const stream_index_pair& s
         Q = Q.inverse();
         quaternion_optical = quaternion_optical.inverse();
 
-        publish_static_tf(transform_ts_, trans, Q, _frame_id[stream], _base_frame_id);
+        publish_static_tf(transform_ts_, trans, Q.normalize(), _frame_id[stream], _base_frame_id);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


rongoldman-tri:
We are seeing a similar denormalized quaternion error for one of our D415 realsense cameras. The problem appears to be that when getting the extrinsics the returned 3x3 rotation matrix is all zeros (line 83) and so when it is used to create the quaternion (line 98): rotationMatrixToQuaternion(ex.rotation) the resulting quaternion is bad.

You might want to check if the rotation matrix is all zero, i.e. ex.rotation[i] == 0 for 0 <= i < 9 and if so reset it to the identity as done on line 90.

The root cause might be from having saved a bad calibration result to the camera.

-- Ron --

doronhi:
@manomitbal , Is the error you see related to the zeros rotation matrix as described by @rongoldman-tri ?
Assuming the transformation is not available, how can you use the IMU and Pose data?
Although your fix do eliminate the error notification I'm not sure it's a good idea as the user will receive the wrong transformation without any indication.
On that note, setting a unity transformation if non is available from the device seems like a bad idea and should be removed.
What do you think?

manomitbal:
@doronhi @rongoldman-tri You were right, I was just simply masking the problem, this still persisted regardless. I think one of the latest patches fixed this issue but I will keep a lookout.

doronhi:
Glad to hear.
Can this issue be closed then?

PR #764, opened 2019-05-10
3D printed mount for T265 + D435

schmidtp1:
- added STL
- added corresponding calibration (extrinsics) in form of URDF
- set as default in rs_d400_and_t265.launch
- added rviz configuration to display colored point cloud with respect to static reference frame

withMount-angle

To use with precalibrated extrinsics, please make sure to slide in T265 all the way and to tighten the clamping mechanism. D435 is fixed against rotation with 2 pins (on top). It uses a **flat head phillips screw 1/4 in - 20 x 1 - 1/2 in** (see below).
mount-frontG
Two **round head combo 1/4 in - 20 x 1 in** can be used to mount it to a back plate.


schmidtp1:
trigger new build

schmidtp1:
@doronhi, what do you think, can we add this? Is it ok to set it as default or would you prefer not to?

doronhi:
Hi @schmidtp1 ,
I'm in the process of printing the model you added so I could try it. A mount to combine D435 and T265 is very useful.
It seems to me, from looking at the model (not printed yet..) that the D435 and T265 look in opposite directions. Is that true?
I think, however, that the STL file itself is a realsense model and should be placed in a more general location, not in the ROS wrapper repository.

schmidtp1:
@doronhi, thanks for looking into this.
D435 and T265 are actually looking in the same direction. I will share some pictures of the assembly.
Can you propose a better location?
Thanks.

antoan:
Pardon my ignorance, but doesn't the spec of "a flat head phillips screw 1/4 in - 20 x 1 - 1/2 in"
imply 1/2 in screw length, in which case I don't see how in could reach the top of the mount, from the bottom as shown in the image above.  It seems way too short?
 
The height of my printed mount is about 1 1/4 in and fits my T265 perfectly. Am I missing something?
(I am based in the UK) I would really appreciate the screw spec equivalent in metric units?

schmidtp1:
@antoan, I think that means **1 1/2 in** the way it is stated. The closest metric should be 40mm. Let me double check and measure tomorrow.

In the mean time, there is also a new version of the mount that you can try below!

schmidtp1:
[bracket_t265nd435_external.zip](https://github.com/IntelRealSense/realsense-ros/files/3346811/bracket_t265nd435_external.zip)

**Instructions:**
- Use M3 screws, 2x 10mm (for D435) & 2x 18mm (for T265), to mount the cameras
- Insert ¼-20 _insert_ nut (for example https://www.mcmaster.com/92397A116) to mount on tripod / use with standard camera mount
(**_Edit_**: One way to insert the nut is to heat it with a soldering iron or hot glue gun and press it into the 3D print.)

snapshot00

antoan:
@schmidtp1 Thanks for clarifying. I tried it with a M6 x 40mm without any luck. 

Regarding the new mount, can I use the same extrinsics/calibration xacro?
https://github.com/IntelRealSense/realsense-ros/blob/bb5c713d57c41c30609862591d778f4c33abd695/realsense2_camera/urdf/mount_t265_d435.urdf.xacro

Cheers

schmidtp1:
@antoan, sorry for the delay. The good news is that the new mount has the same extrinsics :) Yes, you can use the same urdf/xacro.

I measured the screw and the length is around **32mm without head and 37mm with head**. I'm using this one for example: https://www.amazon.com/gp/product/B07C6DHPXL/ref=ox_sc_saved_title_7?smid=A2I6Z759BTJP4J&psc=1

Cheers

antoan:
@schmidtp1 Really appreciate it. 
Will give the 2nd mount a try, too and report back.

zlite:
Philip, could post the STEP file for that new bracket, too, so we can modify its base to fit different vehicles?

schmidtp1:
@zlite, Chris, sorry for the delay. Please see step file below:
[bracket_t265nd435_external.stp.zip](https://github.com/IntelRealSense/realsense-ros/files/3419331/bracket_t265nd435_external.stp.zip)

antoan:
@schmidtp1

I have the new mount with me and just ordered the screws for it.  I just need a link for the 1/4-20 nut you used and I'll be all set, if you don't mind.

Cheers

schmidtp1:
https://www.mcmaster.com/92397A116 (for example)

You can insert the nut for example by heating it with a soldering iron or hot glue gun and pressing it into the 3D print.

doronhi:
The original mount is now available [here](https://dev.intelrealsense.com/docs/2d-occupancy-map). It's a more general place then the ROS-wrapper repository.
Does the second model have advantages over the first model? Should we also pursue publishing it instead or as an additional model in the https://dev.intelrealsense.com/docs pages?

schmidtp1:
@doronhi, the second model can be mounted on a tripod while the first one was designed for / adapted to the robot. Yes, it would be helpful and make it easier to find if we could publish it there as additional model.

palwanishan:
Hi @schmidtp1, I appreciate your contributions. I would like to request for 3D printable "T265 + D455 mount" and ROS tf, please. Thanks for your time and attention!

schmidtp1:
@palwanishan, thanks! @doronhi, please note the request above.

SamerKhshiboun:
Closing this PR since T265 is EOL, in addition to that we are developing ROS2 branches only.
If any user find this useful, he can clone the files and use them.

Issue #649, opened 2019-03-04
Wrong frame for Gyro and Accel


pavloblindnology:
Imu messages should be reported in body frame, not in optical, as it's done in synchronized mode.
https://github.com/intel-ros/realsense/blob/36dd73aae6bc1b654ad085393eac279339d7f5a0/realsense2_camera/src/base_realsense_node.cpp#L1115

RealSenseCustomerSupport:
[Please Ignore - RealSense system comment]

doronhi:
Add PR: https://github.com/IntelRealSense/realsense-ros/pull/766

PR #766, opened 2019-05-12
Imu frame_id

doronhi:
Change frame_id for imu messages to camera_link's coordinates system. Same as is being done for imu's sync messages.


doronhi:
Considering that every sensor may have it's own coordinate system, it makes more sense to send the information in the original coordinate system and match the correct frame_id.
Therefor, #778 will solve #649 and this PR  is closed.

PR #775, opened 2019-05-20
fixed t265 conflict

doronhi:
set_devices_changed_callback called AFTER getDevice()
Keep checking for devices until device is found - for cases where T265 was momentarily taken by another node at the time of query.

Add a 3rd, optional camera, to rs_multiple_devices.launch file.


doronhi:
closed as it is (accidentally) included in #778

PR #778, opened 2019-05-20
Fix t265_link to match ROS coordinate system

doronhi:
Fix t265_link to match pose_frame and to follow ROS coordinate system.

### DIFF
--- a/realsense2_camera/src/t265_realsense_node.cpp
+++ b/realsense2_camera/src/t265_realsense_node.cpp
@@ -72,7 +72,7 @@ void T265RealsenseNode::calcAndPublishStaticTransform(const stream_index_pair& s
 {
     // Transform base to stream
     tf::Quaternion quaternion_optical;
-    quaternion_optical.setRPY(M_PI / 2, 0.0, -M_PI / 2);
+    quaternion_optical.setRPY(M_PI / 2, 0.0, -M_PI / 2);    //Pose To ROS
     float3 zero_trans{0, 0, 0};
 
     ros::Time transform_ts_ = ros::Time::now();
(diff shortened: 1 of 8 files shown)


realsense2_camera/src/t265_realsense_node.cpp
```diff
@@ -72,7 +72,7 @@ void T265RealsenseNode::calcAndPublishStaticTransform(const stream_index_pair& s
 {
     // Transform base to stream
     tf::Quaternion quaternion_optical;
-    quaternion_optical.setRPY(M_PI / 2, 0.0, -M_PI / 2);
+    quaternion_optical.setRPY(M_PI / 2, 0.0, -M_PI / 2);    //Pose To ROS
```
msadowski, line 75: @doronhi should the transform here be: `setRPY(-M_PI / 2, M_PI / 2, 0);` ?

According to the [docs](http://docs.ros.org/lunar/api/tf/html/c++/classtf_1_1Quaternion.html#ab34dfef6a01c1ece07bea92d485b0576) setRPY uses a fixed transform. 

I created a short  Python to test this transform by creating: world->t265_pose->ros_frame transforms and visualizing them in rviz. Here is the resultant transform for setRPY(M_PI / 2, 0.0, -M_PI / 2); equivalent:

image
The frames in order from the left are: world_frame->t265_pose_frame->ros_frame.

Here is the gist of the python script that I used: https://gist.github.com/msadowski/caa15bf42bbcc46979eac1afe9f8876b. 

By changing the transform to setRPY(-M_PI / 2, M_PI / 2, 0) (line 38 of my gist) I'm able to receive this result:
image
(please ignore the translations) which looks correct to me.

Issue #781, opened 2019-05-21
Remove ddynamic_reconfigure in favor of the APT released one


veimox:
Following the [comment](https://github.com/IntelRealSense/realsense-ros/issues/689#issuecomment-494231581) of issue #689, I create this issue to remove the dependency from the repo

Issue #780, opened 2019-05-21
Install librealsense-dkms and -dev from rosdep


veimox:
At Blue Ocean Robotics we use rosdep to install `librealsense-dkms` and `librealsense-dev`. I think you might also want to use this method too instead of making it manual. Instead of doing the PR in rosdistro i think you should take ownership of this. You are more than welcome to either copy-paste our script or directly point to it.

The steps to be done is to create an entry in `base.yaml` in [rosdistro](https://github.com/ros-infrastructure/rosdistro) with the content:

```yaml
librealsense2:
  ubuntu:
    xenial:
      source:
        uri: "https://bitbucket.org/api/2.0/snippets/blue-ocean-robotics/aendn6/master/files/librealsense2.rdmanifest"
```

We only target Ubuntu Xenial, but you need to adapt the entry for your use case. Then, when the entry `librealsense2` is found in the package.xml, that script will be run allowing thus to install the libraries. 

As a note, in the script of the uri there *must* exist this uri that points to a tar file. This can be empty as the one used (that is actually provided by magazino)

This is issue is aligned with #704 and is a pre-step before blooming because AFAIK it is a requirement that a ros package installs its own dependencies automatically.

doronhi:
Thanks for the info!

doronhi:
@veimox ,
Started following your suggestion. Proposed [PR21430](https://github.com/ros/rosdistro/pull/21430)

PR #693, opened 2019-03-21
Use PAL Robotics ddynamic_reconfigure 0.2.0

v-lopez:
ddynamic_reconfigure is being released here: https://github.com/ros/rosdistro/pull/20651

I've only been able to test it with our packages and with an Intel RealSense T265, and I was able to change the frames_queue_size param.

PR #782, opened 2019-05-21
Remove ddynamic reconfigure in favor of the APT released one

veimox:
This PR does three things:

1. Removes the forked ddynamic_reconfigure
2. Moves the package to root of the repo and updates files to use the new paths (.travis.yml)
3. Adapts the the ddynamic_reconfigure code to the new API of the released package

The review is better to be done per commit rather than the whole PR because there are many files moved.

This tackles #781. This is also will simplify the blooming process along with #780.

I have tested this with a D435 and a D435i.


doronhi:
I want very much to thank you for the amount of time and effort you put into this PR.
I ended up merging #693 as it is very close to yours and I preferred to keep the directory realsense2_camera in place as it fit with the idea of separating the package to two, as is done in #690 .
Again, thank you.

PR #795, opened 2019-05-31
Added ddynamic_reconfigure installation note

mayio:


Karsten1987:
this is definitely necessary. Installation as such is highly confusing, please rosdep update for ddynamic_reconfigure does not yield anything for melodic.

PR #804, opened 2019-06-05
enable required tag when launching node

baumanta:
This PR adds support for the ROS functionality of setting a node to be required. If the realsense node is set to be required, all nodes launched in same file will die if the RS node dies.

This is desirable if another node e.g. a planner depends on the realsense node.

### DIFF
--- a/realsense2_camera/launch/includes/nodelet.launch.xml
+++ b/realsense2_camera/launch/includes/nodelet.launch.xml
@@ -5,6 +5,7 @@
   <arg name="tf_prefix"           default=""/>
   <arg name="json_file_path"      default=""/>
   <arg name="rosbag_filename"     default=""/>
+  <arg name="required"            default="false"/>
 
   <arg name="fisheye_width"       default="0"/>
   <arg name="fisheye_height"      default="0"/>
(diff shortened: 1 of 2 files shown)


baumanta:
@doronhi Is that something you would be interested in merging? We are using it on our systems where the realsense supplies the data to other alogrithms, which depend on it. I figured maybe others could use it as well as with camera data this is quite often the case.

jkflying:
Note, this is an optional parameter with default set to false, so it doesn't affect anybody unless they set it to true.

doronhi:
Thank you for your ideas.
I don't really feel like the launch files are part of the package. They feel to me more like some kind of templates, users can change as they see fit.
As such, I consider your modification great as a ROS option but not necessarily as integral part of this package.
I do think about adding an examples directory for launch files which will include also methods for modifying parameters connected to dynamic_reconfigure package. There are many questions regarding what can be done with the launch file. But I haven't thought of the right way and location yet. 
When I do, your launch file will be very welcome there.

Confession: I don't always manage to follow this principle but I try. I specifically retain this PR to be handled later.

baumanta:
@doronhi I understand your reasoning. This might be correct for developers who just run experiment with the package on their laptop and can make changes manually. We are operating in a more production like fashion and we do not want to add a realsense launch file in another completely unrelated repo. I think it is important that the launch file is the interface to the package and contains standard ros functionalities such as 'required'. Right now we cannot use devel branch anymore as we always have to cherry-pick this commit on top. It would make our workflow a lot easier if we could merge this.

doronhi:
OK :)

PR #806, opened 2019-06-07
Use rosdep install to install dependencies

furushchev:


doronhi:
realsense2_camera depends on librealsense2 which is not yet available as a ros package.
Until that time, the .travis check in this PR is bound to fail. and therefor, there is no point, I think, in merging it. 

Until that time, librealsense2 needs to be installed from the PPA at  http://realsense-hw-public.s3.amazonaws.com/Debian/apt-repo
as described in the installation guide of librealsense2.

furushchev:
@doronhi Thank you for the comment.
I know that we need the installation process of `librealsense2` using apt beforehand.
The problem is that a dependency in `package.xml` does make sense only when it is registered in `rosdistro` ( https://github.com/ros/rosdistro ) and is used for installing dependencies automatically (using `rosdep install` command)
Since `librealsense2` is not yet registered there, `rosdep install` command fails on this package and all the packages which depend on this package, and also I think the `rosdistro` ecosystem does not support installing debian files from custom apt repositories. 
I'm interested in why it is necessary to be listed in `package.xml`?

doronhi:
As for now it not necessary for librealsense2 to be listed in package.xml.
I added it as part of my attempts to add realsense2_camera and librealsense2 to rosdistro. See [Add installation of librealsense2](https://github.com/ros/rosdistro/pull/21430) - based on #780, and [realsense2_camera for kinetic](https://github.com/ros/rosdistro/pull/21502)

The first one, as you may see, it's not going so well so either we'll find a solution shortly or I'll remove the dependency from the package.xml file. Again (it was done about 10 month ago).
Since `rosdep install` doesn't work either with librealsense2 dependency in the package.xml or without it, I hope it doesn't do too much damage while it's still there...

fitter22:
Any update on that issue?
> Since `rosdep install` doesn't work either with librealsense2 dependency in the package.xml or without it, I hope it doesn't do too much damage while it's still there...

Are you sure? I managed to run **rosdep install** successfully after I deleted librealsense2 dependency from package.xml.

furushchev:
Sorry for late reply.

@doronhi 

> Since rosdep install doesn't work either with librealsense2 dependency in the package.xml or without it, I hope it doesn't do too much damage while it's still there...

The testing of this PR at least fails on the `rosdep install` part so I think you are not the same environment as "vanilla installation" (Maybe you have a package named "librealsense2" in your catkin workspace)

FYI:
Regardless of setting an option `-r` to ignore errors on `rosdep install`, an error message still shows up.:

```
ubuntu@ubuntu:~/catkin_ws/src/realsense/realsense2_camera[development]$ rosdep install --from-paths . -i -y
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
realsense2_camera: Cannot locate rosdep definition for [librealsense2]
ubuntu@ubuntu:~/catkin_ws/src/realsense/realsense2_camera[development]$ rosdep install --from-paths . -i -y -r
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
realsense2_camera: Cannot locate rosdep definition for [librealsense2]
Continuing to install resolvable dependencies...
#All required rosdeps installed successfully
```

I'll now update this PR by adding a commit to remove `librealsense2` from dependencies to pass the test.
Could you consider merging the PR or any other comments are very welcome!

furushchev:
@doronhi Any feedback on this PR? It will be very nice if this PR is merged. :)

doronhi:
Adding librealsense2 to rosdistro achieved some progress. The latest PR is [here](https://github.com/ros/rosdistro/pull/21996). It still has some build issues at 32 bit environments that I hope to get into in the upcoming weeks. I try to hold the handling of the dependencies issues until then.

seiga-k:
Hi, all!
It seems already done this PR, but I'll have a comment.

How about to add a version attribute to dependency in package.xml? The package.xml can write a version specified dependency.
https://www.ros.org/reps/rep-0140.html#build-depend-multiple

The realsense2_camera package is deeply depended on librealsense2, so I comment as above.

furushchev:
Moved to #1032

PR #814, opened 2019-06-13
Reorganize and add missing dependencies

paulbovbel:
In particular missing:

- development deps of librealsense-dev, dbg
- librealsense-dkms for running on hardware
- rgbd_launch, message_generation

### DIFF
--- a/realsense2_camera/package.xml
+++ b/realsense2_camera/package.xml
@@ -12,20 +12,30 @@
 
   <author email="sergey.dorodnicov@intel.com">Sergey Dorodnicov</author>
   <author email="doron.hirshberg@intel.com">Doron Hirshberg</author>
+  
   <buildtool_depend>catkin</buildtool_depend>
-  <depend>image_transport</depend>
+  
   <depend>cv_bridge</depend>
-  <depend>nav_msgs</depend>
-  <depend>nodelet</depend>  
+  <depend>diagnostic_updater</depend>
+  <depend>ddynamic_reconfigure</depend>
   <depend>genmsg</depend>
+  <depend>image_transport</depend>
+  <depend>nav_msgs</depend>
+  <depend>nodelet</depend>
   <depend>roscpp</depend>
   <depend>sensor_msgs</depend>
   <depend>std_msgs</depend>
-  <depend>message_runtime</depend>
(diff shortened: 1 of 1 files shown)


realsense2_camera/package.xml
```diff
@@ -12,20 +12,30 @@
...
   <depend>sensor_msgs</depend>
   <depend>std_msgs</depend>
-  <depend>message_runtime</depend>
   <depend>tf</depend>
-  <depend>ddynamic_reconfigure</depend>
-  <depend>diagnostic_updater</depend>
-  <depend>librealsense2</depend>
+  
+  <build_depend>librealsense2-dbg</build_depend>
+  <build_depend>librealsense2-dev</build_depend>
```
paulbovbel, line 31: I'm pulling these packages from the amazon debian repos, and you need the `-dev` to build realsense-ros

doronhi: As for now, the packages for librealsense are not available as debian packages in ROS. Therefor it's on hold until they are accepted.

paulbovbel: We build realsense-ros in an internal CI system that requires the dependencies are straightened out in package.xml . Would you consider taking in these changes even if they're not used for upstream ROS builds (yet)?

This change allows us to use your release directly instead of maintaining a fork.

doronhi: Here is my concern: Take a look at the .travis.yaml file we currently use. Will it work if I added the dependencies while librealsense is still not available?

paulbovbel: Ping?

doronhi:
The process of releasing librealsense2 as a debian in the rosdistro is underway. It takes a long time, mainly because of availability and prioritizing.
This PR is not the final form of the solution we are interested in and I am inclined to merge temporary modifications.
I am very sorry for, and appreciate the effort of maintaining a different fork just for package.xml file but I prefer to hold changes to that file until a stable release of librealsense is available in rosdistro.

doronhi:
@paulbovbel ,
Now that librealsense2 is available in ros distro for kinetic (melodic will probably be added in the next ros package distribution), is this PR still needed?

PR #823, opened 2019-06-21
Make T265 less noisy in the info log

patbenatar:


joekeo:
thank you very much!

ihadzic:
Looks like the latest available package for Xenial still does not have d3867401225ab819fdae1efdb57909f294e99e3e, so when I turn on wheel odometry, my log gets spammed. Is there a plan to push a new apt package? The latest that I get is 2.2.6-2xenial-20190821-193308+0000. The repo I am getting it from, per my `/etc/apt/sources.list` is:

```
deb http://realsense-hw-public.s3.amazonaws.com/Debian/apt-repo xenial main
```

doronhi:
The entry in your sources.list file refers to the location of the librealsense2 library.
The latest available package of the ROS wrapper, realsense2_camera, is available, as for now, only as code that you download for this github.
The correction you refer to was merged and included maybe 2 releases ago.
Please check the [README.md](https://github.com/IntelRealSense/realsense-ros/blob/development/README.md) for the installation process.

PR #843, opened 2019-07-10
move ddynamic_reconfigure settings prior to start streaming.

doronhi:
Allowing configuration of parameters that needs setting before streaming starts, such as inter_cam_sync_mode


jasonbeach:
That is fantastic--thanks!

PR #845, opened 2019-07-10
Optionally disable TF pub, add service for querying serial number

5yler:


doronhi:
Thanks for the contribution.
There are a couple of set backs:
The PR should be divided to 2: one for the publish_tf flag and the other for the serial_no service.
* As for the serial_no service, could you describe your case where such a service is needed and add some documentation in the README.md file?
* As for the publish_tf flag, it was recently discussed at #9 where the need for dynamic_tf also came up.
@pavloblindnology is working on a solution to both issues at #169. I'd prefer a solution for dynamic and static tfs together as it is bound to simplify the interface.

doronhi:
With the merge of #902 , I think this PR can be closed. Agree?

PR #848, opened 2019-07-11
Get serial service

5yler:
Separate serial service feature of https://github.com/IntelRealSense/realsense-ros/pull/845

Example usage:

Using multiple robots with otherwise identical hardware configuration, one of the things included in log files to distinguish between the robots is the RealSense serial number. Previously I approached this by launching the RealSense node and manually copying the serial number from the terminal output, setting it as an environment variable on the robot computer, and querying that environment variable in the logging program. However, this is time-consuming and simply being able to query the serial number at runtime also makes swapping between cameras easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,17 @@ After running the above command with D435i attached, the following list of topic
 The "/camera" prefix is the default and can be changed. Check the rs_multiple_devices.launch file for an example.
 If using D435 or D415, the gyro and accel topics wont be available. Likewise, other topics will be available when using T265 (see below).
 
-### Launch parameters
+### Advertised Services
+The camera node advertises a service to query the device serial number:
+- /camera/get_serial
+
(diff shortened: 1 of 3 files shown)


5yler:
Adittionally, I am basing this off of existing implementations in other camera drivers, e.g. [`openni2_camera`](https://github.com/ros-drivers/openni2_camera/blob/93794e4f1d3dd0af4b070ad387d7e690a54a7519/openni2_camera/src/openni2_driver.cpp#L155)

schmidtp1:
I think this is a great idea but should the serial number maybe be published (as latched topic) to make sure it is also in any recording?

doronhi:
I think @schmidtp1 has a good point here. Publishing that information as a latched topic is both safer and cheaper since it's not calling on a service thread in the wrapper.
A good example for it would be the publishing of the intrinsics [see here](https://github.com/IntelRealSense/realsense-ros/blob/31fa75ec835650d319f2d5abd3f7d0ab66fd37f4/realsense2_camera/src/base_realsense_node.cpp#L1816)

doronhi:
I noticed that the serial number is published in the /diagnostics topic.
Does that mean that this PR can be closed?

PR #878, opened 2019-08-06
add asic temperature and projector temperature to diagnostics topic.

doronhi:
Add ASIC, projector and motion module temperatures to diagnostics topic.
mention /diagnostics in README.md


brolinA:
I have tried out the update and I works fine. Thank you @doronhi

Issue #885, opened 2019-08-08
Broken xacro support for multiple cameras


tykurtz:
There are a couple issues that I noticed with the URDF when trying to incorporate multiple camera models onto a robot.

Here are the tf trees after when running both the view_d435_model.launch and the ROS drivers.
```
# Example realsense drivers tf tree for D435
arg tf_prefix := ""
${tf_prefix}_link
|_ ${tf_prefix}_depth_frame
    |_ ${tf_prefix}_depth_optical_frame
|_ ${tf_prefix}_color_frame
    |_ ${tf_prefix}_color_optical_frame
|_ ${tf_prefix}_infra1_frame
    |_ ${tf_prefix}_infra1_optical_frame
|_ ${tf_prefix}_infra2_frame
    |_ ${tf_prefix}_infra2_optical_frame
```

```
# URDF tf tree
camera_link
|_ camera_depth_frame
    |_ camera_color_frame
        |_ camera_color_optical_frame
    |_ camera_depth_optical_frame
    |_ camera_left_ir_frame
        |_ camera_left_ir_optical_frame
    |_ camera_right_ir_frame
        |_ camera_right_ir_optical_frame
```
- [ ] Redundant frames are being published by the node if the urdf is included. This issue is already documented and can be addressed through #777 
- [ ] Inconsistent frame naming notation between the URDF and realsense driver. On the drivers, the monochrome cameras are called infra1 and infra2, and in the URDF it's left and right.
- [ ] The sensor frames derive from the depth frame for the URDF, and from the camera_link frame on the drivers.
- [ ] There is no ${prefix} argument as mentioned [here](https://github.com/IntelRealSense/realsense-ros/issues/494#issuecomment-439451045). This limits the usefulness of having a xacro at all. 

These changes should be straightforward, so I don't mind contributing. I just want to make clear the preferred changes before starting. To me, the left/right notation makes more sense, and is more in line with the `/stereo/left/` and `/stereo/right/` ROS namespace convention. However, making this change would break existing interfaces. I also prefer deriving from camera_link instead of the depth frame, so I would modify the URDF to match that notation.

saikishor:
FYI
https://github.com/IntelRealSense/realsense-ros/pull/957

RealSenseCustomerSupport:
----------------------------------------------

Thanks @saikishor's contribution.

PR #886, opened 2019-08-09
Modified 435 xacro to support multiple cameras

tykurtz:
See issue #885

Apologies for not breaking this up into separate PRs. I did several things here. I just modified the 435, but the same changes could apply to the other sensors if this seems acceptable.

- Modified the tf tree from the xacro to match the driver
- Added a "use_nominal_extrinsics" parameter. The thought behind this is if you are running in simulation, you will want the same frames that are published when running the real sensor. When you are running the real sensor, I don't see the use of publishing the nominal frames
- Added a name parameter and a simple launch file to test adding multiple camera models
- Added an example launch file that displays the d435 camera model with the correct orientation w.r.t. the depth cloud.

### DIFF
--- a/realsense2_camera/launch/rs_d435_camera_with_model.launch
+++ b/realsense2_camera/launch/rs_d435_camera_with_model.launch
@@ -1,0 +1,109 @@
+<launch>
+  <arg name="serial_no"           default=""/>
+  <arg name="json_file_path"      default=""/>
+  <arg name="camera"              default="camera"/>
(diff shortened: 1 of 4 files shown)


adrianramospeon:
I'm working on a project using multiple 435 cameras and stumbled upon the same problem. I started doing exactly what you've done, before finding your very helpful commit. Thanks for posting it. I fully support the merge.

Issue #872, opened 2019-07-31
Published camera_info messages not compatible with image_geometry::StereoCameraModel


tykurtz:
The frame_id of the camera_info message published on `/camera/infra2/camera_info` is camera_infra2_optical_frame. This is not compatible with the `image_geometry::StereoCameraModel` class as [StereoCameraModel::fromCameraInfo()](http://docs.ros.org/melodic/api/image_geometry/html/c++/stereo__camera__model_8cpp_source.html#l00027) makes the assumption that the left and right camera_info messages are in the same frame. So modifying the second camera to use the first camera frame would be more in line with ROS convention.

Without digging into the realsense-ros code, my hacky fix was to subscribe, modify the frame and projection matrix, and republish the modified camera_info as below, but an upstream fix would be preferred if posible.

```python
    def on_camera_info_received(self, camera_info):
        modified_camera_info = camera_info
        modified_camera_info.header.frame_id = 'camera_infra1_optical_frame'
        projection_matrix = list(modified_camera_info.P)
        projection_matrix[3] = projection_matrix[0] * .0499728  # fx * baseline
        modified_camera_info.P = projection_matrix

        self.right_camera_info_publisher.publish(modified_camera_info)
```

doronhi:
I was not aware of that convention. It means that there shouldn't be published a static transform for infra2 nor a topic named `/camera/extrinsics/depth_to_infra2`
Basically it means that infra2 should not be treated as a separate sensor.
It means an interface change.
I realize the logic in this but could you refer me to the official convention doc?

stwirth:
That spec can be found in the comment of the CameraInfo message definition:
http://docs.ros.org/kinetic/api/sensor_msgs/html/msg/CameraInfo.html

tykurtz:
[...]

PR #901, opened 2019-08-23
Add filling correct Tx, Ty values in projection matrix of right camera.

### LINKED ISSUE #427: Tx,Ty not filled in CameraInfo for infra2
According to the next line Tx, Ty is not filled for right image CameraInfo. As the result, baseline cannot be resolved in a standard ROS way. 
https://github.com/intel-ros/realsense/blob/5210e673ec0d266c62590b433c8b568954f4fdfd/realsense2_camera/src/base_realsense_node.cpp#L828

pavloblindnology:
Fixes https://github.com/intel-ros/realsense/issues/427

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1746,6 +1746,14 @@ void BaseRealSenseNode::updateStreamCalibData(const rs2::video_stream_profile& v
     _camera_info[stream_index].P.at(10) = 1;
     _camera_info[stream_index].P.at(11) = 0;
 
+    // Set Tx, Ty for right camera
+    if (stream_index == INFRA2 && _enable[INFRA1])
+    {
+        const auto& ex = getAProfile(INFRA2).get_extrinsics_to(getAProfile(INFRA1));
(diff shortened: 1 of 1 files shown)


doronhi:
It seems in related to #872.
However, if I understand correctly, we either consider infra2 as a separate monocular camera with a unique frame id, as is now, or consider it a part of the stereo couple as is suggested. In the first option, it's Tx, Ty should be zero, as is now. In the second option, as stated in #872, you are right.

I think that this PR should be merged along with removing infra2 frame id, and maybe some more modification I didn't think about yet.

pavloblindnology:
Well, never used stereo_imag_proc and didn't know about the convention. Instead, I always used stereo cameras as 2 separate cameras with different frames and right camera_info having Tx, Ty filled. So it's up to you to decide what to do with this.

doronhi:
@pavloblindnology , Finally I get to thank you for your PR.
I ended up merging #1242 instead as it incorporates your changes but I thank you for your contribution and support in understanding this concept.

doronhi:
I rebased and updated this PR, including your suggestion regarding the frame_id.
Thanks.

PR #906, opened 2019-08-29
fixed build no ddynamic_reconfigure

comwise:
fixed the build bug which is ''error: ‘class ddynamic_reconfigure::DDynamicReconfigure’ has no member named ‘registerVariable’

just to reduce depend the ros lib ddynamic_reconfigure


doronhi:
ddynamic_reconfigure is not included in realsense2_camera for quite some time now.
The reason is that there is no point in maintaining an existing, well maintained package.

PR #921, opened 2019-09-10
fix d435 Gazebo mesh visualization

YueErro:
Current:

Screenshot from 2019-09-10 10-54-41

In [#404 PR](https://github.com/IntelRealSense/realsense-ros/pull/404#issuecomment-404405038) is happening the same.

Fork:

Screenshot from 2019-09-10 10-57-49

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
   - source ~/.bashrc
 
     # download data:
-  - bag_filename="http://realsense-hw-public.s3.amazonaws.com/rs-tests/TestData/outdoors.bag";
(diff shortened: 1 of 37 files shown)


doronhi:
I am not very familiar with Gazebo. I loaded the existing d435.dae file and the new one, the one submitted in this PR. Indeed, the new one seems nicer.
The problem is that the files content is unclear to me.
The content of the files is very different and it is impossible for me to tell if the new model is correct.
Please describe the process by which you created the new d435.dae file. What is the source from which you converted or what modifications, if any, did you do to the original d435.dae file.

YueErro:
As far as I remember, I opened the `d435.dae` file with MeshLab and I exported it again as *.dae format. I have repeated the process and seems to work.

PR #953, opened 2019-10-15
Feature/parametrized xacro

veimox:
I would like to import two models into my xacro but when I do there is a collision with the links. This PR turns the hardcoded name into a parameter and sets the default name to "camera" (thus, it is backwards compatible). 

I also added unit testing which so far only checks to well-formed xacros and not for the actual content but is better than nothing :smile: 

Note: I also move the material definition to another file that can be imported from other xacros. This is needed to avoid the redefinition of the materials when generating the urdf.


aalexanderr:
seems like it is already done with backward compatibility in 7ace746d14520c0c50daa21468c16420ba72f3a1

Issue #987, opened 2019-11-12
Standard settings for D435i to get synchronized IMU and color image stream?


levenberg:
I am trying to use D435i to run the state-of-the-art visual-inertial odometry system. However, there is always something wrong to obtain good data input. One big issue is the synchronization of IMU and color image stream.  After setting `enable_sync=true`, we can get an imu topic: `\color\imu`. However, the rate of this imu topic is so confusing. 

if we set `unite_imu_method=copy`, the imu rate is constant, can be 263 hz, 450hz, 463hz or 650 hz.
if we set `unite_imu_method=linear_interpolation`, the imu rate is always changing. 

Based on these two settings, I cannot see that the IMU is synchronized to the color image stream, because it is always changing. And there seems no way to compare the timestamp of each topic to compute the lag because the timestamp seems to be added by my local computer, not by the realsense itself.

So is there any standard setting for D435i for using the IMU to get high-quality synchronized data to the color image steam. Can you give us a detailed tutorial on how to use it?

doronhi:
Just a clarification: 'enable_sync' causes the image streams to be synchronized but has no effect on the imu. The united imu topic is created by setting the `unite_imu_method`.

That said, there is an explanation on the expected frame rates of the imu in #729.
The Timestamps of both the images and imu are set based on the time stamp given on the device itself. They are not set by the time of arrival to the host computer. Therefor it should be possible to use them for synchronization of IMU and images. However, there is an known issue of different and small latency for each stream that is not fully taken into account yet and should be in upcoming versions.

levenberg:
@doronhi Thank you for your clarification. Because I always get the warnings of 
`HID timestamp not found, switching to Host timestamps.` and `Frame metadata isn't available! (frame_timestamp_domain = RS2_TIMESTAMP_DOMAIN_SYSTEM_TIME)`. But I can get a normal imu stream and color image stream, so I guess it is set as the host timestamps. 

Then for the synchronization, if both of the imu and color image stream are uniformed sampled based on the device itself, there should be like a constant time lag between the closed two topics. However, when I use [kalibr](https://github.com/ethz-asl/kalibr) to calibrate the extrinsics between the two sensors by recording the IMU topic (263hz) and color image topic (30hz),   the output is like `the two sensors are not synchronized`. The Warning is here:
`[ INFO] [1573666068.717761557]: RealSense Node Is Up!
[...]

Issue #1000, opened 2019-11-18
D435i IMU timestamp issue


levenberg:
When I want to run a VIO system on D435i, I always get a warning like `imu message in disorder!`. I checked the IMU topic subscribe node:
```
void imu_callback(const sensor_msgs::ImuConstPtr &imu_msg)
{
if (imu_msg->header.stamp.toSec() <= last_imu_t)
{
ROS_WARN("imu message in disorder!");
return;
}
last_imu_t = imu_msg->header.stamp.toSec();
}
```
It always gives me the warning of "imu message in disorder!", which means the current imu timestamp is equal or even smaller than the last one. It is so weird. Also, I tried to use `imu_msg->header.stamp.toSec() < last_imu_t`, it also gave me the warning. The Realsense sdk version is 2.30.0, the ros wapper is `development` branch. 

Note: by getting back to `librealsense` version: 2.17.1, the IMU timestamp problem is solved. There is no warning about `imu message in disorder!` For realsense ros version, it is 2.1.3.

doronhi:
I think it's the same issue you referred to in #987, right?
I managed to reproduce but didn't get around to do more. It may be related to the different delay times to the different sensors but I'm not sure. Will update when had progress.

PR #1010, opened 2019-12-04
fixed unite_imu_method copy to transmit at gyro's rate.

doronhi:
Fix issues #987, #1000 of timestamps going backwards when using "unite_imu_method:=copy"
Also sends messages at the fixed rate of the gyro. Accel information is the last one arrived at the time the message is sent.

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -171,38 +171,19 @@ namespace realsense2_camera
 
 
     private:
-        class CIMUHistory
+        class CimuData
         {
             public:
-                enum sensor_name {mGYRO, mACCEL};
-                class imuData
-                {
-                    public:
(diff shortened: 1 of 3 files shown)


studennis911988:
@doronhi 
First of all, thanks for updating those issue, however, I'm wondering that does it mean the new pull request can get synchronized image and IMU data no matter which "unite_imu_method" is used?
Thanks in advanced!

doronhi:
> @doronhi
> First of all, thanks for updating those issue, however, I'm wondering that does it mean the new pull request can get synchronized image and IMU data no matter which "unite_imu_method" is used?
> Thanks in advanced!

So far this PR deals with "unite_imu_method:=copy" only. I may change that to include the "linear" as well.
It does nothing regarding the synchronization between Image and IMU. It is a different subject, I think, and I will happily discuss that need if you could open an issue.
Thanks!

PR #1013, opened 2019-12-09
Revert "Fix wrong PointField Type for Colored Point Cloud Messages"

doronhi:
Reverts IntelRealSense/realsense-ros#993

The PR doesn't pass the test.


sahandy:
@doronhi So, does this mean the recommended solution in #993 was wrong? If yes, what would be the correct way to do it?

doronhi:
It caused a problem with the unit-test. I reinserted again and soon will also fix the test.

PR #1032, opened 2019-12-20
use rosdep install for installing dependencies on testing

furushchev:
In this PR, `rosdep install` is used for installing dependencies instead of using `apt-get` each package manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,9 @@ install:
   - sudo rosdep init
   - rosdep update
   - echo "source /opt/ros/kinetic/setup.bash" >> ~/.bashrc
-  - sudo apt-get install ros-kinetic-cv-bridge -y
-  - sudo apt-get install ros-kinetic-image-transport
-  - sudo apt-get install ros-kinetic-tf -y
-  - sudo apt-get install ros-kinetic-diagnostic-updater -y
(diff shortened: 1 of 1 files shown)


doronhi:
The CI run did not fail because of the librealsense2 installation. That is because lines 10-11 in .travis.yaml which are responsible for it are still there. I believe that in order to achieve the goal of this PR they should also be removed.

furushchev:
@doronhi Sorry, I overlooked your comment. I just removed the part that librealsense2 is manually installed.

furushchev:
@doronhi Now it passed the test!

doronhi:
Thanks. It seems OK now. 
I want to wait with merging it for a while, until I gain confidence in the process of releasing librealsense2 package into the ros distro.

doronhi:
The problem with this PR is as follows:
`rosdep install` installs packages from the ros-distro servers. There is no guarantee that at the time of publishing the new realsense2_camera version, the latest version of librealsense2 is available there. In fact, it is most likely not available there but only on Intels' servers (at deb http://realsense-hw-public.s3.amazonaws.com/Debian/apt-repo).
Therefor, although this PR is correct and that is the correct way to install the dependencies, this method might interfere with the automatic tests.

PR #1067, opened 2020-01-29
missing run dependency

AnasIbrahim:


doronhi:
rgbd_launch is only needed to run one specific launch file example.
This is not a requirement of the library.

PR #1095, opened 2020-03-01
Join query thread

doronhi:
fix bug that prevented execution of destructor.


Dominik-FKIE:
Why do you want to delete closeDevice()?
closeDevice() should surround sensor.stop and sensor.close with a try-catch for rs2::error to avoid termination of the process when there is no active stream.

doronhi:
> Why do you want to delete closeDevice()?
> closeDevice() should surround sensor.stop and sensor.close with a try-catch for rs2::error to avoid termination of the process when there is no active stream.

I'm not sure I understand your meaning.
closeDevice() was called from the destructor. There is no need in calling sensor.stop() and sensor.close() as these are called from sensor's own destructor.
As for a try-catch for rs2::error, it was not there to remove in the first place. If you have a suggestion, I am all for increasing stability.

PR #1097, opened 2020-03-02
Considerable CPU saving in BaseRealSenseNode::publishPointCloud()

facontidavide:
There were two missed opportunities for optimization:

- sensor_msgs::PointCloud2 should not be created at any frame, because that would force memory allocation, that is costly.

- using std::list should be forbidden ;)

On "My Machine", we are talking about about 40% less CPU using a D435i (I am looking forward to know if you see similar improvement ratio).

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -317,6 +317,9 @@ namespace realsense2_camera
         stream_index_pair _base_stream;
         const std::string _namespace;
 
+        sensor_msgs::PointCloud2 _msg_pointcloud;
+        std::vector< unsigned int > _valid_pc_indices;
+
     };//end class
 
 }
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1985,7 +1985,8 @@ void BaseRealSenseNode::publishPointCloud(rs2::points pc, const ros::Time& t, co
 
     const rs2::vertex* vertex = pc.get_vertices();
     const rs2::texture_coordinate* color_point = pc.get_texture_coordinates();
-    std::list<unsigned int> valid_indices;
+
+    _valid_pc_indices.clear();
     for (size_t point_idx=0; point_idx < pc.size(); point_idx++, vertex++, color_point++)
     {
         if (static_cast<float>(vertex->z) > 0)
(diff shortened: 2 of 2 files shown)


facontidavide:
To be fair, once the "decimation" filter is applied, the CPU saving is less dramatic, but still significant :smile:

doronhi:
On my machine it was also around 35% less CPU. Thanks!

jorgemiar:
Applied this and no longer seem to get the following error message after launching the ros node:

`[ERROR] [1583581271.939449058]: An error has occurred during frame callback: Error occured during execution of the processing block! See the log for more info`

I do still get these warnings though:

```
 07/03 11:56:51,937 WARNING [139780868249344] (messenger-libusb.cpp:42) control_transfer returned error, index: 768, error: No data available, number: 61
 07/03 11:56:52,000 WARNING [139780868249344] (messenger-libusb.cpp:42) control_transfer returned error, index: 768, error: No data available, number: 61
 07/03 11:56:52,142 WARNING [139780868249344] (messenger-libusb.cpp:42) control_transfer returned error, index: 768, error: Resource temporarily unavailable, number: 11
 07/03 11:56:52,205 WARNING [139780868249344] (messenger-libusb.cpp:42) control_transfer returned error, index: 768, error: Resource temporarily unavailable, number: 11
 07/03 11:56:52,267 WARNING [139780868249344] (messenger-libusb.cpp:42) control_transfer returned error, index: 768, error: Resource temporarily unavailable, number: 11

```
**Update:**

Not sure the errors above had anything to do with this PR but they disappeared after running `sudo purge ros-kinetic-librealsense2`. Still getting the following warnings when visualizing through rviz:

```
 07/03 16:49:26,051 WARNING [139661733263104] (types.cpp:49) Out of frame resources!
 07/03 16:49:26,052 ERROR [139661733263104] (synthetic-stream.cpp:48) Exception was thrown during user processing callback!
```

doronhi:
The "Out of frame resources!" message usually mean that your machine is overloaded and processing cannot keep up with incoming frames. You did not mention what machine you are running on or what frame rate you use but these could be the reason.
As this is discussion is no longer connected to this specific PR, if you'd like, I would be happy to continue the discussion in a separate issue.

YoshuaNava:
Hi all,
This was a really nice contribution. Thank you very much!

Nonetheless, I wanted to report that I recently found that it breaks the code from:

https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/src/base_realsense_node.cpp#L155

I have a fork of the repo which uses the function toggleSensors to turn off/on the sensors during operation, and after the change I'm getting a segmentation fault when the program reaches this point: https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/src/base_realsense_node.cpp#L2088

I also have another function to toggle RGB on and off (as I've found it decreases the metadata quality, i.e. exposure values are fixed and in a different scale). And the same issue happens. Both toggle functions are called by a ROS service.

@facontidavide Could you help in debugging this issue? As you know your code and your train of thought the best.

Thank you in advance.


Best,
Yoshua Nava

facontidavide:
Hi, I will try to figure this out.

Inspecting the code I can find no reason why this is happening. It is really weird! I will try to reproduce it

YoshuaNava:
> Hi, I will try to figure this out.
> 
> Inspecting the code I can find no reason why this is happening. It is really weird! I will try to reproduce it

Thanks a lot. If there is any further info or test needed from my side please let me know.

facontidavide:
@YoshuaNava , I think that I need to have a look to your fork, because in the official repo there is no way to easily call that method.

Otherwise it will take me a while to reproduce this :(

PR #1116, opened 2020-03-09
added launch file

thebrickhouse21:


doronhi:
The included launch files are meant as a reference for users to modify as they see fit. We try to avoid over populating this included list of launch files. I am happy that you managed to produce one that match your needs and thank you for sharing your experience and insights but I prefer not to add it to the repository.

Issue #999, opened 2019-11-18
D435 sensor model inconsistent with datasheet


tpet:
The URDF model seems to be inconsistent with the datasheet.

In forward direction the camera_link in the URDF model is placed at the level of the bottom screw mount (relative x = 0):
https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_description/urdf/_d435.urdf.xacro#L49

The bottom screw mount position is not given in the datasheet numerically (as far as I can tell) but it seems to be in the rear half at least:
https://www.intel.com/content/dam/support/us/en/documents/emerging-technologies/intel-realsense-technology/Intel-RealSense-D400-Series-Datasheet.pdf#page=106

According to
https://www.intel.com/content/dam/support/us/en/documents/emerging-technologies/intel-realsense-technology/Intel-RealSense-D400-Series-Datasheet.pdf#page=64
depth origin should be only (4.2mm = 0.0042m) from the front glass, i.e., in the very front.

I'm not sure what should be the design position of the bottom screw mount to tell the correction needed.

RealSenseCustomerSupport:
----------------------------------------------

Hi @tpet

We will take a look at our datasheet to make sure we have as much data as we possibly can there.  For reference we do have an updated datasheet from the version that you reference please use the following link to the latest datasheets for all our products:  https://dev.intelrealsense.com/docs/datasheets

What I've included below is a link to the CAD models files that is available for all our camera products.  The CAD model can provide you the ability to determine distances from the bottom tripod type screw location and use that for understanding mechanical with respect to your setup and configuration.
https://dev.intelrealsense.com/docs/cad-files

Also have a look at the latest PR updates that have gone into ROS:  https://github.com/IntelRealSense/realsense-ros

RealSenseCustomerSupport:
----------------------------------------------

Hi @tpet

Do you require further assistance?  If not then the thread can be closed.

tpet:
Will you fix the URDF model? Or you don't think it's inconsistent with the datasheet?

RealSenseCustomerSupport:
----------------------------------------------

Don't think it's inconsistent.

tpet:
Let me try to explain what I think is wrong.
The issue is with the camera frames in URDF. E.g., camera_depth_optical_frame is at the same x level as the camera_link and camera_bottom_screw_frame:
https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_description/urdf/_d435.urdf.xacro#L78
[...]

PR #1124, opened 2020-03-14
correct offset between camera_link and base_link

mcamurri:
As discussed in #999, the `camera_link` frame has x coordinates that are coincident with the one of the `camera_bottom_screw_frame`. 

The extrinsic parameters stored inside the device (correct me if I'm wrong) should have no knowledge of the distance between the optical frame and the bottom screw hole. The extrinsics are indeed retrieved by calibration with a checkerboard and can only provide the mutual location between the two camera optical frames and the IMU.

Therefore, the `camera_link` should be coincident with the `infra1_frame`, as confirmed by RViz
image

This PR adjusts the offset of the mesh and of the `camera_link` from `camera_bottom_screw_hole` by introducing new parametric quantities retrieved from the datasheet. In this way, the `camera_link` frame and their children are at the right offset from the front plate:
image

### DIFF
--- a/realsense2_description/urdf/_d435.urdf.xacro
+++ b/realsense2_description/urdf/_d435.urdf.xacro
@@ -28,6 +28,12 @@ aluminum peripherial evaluation case.
     <xacro:property name="d435_cam_height" value="0.025"/>
     <xacro:property name="d435_cam_depth" value="0.02505"/>
     <xacro:property name="d435_cam_mount_from_center_offset" value="0.0149"/>
+    <!-- glass cover is 0.1 mm inwards from front aluminium plate -->
+    <xacro:property name="d435_glass_to_front" value="0.1e-3"/>
+    <!-- see datasheet Revision 007, Fig. 4-4 page 65 -->
+    <xacro:property name="d435_zero_depth_to_glass" value="4.2e-3"/>
+    <!-- convenience precomputation to avoid clutter-->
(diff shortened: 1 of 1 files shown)


doronhi:
This correction is very appreciated.
However, I have this correction: camera_link is defined to be the same as depth_frame and infra1_frame. These are the values produced also by realsense2_camera node. camera_link is not aligned with the bottom screw.

mcamurri:
> This correction is very appreciated.
> However, I have this correction: camera_link is defined to be the same as depth_frame and infra1_frame. These are the values produced also by realsense2_camera node. camera_link is not aligned with the bottom screw.

I'm not exactly sure what do you mean by that. 

Are you saying that the `camera_link` and `bottom_screw_frame` are not coincident without this PR if you run the driver live? 

That is not possible because there is  a `0` on the `x` field of the joint connecting `camera_link` and `bottom_screw_frame` [here](https://github.com/IntelRealSense/realsense-ros/pull/1124/files#diff-87e047b9db6df86f3e7f1ecf5fffa916L47).

AFAIK, the driver does not produce the TF between `camera_link` and `bottom_screw_frame`, because the corresponding joint it is not in the `if` group controlled by `use_nominal_extrinsics`. If it did, there would be the same TF published twice (one by the driver, one by the `robot_state_publisher`).

doronhi:
@mcamurri ,
Sorry, I got it wrong. Looking at the model again I see I got confused between camera_link which coincides with depth_frame and base_link which indeed coincides with bottom_screw_frame.
Everything seems fine to me.

PR #1125, opened 2020-03-15
fix use_nominal_extrinsics arg/parameter usage

mcamurri:
The macros defined in the device files e.g.,  `_d435.urdf.xacro` have a parameter called `use_nominal_extrinsics`, but they are also dependent on an undefined argument [with the same name](https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_description/urdf/_d435.urdf.xacro#L75).

The `test_*_camera.urdf.xacro` files which are responsible for instantiating the macro do not define the argument either.

This works as long as the `xacro` command is called from a launchfile with the argument defined, but when called in isolation without any argument they fail.

There is no reason to depend on a argument in the macro definition files (e.g., `_d435.urdf.xacro`). Only the test files should depend on arguments, as calling `xacro` on the `_d435.urdf.xacro` has no effect because the macro is defined but not instantiated. 

This PR fixes that by removing the argument dependency on the `_d435.urdf.xacro` and other models, and introducing it in the `test_d435_camera.urdf.xacro` and other models.

In this way, calling `xacro` on either of the two with no arguments produces no error.

This PR was tested by launching all the `view_*_model.launch` files provided.

### DIFF
--- a/realsense2_description/urdf/_d415.urdf.xacro
+++ b/realsense2_description/urdf/_d415.urdf.xacro
@@ -73,7 +73,7 @@ aluminum peripherial evaluation case.
     </link>
 
     <!-- Use the nominal extrinsics between camera frames if the calibrated extrinsics aren't being published. e.g. running the device in simulation  -->
-    <xacro:if value="$(arg use_nominal_extrinsics)">
+    <xacro:if value="${use_nominal_extrinsics}">
       <!-- camera depth joints and links -->
       <joint name="${name}_depth_joint" type="fixed">
         <origin xyz="0 0 0" rpy="0 0 0"/>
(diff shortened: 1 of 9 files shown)


saikishor (approved):

tykurtz:
Thanks for fixing my mistake and the detailed explanation! To replicate the error on master you can run
```sh
xacro $(rospack find realsense2_description)/urdf/test_d435_camera.urdf.xacro
```
which will return `Undefined substitution argument use_nominal_extrinsics`. 

I can confirm with this change you can again use the `sensor_d435` macro without needing to define `xacro:arg name="use_nominal_extrinsics"`

LGTM!

PR #1172, opened 2020-04-17
Adding respawn option on the nodelet launch

alexfneves:
The realsense node dies occasionally. Respawn should be available for this kind of scenario, much like required already is.


weeshal:
Is this planning to be merged?

Issue #1165, opened 2020-04-13
Add USB plug/cable to model


JStech:
I recently broke a USB cable plugged into a D415 as a result of a self-collision. I've updated our model to include a plug with a short length of cable to prevent this, but it'd be a nice feature to have the plug in the URDF with a parameter controlling whether it is included.

MartyG-RealSense:
Thanks so much for your suggestion @JStech - I have labelled this message with an "Enhancement" (Feature Request) tag so that it can be tracked by @doronhi the RealSense ROS wrapper developer.

RealSenseCustomerSupport:
----------------------------------------------

Hi JStech,

Are you saying you physically "broke" the USB cable? ... by cutting it?

Thanks

JStech:
I bent it. The robot squeezed the camera past its forearm link, and there was no collision with the camera body, but the plug wasn't included in the model, so it didn't survive.
![IMG_20200411_122941](https://user-images.githubusercontent.com/519268/79609312-c290eb00-80b3-11ea-9a1c-37eef9ca0e03.jpg)

RealSenseSupport:
Oh, even the locking cable plug is bent? ... that looks hit very hard.

MartyG-RealSense:
Hi @JStech    Your pull request still has some conflicts that need to be resolved in order for the contribution to be merged into the SDK.   Are you still working on this pull please?   Thanks!

https://github.com/IntelRealSense/realsense-ros/pull/1173

doronhi:
> Hi @JStech Your pull request still has some conflicts that need to be resolved in order for the contribution to be merged into the SDK. Are you still working on this pull please? Thanks!
> 
> #1173

Thanks for the reminder @MartyG-RealSense . @JStech , it looks like a nice addition. It would be great if you could update your PR and we'll try to merge it by the next release.

JStech:
Thanks for the ping. Should be good to go.

doronhi:
Thanks for the PR. It's merged now.

MartyG-RealSense:
As the PR has been merged, I will close this case now.  Thanks everyone!

PR #1173, opened 2020-04-18
created plug, added to D415 and D435

JStech:
This addresses #1165 by adding an option to include a USB plug in the URDF models for the D415 and D435. The plug includes a collision body, to prevent a real plug self-colliding with a robot arm.


JStech:
It looks like the merge conflict that I just resolved arose because `development` now targets Noetic. I don't have a Noetic install handy, so I was unable to test my resolution, but it was pretty simple and should be good to go.

p0rys:
Hey,

for me it seems that the "add_plug" argument is missing in the launch file "view_d435i_model.lauch", the urdf-files _d435.urdf.xacro and _d435i.urdf.xacro and all test_d*_camera.urdf.xacro files. 
Also the plug is misaligned for the d435(i). 

I'm not able to provide a patch right now but maybe someone else could.

PR #1190, opened 2020-05-08
update dependancies for d400 line

acetousk:
In the process of downloading and using this, I came across an issue when running: `roslaunch realsense2_camera rs_d435_camera_with_model.launch` this is because the dependency *ros-melodic-realsense2-description* was not installed. This simply makes it easier for the users with the d400 line of sensors not have the same problem that I had.


doronhi:
Thank you for the remark and PR.
I hope you don't mind I took the liberty of altering your explanation a bit. I hope I kept the essence and spirit.

PR #1204, opened 2020-05-18
WIP: fixed usage of use_nominal_extrinsics parameter

procopiostein:
this parameter was being incorrectly parsed.
its value was not taken into account in each device's xacro.

this PR fixed the parameter usage and update the tests to use it properly.


saikishor:
:+1: 
Nice fix! Looks good to me!

saikishor (approved):
+1
Nice fix! Looks good to me!

saikishor (changes requested):
@procopiostein can you change `${arg use_nominal_extrinsics}` by `$(arg use_nominal_extrinsics)`

mcamurri:
Isn't this the same as #1125 ?

procopiostein:
@mcamurri not only it is the same but yours is much better documented and ready to merge. thanks, I'll close this.

PR #1256, opened 2020-06-23
[realsense2_camera] Service call to run self calibration

YoshuaNava:
Added a service call to run self calibration.

I also added a helper functions that
* Extract the current calibration tables.
* Attempt self-calibration a pre-defined number of times.
* Check that the calibration health score is good, and the result is usable.
* Write the results to the device.
using the self-calibration API.

Another PR closely related to this one: https://github.com/IntelRealSense/librealsense/pull/6600
(There is some code overlap in the self calibration helpers functions)


doronhi:
Hi @YoshuaNava ,
Thanks for the PR. I still didn't get the chance to test it but a few notes stood out.
1. The code doesn't compile. Check out the travis output. It seems like a minor issue of casting but still needs to be fixed.
2. Using cout to print the messages is not the way. You should use the logger.
3. You check the calibration results to 3 levels: OK, recommend-to-do-it-again, fail. The user may not have access to the console. The message and the results, should be set into the response of the service and not printed to the console.

I hope I'll get around to check the performance and carry on with the merging process soon.
Thanks.

YoshuaNava:
Hi @doronhi,

> The code doesn't compile. Check out the travis output. It seems like a minor issue of casting but still needs to be fixed.

Thanks. I'll fix the build issue.

> Using cout to print the messages is not the way. You should use the logger.

I guess you mean the ROS logging mechanism. I have an open question about this which I posted below.

> You check the calibration results to 3 levels: OK, recommend-to-do-it-again, fail. The user may not have access to the console. The message and the results, should be set into the response of the service and not printed to the console.

You're right. I will change this.

Two additional questions:

1) Should we do anything special for devices that are connected through USB 2.x?
2) I also submitted an [example to librealsense](https://github.com/IntelRealSense/librealsense/pull/6600), which users a similar tooling (the open device-calibrate-iterate-return to user way of handling calibration).
    - Is this something we could push forward? Where I could actually grab the calibration helpers from librealsense directly. In that case, the logging mechanism would be that of the librealsense lib instead of ROS. Also, the ROS 2 wrapper could benefit as well in the short term.


I'll fix a small issue with the performance which I discovered while testing today and I'll notify you to start the tests..

Thank you.

Best,
Yoshua

doronhi:
I had more thought on the issue.
It's best practice to do self-calibration with minimal resolution and maximal frame rate. It increases speed, accuracy and chance of success. This is especially important in the ROS scenario where no GUI is available. 
An example can be seen in [depth_auto_calibration_example.py](https://github.com/IntelRealSense/librealsense/blob/f7cdf6e8961e1709f6d864bdb33095c00a671ca7/wrappers/python/examples/depth_auto_calibration_example.py#L31)
In order to achieve that, in [realsense-viewer](https://github.com/IntelRealSense/librealsense/blob/f7cdf6e8961e1709f6d864bdb33095c00a671ca7/common/on-chip-calib.cpp#L361) there is a flow of stopping the viewer, starting the sensors with reduced resolution and restoring the viewer.

If you did that in realsense2_camera, subscribers will suddenly get images in different resolution and react unpredictably. It means, for instance, that publishers should stop publishing while calibration process is at work. More issues will probably arise with the callbacks, filters and more.

You could stop all the realsense2_camera nodes, restart them at optimal settings for calibration, do the calibration and then restore them to your own settings. But instead, I wonder why not use librealsense2's depth_auto_calibration_example.py? 
You can wrap it if running it remotely is a problem or modify it to choose a specified camera or improve it as you see fit. It will be easier to maintain and be more general in its use.

YoshuaNava:
> Hi,
> Please wait before fixing. New considerations were brought up to my attention and I will discuss it further tomorrow.

Will do. Thanks for notifying.

doronhi:
It seems that the option to auto-calibrate using librealsense2's python examples are easy enough and make adding this PR to the complexity of the wrapper unnecessary.

YoshuaNava:
Dear @doronhi,

Sorry for the long time without replying. I didn't get a notification for your message because it was edited over another reply you made on this thread.

Sounds great. I'll try out the python examples and get back to you.

Thank you.

Best,
Yoshua

PR #1297, opened 2020-07-21
Update rs_camera.launch RGB camera resolution to include L515.

krazycoder2k:
Currently 640x480 is not supported on L515. Update default RGB resolution to one that's supported across all RealSense cameras.


doronhi:
640x480 IS supported by the L515. type `rs-enumerate-devices` to confirm.

PR #1308, opened 2020-07-27
chaned 2.2.8 to be compatible with ros indigo and nvidia tk1

mikexyl:


doronhi:
I didn't check so I'll take your word for it that with this change alone the version fits ROS Indigo.
I am, however, inclined not to add modifications and start supporting EOL versions of ROS.
Thank you for your contribution but I will avoid merging this PR in the main stream.

Issue #1305, opened 2020-07-26
bug: inconsistent argument `add_plug` in urdf and launch files


hanzheteng:
The most recent merged pull request #1173 brings a new bug for launch files. 

https://github.com/IntelRealSense/realsense-ros/blob/6a9cc994e3a9f63ddff1938a86ff04ca990d2551/realsense2_description/urdf/_d435.urdf.xacro#L145

This line brings a new bug -- `add_plug` is not defined by default in this file, and not introduced as an argument in launch file:

https://github.com/IntelRealSense/realsense-ros/blob/6a9cc994e3a9f63ddff1938a86ff04ca990d2551/realsense2_camera/launch/rs_d435_camera_with_model.launch#L105

I revert to version 2.2.15 and it works just fine. 

Not sure if the same issue happens to file `_d415.urdf.xacro` and/or other launch files. Please double check. Thanks.

doronhi:
Thanks for noticing and investigating. We'll check.

doronhi:
Hi, with #1335 by @lucarinelli merged, can this issue be closed?

hanzheteng:
Sure. I have no problem with the previous version and am not able to test the latest version. If you have tested and feel good, go ahead and close it.

PR #1335, opened 2020-08-18
set a false default value for add_plug

lucarinelli:
This seems to solve #1305. 
I tested it by launching `roslaunch realsense2_description view_d435_model.launch` which uses `add_plug:=true` at line 2 and `roslaunch realsense2_description view_d435i_model.launch` which does not define `add_plug`.

### DIFF
--- a/realsense2_description/urdf/_d435.urdf.xacro
+++ b/realsense2_description/urdf/_d435.urdf.xacro
@@ -14,6 +14,7 @@ aluminum peripherial evaluation case.
   <xacro:include filename="$(find realsense2_description)/urdf/_usb_plug.urdf.xacro" />
 
   <xacro:macro name="sensor_d435" params="parent *origin name:=camera use_nominal_extrinsics:=false">
+    <xacro:arg name="add_plug" default="false" />
     <xacro:property name="M_PI" value="3.1415926535897931" />
 
     <!-- The following values are approximate, and the camera node


lucarinelli:
CI seems to be failing because of unrelated problems with librealsense version

christian-rauch:
I have the same issue (`Undefined substitution argument add_plug`) and this PR fixes that issue.

@lucarinelli You can try to force-push the repo state again to force a CI restart: https://stackoverflow.com/a/36538741/8144672.

lucarinelli:
[The problem](https://travis-ci.org/github/IntelRealSense/realsense-ros/jobs/718899644#L2941) seems to be a mismatch between the librealsense version currently requested by the development branch and the one installed by CI.

```
CMake Warning at realsense/realsense2_camera/CMakeLists.txt:43 (find_package):

  Could not find a configuration file for package "realsense2" that is

  compatible with requested version "2.37.0".

  The following configuration files were considered but not accepted:

    /usr/lib/x86_64-linux-gnu/cmake/realsense2/realsense2Config.cmake, version: 2.36.0

CMake Error at realsense/realsense2_camera/CMakeLists.txt:47 (message):

  

  

   Intel RealSense SDK 2.0 is missing, please install it from https://github.com/IntelRealSense/librealsense/releases
```

I saw that @doronhi bumped up librealsense version from 2.36.0 to pre-release 2.37.0 in the CMakeLists.txt in ed8145d

@christian-rauch That said I'm not sure restarting CI could result in a passing check.

By the way this is not that big of a change, I don't know if CI check is really needed.

realsense2_description/urdf/_d435.urdf.xacro
```diff
@@ -14,6 +14,7 @@ aluminum peripherial evaluation case.
   <xacro:include filename="$(find realsense2_description)/urdf/_usb_plug.urdf.xacro" />
 
   <xacro:macro name="sensor_d435" params="parent *origin name:=camera use_nominal_extrinsics:=false">
+    <xacro:arg name="add_plug" default="false" />
```
seebq, line 17: The other way to fix this is to add this arg like the others, see:

https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_description/urdf/_d415.urdf.xacro#L15

My suggestion is on the line above to do:

```
<xacro:macro name="sensor_d435" params="parent *origin  name:=camera  use_nominal_extrinsics:=false  add_plug:=false">
```

seebq (changes requested):
Thanks for doing this, the other files that I checked seem to have this arg, but it was missed on the _d435.urdf.xacro.

heads up @alecgraves!

realsense2_description/urdf/_d435.urdf.xacro
seebq: Yes, it looks like it was just left off in this merge:

https://github.com/IntelRealSense/realsense-ros/commit/6a9cc994e3a9f63ddff1938a86ff04ca990d2551

It was added to the 415 macro:

https://github.com/IntelRealSense/realsense-ros/commit/6a9cc994e3a9f63ddff1938a86ff04ca990d2551#diff-27a1d25d2cf34294ff181fcac43ae14cR15

But not the 435 macro:

https://github.com/IntelRealSense/realsense-ros/commit/6a9cc994e3a9f63ddff1938a86ff04ca990d2551#diff-87e047b9db6df86f3e7f1ecf5fffa916R16

PR #1352, opened 2020-08-26
Update CMakeLists.txt

farhangnaderi:
Version realsense2 2.36.0 is the real SDK but Cmake is looking for 2.37.0. Otherwise CMake wont work.

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -40,7 +40,7 @@ endif()
 if (WIN32)
 find_package(realsense2 CONFIG REQUIRED)
 else()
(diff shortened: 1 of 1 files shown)


doronhi:
realsense2_camera version 2.2.16 is intended to work with librealsense2 version 2.37.0. Install it and there is no need to modify the CMakeLists.txt file.

farhangnaderi:
@doronhi How did you install the wrapper?

doronhi:
> @doronhi How did you install the wrapper?

Sorry, I didn't quite understand the question. Can you elaborate? (preferably on a new issue as I would also like to close this PR if it is ok with you)

PR #1388, opened 2020-09-15
Add base_frame_id argument to rs_camera.launch

Nerlant:
This allows to override the base frame id if it does not correspond with $(arg tf_prefix)_link.


doronhi:
Thank you for the PR.
Although it is correct, one might ask what about the rest of the frame_ids listed in nodelet.launch.xml?
Anyone can use them in their own launch file but I don't think we need to copy all the available parameters to all the launch files.

PR #1396, opened 2020-09-17
[Bugfix] various fixes for canonical ROS2

AustinDeric:
# What this PR does
* use canonical ROS2 includes
* add_library should not contain header files
* remove extraneous depends from realsense2_node

## Developer Notes:
* @doronhi was the eloquent branch recently force pushed to? If so, please don't do that, it makes rebasing extremely painful. ty!
* This PR should probably be against a foxy branch, but none exists. (I am currently using this for foxy)
* There are other fixes necessary, but this is my first pass. Well see how this PR goes


realsense2_camera/CMakeLists.txt
```diff
@@ -106,32 +105,23 @@ if(NOT realsense2_FOUND)
...
     src/realsense_node_factory.cpp
     src/base_realsense_node.cpp
     src/t265_realsense_node.cpp
-    )
-
+)
 
 target_link_libraries(${PROJECT_NAME}
     ${realsense2_LIBRARY}
     )
```
AustinDeric, line 120: `target_link_libraries` shouldn't be necessary because it should be handled by `ament_target_dependencies`. However, after the rebase i couldn't get it to work, so i had to leave it.

doronhi:
Installing on a clean Ubuntu18 with eloquent, the file found after running rosdep install was:
image_transport/image_transport.h and not image_transport/image_transport.hpp.

So the fix in the following file was not accepted.
realsense2_camera/include/realsense2_camera/base_realsense_node.h

AustinDeric:
Thanks! I'll try again on that dep. a few other updates too.

doronhi:
Looking forward to it. Thanks.

Issue #1443, opened 2020-10-12
Missing <exec_depends> in package.xml


jarvisschultz:
Part of this issue mentioned in #678 and the issue was closed without resolution. However, I think the reasoning is not quite right. The `realsense2_camera` package provides a launch file, `rs_rgbd.launch`, that includes a launch file from the `rgbd_launch` package. According to the [catkin documentation](https://docs.ros.org/melodic/api/catkin/html/howto/format2/catkin_library_dependencies.html#exec-depend), that is exactly one of the reasons to include an `<exec_depend>`. Moreover, there are numerous other packages that provide nodes and launch files used by this package that are missing `<exec_depend>` tags. If this package were to utilize [the `roslaunch_add_file_check` macro provided by roslaunch](http://wiki.ros.org/roslaunch#Catkin), the unit tests would fail as a result (indicating that there really are missing `<exec_depend>` tags). See below for a run of the unit test:

```
jarvis@jschultz:~/realsense-ros/realsense2_camera/launch ⟫ /opt/ros/kinetic/share/roslaunch/scripts/roslaunch-check .
checking *.launch in directory .
checking ./rs_rgbd.launch
Missing package dependencies: realsense2_camera/package.xml: rgbd_launch
checking ./demo_t265.launch
Missing package dependencies: realsense2_camera/package.xml: rviz
checking ./opensource_tracking.launch
Missing package dependencies: realsense2_camera/package.xml: robot_localization, rtabmap_ros, imu_filter_madgwick
checking ./rs_camera.launch
checking ./rs_rtabmap.launch
Missing package dependencies: realsense2_camera/package.xml: rtabmap_ros
checking ./rs_aligned_depth.launch
checking ./rs_multiple_devices.launch
checking ./rs_from_file.launch
checking ./rs_d435_camera_with_model.launch
Missing package dependencies: realsense2_camera/package.xml: robot_state_publisher, rviz
checking ./rs_d400_and_t265.launch
checking ./demo_pointcloud.launch
Missing package dependencies: realsense2_camera/package.xml: rviz
checking ./rs_t265.launch
...writing test results to /home/jarvis/.ros/test_results/realsense2_camera/rosunit-roslaunch_check_sense2_camera_launch.xml
FAILURE:
[./rs_rgbd.launch]:
        Missing package dependencies: realsense2_camera/package.xml: rgbd_launch
[./demo_t265.launch]:
        Missing package dependencies: realsense2_camera/package.xml: rviz
[...]

PR #1444, opened 2020-10-12
Add missing <exec_depend> tags

jarvisschultz:
Addresses #1443


doronhi:
The realsense2_camera package is installed on all sorts of machines. Our goal is to keep the list of dependencies as minimal as possible. As a move in this direction we already split the original package into realsense2_camera and realsense2_description packages.
All those additional dependencies are required for certain demonstration launch files only but not for the node itself. The overweight of these packages, not all of them needed for most users will do more damage then good. Therefor I will not accept this PR.
We could think of splitting the package again and adding an "examples" package or maybe some other idea.

jarvisschultz:
That all makes sense. I think the idea of making an `examples` repository is a very good idea. Thanks for the response!

PR #1450, opened 2020-10-15
Add: ros noetic branch with cpp 14 CMake

Darkproduct:
We succeeded with building and running this on noetic with the `legacy` branch on [librealsense](https://github.com/IntelRealSense/librealsense). We suggest to merge this or create a new branch.

### DIFF
--- a/realsense_camera/CMakeLists.txt
+++ b/realsense_camera/CMakeLists.txt
@@ -8,7 +8,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 #set(CMAKE_VERBOSE_MAKEFILE on)
 
 # Set compile flags
-set(CMAKE_CXX_FLAGS "-fPIE -fPIC -std=c++11 -O2 -D_FORTIFY_SOURCE=2 -fstack-protector -Wformat -Wformat-security -Wall ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "-fPIE -fPIC -std=c++14 -O2 -D_FORTIFY_SOURCE=2 -fstack-protector -Wformat -Wformat-security -Wall ${CMAKE_CXX_FLAGS}")
 # Flags executables
 set(CMAKE_EXE_LINKER_FLAGS "-pie -z noexecstack -z relro -z now")
 # Flags shared libraries


doronhi:
I tested using -std=c++11 on Noetic (focal) and merged #1474 . Is there a reason to switch to -std=c++14?

Darkproduct:
We have a Intel RealSense R200, which is no longer supported by version 2.0 on [librealsense](https://github.com/IntelRealSense/librealsense). But with the `legacy` branch from [librealsense](https://github.com/IntelRealSense/librealsense) the R200 still works. 

For this setup, I created a new branch from `IntelRealSense:indigo-devel` called `noetic-devel`. The change to `-std=c++14` fixed all the compiler errors.

Therefore I created this merge request to merge to `indigo-devel` and not to the `development` branch, because we needed to support the older version on ros-noetic.

doronhi:
Thanks for clarifying the context for me. I think I rather have 1 unmaintained branch in the repository than a couple as I fear a branch called noetic-devel will confuse users. If the same version works on indigo and noetic, I'd rather rename indigo-devel branch to legacy-devel branch and use some phrase like this (didn't test it - could have syntax issues...):
```
if [[ $(lsb_release -sc) == "focal" ]]; then _cpp_compiler=c++14;
else _cpp_compiler=c++11;
fi
```
What do you think?

Darkproduct:
Sounds good to me. Then the `legacy` branch from [librealsense](https://github.com/IntelRealSense/librealsense) will be compatible to this new `legacy` branch here. It would be nice, if you can point this out somewhere. Additionally, the `legacy` branch wouldn't suggest to not be compatible to newer ros versions. We use ros melodic and noetic and haven't had issues so far.

realsense_camera/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 #set(CMAKE_VERBOSE_MAKEFILE on)
 
 # Set compile flags
-set(CMAKE_CXX_FLAGS "-fPIE -fPIC -std=c++11 -O2 -D_FORTIFY_SOURCE=2 -fstack-protector -Wformat -Wformat-security -Wall ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "-fPIE -fPIC -std=c++14 -O2 -D_FORTIFY_SOURCE=2 -fstack-protector -Wformat -Wformat-security -Wall ${CMAKE_CXX_FLAGS}")
 # Flags executables
```
doronhi, line 12: The compiler should be selected according to the OS version.

doronhi (changes requested):
Based on our discussion, the compiler should be selected with something of the form:
```
if [[ $(lsb_release -sc) == "focal" ]]; then _cpp_compiler=c++14;
else _cpp_compiler=c++11;
fi
```

Could you please fix and test? Then I can rename the branch.

Nir-Az:
We will close this PR as won't do as we no longer developing ROS1.

Thanks for your effort.

Issue #1417, opened 2020-09-24
Frequency diagnostics not published until getNumSubscribers > 0


mikepurvis:
Due to issues with hardware and firmware reliability, our robot has a power cycle retry loop on our realsense cameras. We use the built-in FrequencyDiagnostics from this driver to determine whether the camera is up, to avoid our monitoring process needing to actually receive all of the published image data.

Unfortunately, the way the driver is structured at present, the frequency diagnostic does not get updated or even published unless the RealSense image topics are being subscribed to. The diagnostic update happens in `publishFrame`, here:

https://github.com/IntelRealSense/realsense-ros/blob/e273f04d480f9331b1919f24f59e7712fd2bc688/realsense2_camera/src/base_realsense_node.cpp#L2233

And there are several places this is guarded behind a `getNumSubscribers > 0` condition, both one internal to `publishFrame`, but also a separate one in `publishAlignedDepthToOthers`, which does additional processing before calling through to `publishFrame`.

In my opinion, this is counter to the spirit of diagnostics, and could be especially confusing in cases such as a gripper cam where the logic which consumes the images is intentionally shut down when the gripper is not actively in use. Under the current model, that device would get flagged by diagnostic_aggregator as having a stale diagnostic during the time that there is no active subscriber.

Would we entertain a pull request to fix this so that the frequency diagnostics are always published regardless of subscriber status? Currently the calls are:

```
frame_callback -> publishAlignedDepthToOthers -> publishFrame -> publish
frame_callback -> publishFrame -> publish
```

IMO the non-depth case should be moved to a utility function similar to `publishAlignedDepthToOthers`, and that would make all of this logic easier to follow and understand. However, for the very immediate, my proposed approach would be to move the subscriber check from `publishFrame`, making it instead external to where it is called from `frame_callback`, and then outside that check (in both the depth and non-depth cases), calling the `FrequencyDiagnostics::update` method.

Does this approach seem reasonable? I also note that the current ROS 2 branch of this repo doesn't have any of the diagnostics stuff included at all, so I'm not sure how strong the commitment is to it.

doronhi:
Hi @mikepurvis ,
Thank you for this elaborate use case.
I agree with what you said and would love to have a PR according to your suggestion.

As for the ROS2, the topic is currently missing only because of a technical issue:
[...]

PR #1471, opened 2020-10-28
Asynchronous stream frequency diagnostics

kubja:
- This pull request modifies frequency diagnostics to run independently from the frame callbacks. In the current implementation, frequency is updated with frame callback when nodes were subscribed. The asynchronous setup is more suitable for sensor health monitoring and enables to detect that there is a connection issue even when no node is subscribed. The argument was also made in issue #1417 .
- The diagnostics messages for each stream are published alongside the temperature diagnostics update. The loop update rate was lowered from 10s to 1s.
- The function was tested with Intel Realsense D435i.

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -36,9 +36,13 @@ namespace realsense2_camera
         diagnostic_updater_.add(frequency_status_);
       }
 
-      void update()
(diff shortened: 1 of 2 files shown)


mikepurvis:
Looks awesome! Thanks for taking a stab at this; your change also looks a lot simpler than what I was picturing, so that's great.

PR #1488, opened 2020-11-04
[capability] Option to enable infrared RGB frame

130s:
I'm opening this PR in place of the author @abhijitmajumdar

## Problem(s) this PR aims to resolve

Transformation between depth and RGB frames relies on both transforms (camera extrinsics) and camera info (camera intrinsics) to be accurate. Some points were seen missing when either one of them isn't as accurate.

## Proposed approach
- Although improving accuracy of the in/extrinsic parameters might be the most direct approach, we ruled it out as it might be in Intel's proprietary domain.
- Avoid transformation b/w depth and RBG where the 1-to-1 mapping between the point cloud and the RGB image happens, which requires high accuracy.
   - Since depth (in most realsenses') is infrared through the frame of reference of the IR camera, obtaining RGB from the infrared frame should actually work.

Currently although Realsense SDK / librealsense supports this, the ROS driver does not have a way to enable this. This PR suggests that.

## Changes in this PR
- Adds a new ROS param `infrargb` flag to the realsense node, which enables the a realsense camera to configure the infrared camera (which is usually used for depth estimation) to also be used as the RGB camera.
  - Limitations:
    - This is only possible with realsense cameras which use infrared cameras that support RGB stream (for example the `realsense D415` does, but the `realsense D435` and `realsense L515` do not).
  - Advantages:
    - This is helpful because the frame of reference for the depth frame and the RGB frame is the same eliminating frame conversion errors
  - Disadvantages:
    - The depth exposure is the same as the RGB exposure so fixing color image and fixing point cloud through exposure may have contradictory results and require figuring out a happy medium.
    - The RGB image now will have the infrared pattern in it since the infrared cameras do not have a filter for IR for the obvious reason

## Test

In private commercial applications in industrial material handling in logistics, this change has been extensively used in the production system for 12+ months.

(Btw thank you intelrealsense.com for the [nice blog post about the Realsense integration at us, Plus One Robotics](https://www.intelrealsense.com/plus-one-robotics/).)

## Open questions
- Would usage documentation be required? If so which one is the appropriate files to add such an info?

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,10 @@ Setting *unite_imu_method* creates a new topic, *imu*, that replaces the default
 - **publish_tf**: boolean, publish or not TF at all. Defaults to True.
 - **tf_publish_rate**: double, positive values mean dynamic transform publication with specified rate, all other values mean static transform publication. Defaults to 0 
 - **publish_odom_tf**: If True (default) publish TF from odom_frame to pose_frame.
+- **infrargb**: When set to True (default: False), it configures the infrared camera to stream in RGB (color) mode, thus enabling the use of a RGB image in the same frame as the depth image, potentially avoiding frame transformation related errors. When this feature is required, you are additionally required to also enable `enable_infra:=true` for the infrared stream to be enabled.
+  - **NOTE** The configuration required for `enable_infra` is independent of `enable_depth`
+  - **NOTE** To enable the Infrared stream, you should enable `enable_infra:=true` NOT `enable_infra1:=true` nor `enable_infra2:=true`
+  - **NOTE** This feature is only supported by Realsense sensors with RGB streams available from the `infra` cameras, which can be checked by observing the output of `rs-enumerate-devices`
 
 
 ### RGBD Point Cloud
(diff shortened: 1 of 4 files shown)


doronhi:
Thanks for the PR.
Documentation is indeed needed and best to put into the README.md file in the "Launch parameters" section.
Be sure to mention in the documentation that for RGB format `enable_infra:=true` should be used and NOT `enable_infra1:=true`.
Also, this option should be put into the nodelet.launch.xml file and rs_camera.launch file.

130s:
I see. This week we'll be hectic so we'll likely come back to updating this PR.

130s:
@abhijitmajumdar and I completed addressing review comments. Please re-review.

abhijitmajumdar:
> @abhijitmajumdar and I completed addressing review comments. Please re-review.

@130s I approve the changes. I dont have write access to formally approve this though

doronhi:
> @abhijitmajumdar and I completed addressing review comments. Please re-review.

Thanks,
I liked the detailed description.

PR #1489, opened 2020-11-04
Avoid z16h format

doronhi:
Force depth format to be Z16 and avoid Z16H which is not decoded yet by the wrapper.
Add notification if connected using USB2.1 port.


130s:
I see the change from https://github.com/IntelRealSense/realsense-ros/pull/1488 is included in this PR. Just wondering, if you want me to include the commit you added to https://github.com/IntelRealSense/realsense-ros/pull/1488 for the convenience or whatever purpose, I can do that. Please let us know.

doronhi:
Thank you for noticing. I must have mixed all the branches together when testing them. Thanks for the kind offer but I prefer to "clean my own mess" and keep the PRs separated.

PR #1498, opened 2020-11-12
Fix ROS2 launch file to allow changing parameters from default config

QQting:
Before this PR, the config is not used. Both the node_namespace and node_name are incorrect set to apply the config d435i.yaml. Users may not be able to know how to change the width and height. 

Therefore, I modified the launch file to change the default color_width and color_height in this PR.
This PR also gives an example to launch multiple cameras by assigning different serial_no in **rs_multi_camera.launch.py**.


doronhi:
@QQting ,
Sorry for taking so long to respond.
I do appreciate the concept of allowing parameters in the launch command, the way your example do:
`ros2 launch realsense2_camera rs.launch.py open_rviz:=true`
I think though, it's best if rviz is not part of realsense2_camera's installation. It's place is with realsense2_description. 
I also prefer rs_multi_camera.launch.py to allow setting serial_no from command line since when this package is installed as a debian I fear launch file's are much less accessible.
All in all, I liked your ideas and used you code as an example but preferred a different implementation as in #1557

PR #1502, opened 2020-11-15
Fix Option to enable infrared RGB frame

doronhi:
replaced both "infrargb" and "enable_infrargb" with "infra_rgb"
refers to #1488


130s:
This broke API technically speaking. I understand this was merged only a few days after https://github.com/IntelRealSense/realsense-ros/pull/1488 that introduced the original change was merged, so in the global community it was likely not a big deal at all, but it caused backlash for my work as we had been referencing to https://github.com/IntelRealSense/realsense-ros/pull/1488 until recently. So I think it would've been great if we got to know about this change ;)

CC @abhijitmajumdar

doronhi:
> This broke API technically speaking. I understand this was merged only a few days after #1488 that introduced the original change was merged, so in the global community it was likely not a big deal at all, but it caused backlash for my work as we had been referencing to #1488 until recently. So I think it would've been great if we got to know about this change ;)
> 
> CC @abhijitmajumdar

Noted, sorry for the problems it caused.
I usually make notes regarding API changes in the release notes of a new release. Since both PRs came out together in version 2.2.20, I didn't feel it was needed there.

PR #1508, opened 2020-11-18
Add urdf file for l515

doronhi:
Use the available usb plug model, although it has screw connectors that are not available in the real L515. To be added when correct usb plug model is available.


MikeHallettUK:
Can you please also add the D455 equivalents (like  D435i and IMU .xacro files ?) into realsense-ros/realsense2_description/urdf/ ?

Thanks

doronhi:
> Can you please also add the D455 equivalents (like D435i and IMU .xacro files ?) into realsense-ros/realsense2_description/urdf/ ?
> 
> Thanks

Will try to create a similar PR for D455 shortly.

Issue #1523, opened 2020-11-25
How to get RGB image from realsense L515 depth camera


kashishdhal:
Hello,

I am using the L515 camera to get the point cloud with the launch file **rs_camera.launch**. However, I want to get an RGB image along with the depth image. But I am not able to see **/camera/color/image_raw** topic being published. I am only able to see the depth data coming over the ROS topics.

Please help me to get the RGB image.

Regards,
Kashish Dhal

yuzhou42:
Hi Kashish, I just figure this out. In order to get the l515 color image, you will need to change the image size in the launch file. Because l515 doesn't support 640x480. You can choose anyone from 960x540, 1280x720 and 1920x1080.

Yu

kashishdhal:
Hello @yuzhou42 ,

Thanks for the reply! One question, will that also change the resolution of the depth image?

yuzhou42:
Hi @kashishdhal, it won't change the resolution of depth. They are using different parameters.

kashishdhal:
Hello @yuzhou42 ,

Thanks, it worked!

karanpandyaa:
I had same issue but the rgb topics got published when i ran:
 **roslaunch realsense2_camera rs_aligned_depth.launch**

PR #1536, opened 2020-12-01
Update rs_camera.launch

kashishdhal:
Invalid frame dimensions for color image in launch file. Refer to #1523

### DIFF
--- a/realsense2_camera/launch/rs_camera.launch
+++ b/realsense2_camera/launch/rs_camera.launch
@@ -24,8 +24,8 @@
   <arg name="enable_infra2"       default="false"/>
   <arg name="infra_rgb"           default="false"/>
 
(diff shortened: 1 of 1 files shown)


doronhi:
Color image of size 640x480 should be available for L515. Duplication of #1297

kashishdhal:
Hello @doronhi ,

It will be great if you can help me with Color image of size 640x480. I am not able to launch it, please help.

Regards,
Kashish Dhal

doronhi:
L515 color image has a 640x480 resolution at 15, 30 and 60Hz. So far I can't reproduce the issue where this resolution is not available. What is the output of `rs-enumerate-devices`?
Also, please open an issue "Resolution of 640x480 is not available for L515, color image" and copy your question there. This discussion is getting long and may result is something that will benefit many users.

PR #1564, opened 2020-12-10
Fix "Undefined substiution argument add_plug" for _d415.urdf.xacro

Tuebel:
This was a syntax issue using the xacro parameter.

### DIFF
--- a/realsense2_description/urdf/_d415.urdf.xacro
+++ b/realsense2_description/urdf/_d415.urdf.xacro
@@ -136,7 +136,7 @@ aluminum peripherial evaluation case.
       <link name="${name}_color_optical_frame"/>
     </xacro:if>
 
-    <xacro:if value="$(arg add_plug)">
+    <xacro:if value="${add_plug}">
       <xacro:usb_plug parent="${name}_link" name="${name}_usb_plug">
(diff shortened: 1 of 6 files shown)


meyerj:
I just had the same issue. [_d435.urdf.xacro](https://github.com/IntelRealSense/realsense-ros/blob/64066c27816ad64bb0b4f284992300286b5496d9/realsense2_description/urdf/_d435.urdf.xacro#L146) and [_l515.urdf.xacro](https://github.com/IntelRealSense/realsense-ros/blob/64066c27816ad64bb0b4f284992300286b5496d9/realsense2_description/urdf/_l515.urdf.xacro#L146) are also affected.

The unit test fails with
```sh
$ xacro realsense2_description/tests/one_of_each.xacro
Undefined substitution argument add_plug
when instantiating macro: sensor_d415 (../urdf/_d415.urdf.xacro)
in file: one_of_each.xacro
```

I pushed my patch to https://github.com/IntelRealSense/realsense-ros/compare/development...meyerj:fix/realsense-description-undefined-substitution-add-plug, but then found this pull request when I wanted to open a new one.

Tuebel:
Right, I added these fixes to this pull request. Moreover, I noticed that the `add_plug` args were not present in the test_....urdf.xacro

meyerj:
Another solution was suggested in #1621, but I think this one is more flexible and consistent with other arguments like `use_external_extrinsics` (https://github.com/IntelRealSense/realsense-ros/pull/1621#issuecomment-774559627).

PR #1597, opened 2020-12-28
Add QoS in realsense parameters list

gurbain:
Hi,

In the current version of the ROS2 foxy realsense node, the Quality of Service (QoS) is set to ["sensor data"](https://github.com/IntelRealSense/realsense-ros/blob/foxy/realsense2_camera/src/base_realsense_node.cpp#L934).

I'm not sure this choice is motivated by technical considerations about the sensor streaming capabilities, but there are many more [QoS profiles available](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h).

Moreover, the "sensor data" profile defines a "best effort" reliability policy, which is generally incompatible with default configurations of subscribing nodes, such as [rviz](https://github.com/ros2/rviz), [image_view](https://github.com/ros-perception/image_pipeline/blob/ros2/image_view/src/image_view_node.cpp), [rqt_image_view](https://github.com/ros-visualization/rqt_image_view/tree/foxy-devel), [depth_image_proc](https://github.com/ros-perception/image_pipeline/tree/ros2/depth_image_proc) (for 3D pointcloud reconstruction), and so forth... This could hold adoption of the realsense sensors in ROS2 as they cannot be visualized without setting up complex QoS policies.

Either I am missing something, either I suppose it could be useful to let the user choose itself another QoS profile from the yaml parameter file, as suggested in the commits attached to this PR.

### DIFF
--- a/realsense2_camera/config/d435i.yaml
+++ b/realsense2_camera/config/d435i.yaml
@@ -8,13 +8,16 @@ camera1:
         color_fps: 30.0
         color_height: 480
         color_width: 640
+        color_qos: "SENSOR_DATA"
 
         depth_fps: 30.0
         depth_height: 720
         depth_width: 1280
+        depth_qos: "SENSOR_DATA"
 
         infra_height: 720
         infra_width: 1280
+        infra_qos: "SENSOR_DATA"
 
         gyro_fps: 0.0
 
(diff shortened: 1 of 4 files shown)


doronhi:
Thanks for the contribution!
It seems to me that "best effort" reliability is best for sending vast amount of data, which is needed in real time - like images. I expect that the user prefers to receive images on run time and risk losing a few images than having them all but with greater delay.
As the default configuration for many other nodes is different, I could be mistaken and therefor may also change the default QoS for images to something else. If you know of any convention regarding QoS for different types of topics I would appreciate it if you could point me to a source.

gurbain:
Hi @doronhi,

Yes, what you're saying makes sense. I'm not an expert in QoS at all, and I honestly cannot tell which of "sensor data" or "default" profiles are the most logical choice to set by default. On a theoretical level, the "sensor data" profile could be more suited to image streaming but on a practical one, it can be a pain to use in combination with other nodes and the "default" profile have a point here.

However, in any case, offering the possibility to the user to choose the QoS in the publisher looks the most important to me, since several vizualization and processing nodes are generally configured in "default" profile (i.e., with a "reliable" reliability policy), which is incompatible with "best effort" reliability policy in the "sensor data" profile.

PR #1610, opened 2021-01-06
Fix mass of d415

awesome-manuel:
Corrected mass is taken from https://www.intelrealsense.com/wp-content/uploads/2020/06/Intel-RealSense-D400-Series-Datasheet-June-2020.pdf, Table 3-43.


doronhi:
Thanks!

PR #1618, opened 2021-01-12
[feature] Reset driver node on the fly

130s:
# Background of PR
This PR includes changes made by @malban and is an alternative to https://github.com/IntelRealSense/realsense-ros/pull/1504, which consists of more changes. To make review easier this PR aims to provide a more granular change.

# Target problem
- For whatever the cause (due to software/hardware reason), there can be cases where 2D or 3D image topics aren't published (e.g. the one described in https://github.com/IntelRealSense/librealsense/issues/7151).
- During runtime, when the device somehow gets disconnected and re-connected back in in the USB device level, Realsense node won't function (haven't found a ticket about this issue).

For automated system that relies on the images published by the Realsense, once that issue happens the system stops functioning.

# Approach

Detail is dscribed in https://github.com/IntelRealSense/librealsense/issues/7151#issuecomment-686404580. In short,
- We at Plus One Robotics found resetting the realsense node fixes the issue in question 99% time.
   
Note as mentioned there, this PR offers NOT a fix for the mentioned issue but aims to offer a better workaround.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -150,18 +150,19 @@ BaseRealSenseNode::~BaseRealSenseNode()
     std::set<std::string> module_names;
     for (const std::pair<stream_index_pair, std::vector<rs2::stream_profile>>& profile : _enabled_profiles)
     {
-        std::string module_name = _sensors[profile.first].get_info(RS2_CAMERA_INFO_NAME);
-        std::pair< std::set<std::string>::iterator, bool> res = module_names.insert(module_name);
-        if (res.second)
+        try
         {
-            try
+            std::string module_name = _sensors[profile.first].get_info(RS2_CAMERA_INFO_NAME);
+            std::pair< std::set<std::string>::iterator, bool> res = module_names.insert(module_name);
+            if (res.second)
             {
                 _sensors[profile.first].stop();
                 _sensors[profile.first].close();
             }
-            catch(const rs2::wrong_api_call_sequence_error& e)
-            {
-            }
+        }
+        catch (const rs2::error& e)
+        {
+            ROS_ERROR_STREAM("Exception: " << e.what());
         }
     }
 }
--- a/realsense2_camera/src/realsense_node_factory.cpp
+++ b/realsense2_camera/src/realsense_node_factory.cpp
@@ -30,7 +30,8 @@ std::string api_version_to_string(int version)
 }
 
 RealSenseNodeFactory::RealSenseNodeFactory():
-	_is_alive(true)
+	_is_alive(true),
+	_initialized(false)
 {
 	rs2_error* e = nullptr;
 	std::string running_librealsense_version(api_version_to_string(rs2_get_api_version(&e)));
@@ -55,10 +56,25 @@ RealSenseNodeFactory::RealSenseNodeFactory():
 
 RealSenseNodeFactory::~RealSenseNodeFactory()
 {
+	_initialized = false;
 	_is_alive = false;
 	if (_query_thread.joinable())
 	{
 		_query_thread.join();
+	}
+
+	try
+	{
+		_realSenseNode.reset();
+		if (_device)
+		{
+			_device.hardware_reset();
+			_device = rs2::device();
+		}
+	}
+	catch (const rs2::error& e)
+	{
+		ROS_ERROR_STREAM("Exception: " << e.what());
 	}
 }
 
@@ -220,23 +236,12 @@ void RealSenseNodeFactory::getDevice(rs2::device_list list)
 
 void RealSenseNodeFactory::change_device_callback(rs2::event_information& info)
 {
-	if (info.was_removed(_device))
+	if (_initialized)
 	{
-		ROS_ERROR("The device has been disconnected!");
-		_realSenseNode.reset(nullptr);
-		_device = rs2::device();
-	}
-	if (!_device)
-	{
-		rs2::device_list new_devices = info.get_new_devices();
-		if (new_devices.size() > 0)
+		if (info.was_removed(_device))
 		{
-			ROS_INFO("Checking new devices...");
-			getDevice(new_devices);
-			if (_device)
-			{
-				StartDevice();
-			}
+			ROS_ERROR("The device has been disconnected!");
+			reset();
 		}
 	}
 }
@@ -254,6 +259,13 @@ bool RealSenseNodeFactory::toggle_sensor_callback(std_srvs::SetBool::Request &re
 
 void RealSenseNodeFactory::onInit()
 {
+	auto nh = getNodeHandle();
+	_init_timer = nh.createWallTimer(ros::WallDuration(0.01), &RealSenseNodeFactory::initialize, this, true);
+}
+
+void RealSenseNodeFactory::initialize(const ros::WallTimerEvent &ignored)
+{
+	_device = rs2::device();
 	try
 	{
 #ifdef BPDEBUG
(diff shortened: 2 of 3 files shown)


doronhi:
Thanks for the simplified PR.

Tested on Ubuntu18.04, the call:
`rosservice call /camera/realsense2_camera/reset`

causes realsense2_camera node to die instantly.

Is this not the way to activate the "reset" service?

malban:
> Tested on Ubuntu18.04, the call:
> `rosservice call /camera/realsense2_camera/reset`
> 
> causes realsense2_camera node to die instantly.
> 
> Is this not the way to activate the "reset" service?

Yeah, that is how to call the reset service, but no, it shouldn't kill the node, it should just cause the driver to stop and get re-initialized.

I'll see if I can reproduce.

malban:
I'm not able to reproduce the crash on reset.  I'm testing with a d415 with LibRealSense v2.41.0 on 16.04.   I'll see about 18.04.

Are there any error messages?

130s:
@doronhi Thanks for testing. We'll continue verifying on Ubuntu 18.04 with D415 (and potentially with L515 too).

By any chance did you use other hardware than those listed above? If so we may not have a good insight as we don't own nor use other hardware.

doronhi:
I used D435i, though I doubt that's what makes the difference...
On Ubuntu 18.04 I got no error messages.

Tested with the same D435i on Ubuntu16.04 and the reset process succeeded.
However, half the times I also got the error message: "Tried to advertise a service that is already advertised in this node [/camera/enable]" I wonder if that behaves differently in Melodic.

doronhi:
Tested again with Melodic. The issue I had was with librealsense2 version 2.41.0. It is a known bug described in [#8154](https://github.com/IntelRealSense/librealsense/issues/8154) and fixed in [#8234](https://github.com/IntelRealSense/librealsense/pull/8234).
With the fix or with librealsense2 version 2.40.0 the reset service works ok in Ubuntu18.04 as well.

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -220,23 +236,12 @@ void RealSenseNodeFactory::getDevice(rs2::device_list list)
...
-		if (new_devices.size() > 0)
+		if (info.was_removed(_device))
 		{
-			ROS_INFO("Checking new devices...");
-			getDevice(new_devices);
-			if (_device)
-			{
-				StartDevice();
-			}
+			ROS_ERROR("The device has been disconnected!");
```
doronhi: _device = rs2::device(); // This will prevent failing attempt to hardware_reset() the non-existing device inside the following reset() function.

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -220,23 +236,12 @@ void RealSenseNodeFactory::getDevice(rs2::device_list list)
 
 void RealSenseNodeFactory::change_device_callback(rs2::event_information& info)
 {
-	if (info.was_removed(_device))
+	if (_initialized)
```
doronhi: Why not avoid the extra _initialized parameter and use if(_device) instead? See comment for line 397

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -326,44 +343,90 @@ void RealSenseNodeFactory::onInit()
...
-	_realSenseNode->publishTopics();
+	catch (const rs2::error& e)
+	{
+		ROS_ERROR_STREAM("Exception: " << e.what());
+	}
+}
+
+bool RealSenseNodeFactory::reset()
+{
+	if (!_initialized)
```
doronhi: I could not find the case where the variable "_initialized" is needed. Can you describe such a case?
The test here is redundant, I think:
	// if (!_initialized)
	// {
	// 	return false;
	// }

	// _initialized = false;
The rest of

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -150,18 +150,19 @@ BaseRealSenseNode::~BaseRealSenseNode()
     std::set<std::string> module_names;
     for (const std::pair<stream_index_pair, std::vector<rs2::stream_profile>>& profile : _enabled_profiles)
     {
-        std::string module_name = _sensors[profile.first].get_info(RS2_CAMERA_INFO_NAME);
-        std::pair< std::set<std::string>::iterator, bool> res = module_names.insert(module_name);
-        if (res.second)
+        try
```
doronhi, line 153: Is this change necessary? Did you encounter an exception in the area not previously inside the try-catch section?

doronhi (changes requested):
Please test if the _initialized parameter can be avoided and the code a little simpler.
Also, please test if the "Tried to advertise a service that is already advertised in this node [/camera/enable]" error message can be avoided.

malban:
> Please test if the _initialized parameter can be avoided and the code a little simpler.
Also, please test if the "Tried to advertise a service that is already advertised in this node [/camera/enable]" error message can be avoided.

Will do.  Thanks for reviewing.

realsense2_camera/src/realsense_node_factory.cpp
malban: I think I was intending this to prevent multiple overlapping calls to reset if it were to get called twice in quick succession.  But that would only be an issue if the node handler was multi-threaded, which it's not.  So I think you are right that using `if(_device)` is cleaner.

realsense2_camera/src/base_realsense_node.cpp
malban: At one point yes, we were seeing some unhandled realsense exceptions in the destructor.  This was originally developed against an older version of librealsense though, so it's possible those have been fixed in the latest version.  So to some degree, this may be preventative now.

malban:
Also, I tested 18.04 and was able to reproduce the failure you had, but thankfully it looks like that is already a known issue that has been fixed.

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -60,6 +60,20 @@ RealSenseNodeFactory::~RealSenseNodeFactory()
...
+		if (_device)
+		{
+			_device.hardware_reset();
+			_device = rs2::device();
+		}
+	}
+	catch (const rs2::error& e)
+	{
+		ROS_ERROR_STREAM("Exception: " << e.what());
+	}
```
doronhi, line 72: I think this hardware reset at the destructor is unnecessary. There is the initial_reset option if needed.

doronhi (changes requested):
Sorry I forgot to mention this last time. And also, sincerely apologizes for the long time it takes me to review.

realsense2_camera/src/realsense_node_factory.cpp
malban: Ok, yeah, that makes sense.

doronhi: Hi @malban, Do you think it will be possible for you to remove that part so it could be merged?

malban: Yeah, I'll look take care of it today.  Thanks for the reminder.

doronhi:
Thank you so much!
I removed the following section from the `~RealSenseNodeFactory() ` function:
```
	try
	{
		_realSenseNode.reset();
		_device = rs2::device();
	}
	catch (const rs2::error& e)
	{
		ROS_ERROR_STREAM("Exception: " << e.what());
	}
``` 
Please submit a new PR if you think I missed something and this section was indeed intentional.

PR #1621, opened 2021-01-14
fix add_plug in d415 test

fmessmer:
fixes the urdf tests for `d415` by using `add_plug` the same way as the other sensors

### DIFF
--- a/realsense2_description/urdf/_d415.urdf.xacro
+++ b/realsense2_description/urdf/_d415.urdf.xacro
@@ -12,7 +12,8 @@ aluminum peripherial evaluation case.
   <xacro:include filename="$(find realsense2_description)/urdf/_materials.urdf.xacro" />
   <xacro:include filename="$(find realsense2_description)/urdf/_usb_plug.urdf.xacro" />
 
-  <xacro:macro name="sensor_d415" params="parent *origin  name:=camera  use_nominal_extrinsics:=false  add_plug:=false">
+  <xacro:macro name="sensor_d415" params="parent *origin  name:=camera  use_nominal_extrinsics:=false">
+    <xacro:arg name="add_plug" default="false" />
     <xacro:property name="M_PI" value="3.1415926535897931" />
 
     <!-- The following values are approximate, and the camera node


meyerj:
Another solution was already suggested by @Tuebel in https://github.com/IntelRealSense/realsense-ros/pull/1564.

Either `add_plug` is an [xacro argument](http://wiki.ros.org/xacro#Rospack_commands) (declared with an `<arg>` tag, expanded  with `$(arg name)`) that can be set at the `xacro` command line, or it is a [macro parameter](http://wiki.ros.org/xacro#Macros) (declared with `<xacro:macro ... params="name">` and expanded with `${name}`). Those are completely unrelated concepts, unless one is set using an expansion of the other explicitly. Both can have default values.

At the moment it is inconsistent in [_d415.urdf.xacro](https://github.com/IntelRealSense/realsense-ros/blob/c11f8ff2db6a334bedd555e41f9d556a85b9b2bf/realsense2_description/urdf/_d415.urdf.xacro) because `add_plug` is defined as a macro parameter:
https://github.com/IntelRealSense/realsense-ros/blob/c11f8ff2db6a334bedd555e41f9d556a85b9b2bf/realsense2_description/urdf/_d415.urdf.xacro#L15

but then used as an xacro argument:
https://github.com/IntelRealSense/realsense-ros/blob/c11f8ff2db6a334bedd555e41f9d556a85b9b2bf/realsense2_description/urdf/_d415.urdf.xacro#L139

For the macro the syntax would be `${add_plug}`, like for the other parameters `${name}` or `${use_nominal_extrinsics}`.

In [_d435.urdf.xacro](https://github.com/IntelRealSense/realsense-ros/blob/c11f8ff2db6a334bedd555e41f9d556a85b9b2bf/realsense2_description/urdf/_d435.urdf.xacro) and [_l515.urdf.xacro](https://github.com/IntelRealSense/realsense-ros/blob/a8da5601f9387f188255cd79540a1aac858263ec/realsense2_description/urdf/_l515.urdf.xacro) `add_plug` is declared as and used as an xacro argument instead:
https://github.com/IntelRealSense/realsense-ros/blob/c11f8ff2db6a334bedd555e41f9d556a85b9b2bf/realsense2_description/urdf/_d435.urdf.xacro#L17

So there are two options to fix this:

- With this PR (https://github.com/IntelRealSense/realsense-ros/pull/1621) `add_plug` becomes an xacro argument consistently for all camera models, but it is not possible to pass the option as a parameter if the macro is included in another urdf file. So that is inconsistent with other macro parameters like `use_nominal_extrinsics`.
- On the other hand, with https://github.com/IntelRealSense/realsense-ros/pull/1564, `add_plug` is evaluated as a macro argument with `${add_plug}` consistently. Only the non-macro `test_*_camera.urdf.xacro` files (a stand-alone camera, like [test_d415_camera.urdf.xacro](https://github.com/IntelRealSense/realsense-ros/blob/3b3136e6e56b8bd21c438d5b0cd16f1c1d5baf86/realsense2_description/urdf/test_d415_camera.urdf.xacro#L4)) then declare the xacro argument that can be passed at the command-line or in a launch file, and the argument value then gets assigned to the respective macro parameter.

So I suggest to close this PR in favor of https://github.com/IntelRealSense/realsense-ros/pull/1621, but both solutions can potentially break an existing setup for somebody, depending on which of the urdf files is included.

doronhi:
Thanks. I eventually chose to go with #1621

PR #1634, opened 2021-01-20
Ordered PointCloud - foxy

doronhi:
add flag: ordered_pc - default to false. When set to true, pointcloud is published in an ordered format.

### DIFF
--- a/realsense2_camera/include/realsense2_camera/base_realsense_node.h
+++ b/realsense2_camera/include/realsense2_camera/base_realsense_node.h
@@ -251,6 +251,7 @@ namespace realsense2_camera
         float _depth_scale_meters;
         float _clipping_distance;
         bool _allow_no_texture_points;
+        bool _ordered_pc;
 
         double _linear_accel_cov;
         double _angular_velocity_cov;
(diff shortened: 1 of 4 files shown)


joelbudu:
Hello @doronhi, please is it possible for this feature to be merged into the eloquent branch as well? Would be very grateful

doronhi:
Add #1644.
Notice that the eloquent distribution is EOL. I added this feature to the eloquent branch but no debian can be released. I suggest moving on to foxy.

joelbudu:
Noted with thanks @doronhi . This may not be directly related to this repository but I need help urgently in getting support for the L515 camera in mentioned in this issue I created [https://github.com/intel/ros2_intel_realsense/issues/161](url) in the [https://github.com/intel/ros2_intel_realsense](url) repo. That is why I tried using the eloquent branch in this repository but it seems slight differences is causing crashes in my application. Could you be of help please?

doronhi:
The wrapper here supports the L515 device.
To my best understanding, the repository at https://github.com/intel/ros2_intel_realsense/ is just about to become unmaintained or is already so. Therefore, it's a good thing you switched to the wrapper in this repository.
I can't understand from your description what are the issues you are facing so I could help. I think it's best if you opened an issue describing the problems where they could be addressed.

joelbudu:
So I am actually using the [https://github.com/intel/ros2_grasp_library](url) which is part of the  [https://github.com/intel/robot_devkit](url) package. When I use the eloquent version from this repository instead of the dashing version, I get a Segmentation fault error when using the /plan_grasps service. I suspect there are differences in the structure of PointCloud2 messages being sent. I'm not sure there are active maintainers for some of the repositories. Thanks

doronhi:
@joelbudu, This could be related to #1664

joelbudu:
@doronhi Yes, that's the same issue I'm facing. I actually did some debugging and ended up at the same point where pcl::fromPCLPointCloud2() throws a segmentation fault

PR #1637, opened 2021-01-21
Add include guard to materials

LuAPi:
Prevents material already defined errors when adding multiple types of camera to a robot.

I was making a URDF for the d455 based on the d435 as I have a robot with both on and I ran into a material is not unique error. This fixes that.

I think it would actually be better if the materials were macros like
```
<xacro:macro name="aluminum_material">
  <material name="aluminum">
    <color rgba="0.5 0.5 0.5 1"/>
  </material>
<xacro:macro name="plastic_material">
    <material name="plastic">
      <color rgba="0.1 0.1 0.1 1"/>
    </material>
</xacro:macro>
```
Then changing the places they get used from `<material name="plastic"/>` to `<xacro:plastic_material />`
That way you don't end up putting very generic names like "aluminum" and "plastic" in the global namespace.

### DIFF
--- a/realsense2_description/urdf/_materials.urdf.xacro
+++ b/realsense2_description/urdf/_materials.urdf.xacro
@@ -8,11 +8,15 @@ Collection of materials to be used in other macros.
 This avoids the redefinition of materials in case multple cameras are imported.
 -->
 
-<robot>
-  <material name="aluminum">
-    <color rgba="0.5 0.5 0.5 1"/>
-  </material>
-  <material name="plastic">
-    <color rgba="0.1 0.1 0.1 1"/>
-  </material>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:property name="realsense_materials_defined" default="false" />
+  <xacro:unless value="${realsense_materials_defined}">
+    <xacro:property name="realsense_materials_defined" value="true" />
+    <material name="aluminum">
+      <color rgba="0.5 0.5 0.5 1"/>
+    </material>
+    <material name="plastic">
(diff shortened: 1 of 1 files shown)


doronhi:
I think your suggestion regarding the materials as macros is very reasonable.
Would you care to create a PR so it can be tested?
Also, a URDF for d455 is definitely something that would benefit many users. If already created, would you care to share and create a PR for that as well?

SamerKhshiboun (approved):

SamerKhshiboun:
@Nir-Az you I think can merge this PR for ros1-legacy.
I have duplicated this PR for ros2-development and assigned you as reviewer
#2689

civerachb-cpr:
Can you please bloom the ROS1 version of this package with that fix? The latest compiled .deb version still appears to be 2.3.2, which has the bug that this PR fixes.

Nir-Az:
> Can you please bloom the ROS1 version of this package with that fix? The latest compiled .deb version still appears to be 2.3.2, which has the bug that this PR fixes.

@civerachb-cpr sorry, we are no longer maintaining the ROS1 version, focus is on ROS2 only.

jeff-o:
Ouf, just ran into this bug on a robot I'm working on. This fix would be very welcome in the ROS1 .deb!

aditya2592:
I ran into this issue and found this PR. The PR suggests the following change but this change is not made in the PR itself. Is this expected? I think without this change, we still get the message `Error:   material 'aluminum' is not unique.`
```
Then changing the places they get used from <material name="plastic"/> to <xacro:plastic_material />
```

PR #1659, opened 2021-02-01
Emitter on off updates

BarzelS:


doronhi:
Thanks for this PR.
It needs some cleaning and also it contains changes I cannot accept, such as removing support for T265.
The logic of deciding which frame has IR projection is based on the sequential number of the frame. This seems a bit shakey.
I don't think there is a need to duplicate the variables _enable_emitter, _emitter_on_off, _enable_auto_exposure, manual_exposure. There is just a need to add a variable for deciding how to handle the situation (i.e. send in half the frame rate or not)
I will have to find the time to consider other implications, including the reliability of the feature in the device itself.
For all those reasons I think I may use this PR as a reference when the time comes.

PR #1668, opened 2021-02-04
Fix ordered PointCloud2 structure

joelbudu:
PR #1627 modifies the point-cloud message width & height but does not update the row_step to match. pcl::fromPCLPointCloud2() uses row_step to advance through the image, and hence segfaults when fed a message generated with ordered_pc message.

This fixes the structure of the pointCloud2 msg created


doronhi:
Duplication of #1676

Issue #1664, opened 2021-02-03
ordered_pc generates bogus messages


mferenduros:
PR #1627 [modifies the point-cloud message width & height](https://github.com/IntelRealSense/realsense-ros/commit/c5403a28af7a82a30e970f78d4e2c1a3649686b0#diff-a54ad2538ccd7b6cce4c6c6d8f57af062bded3d802b88e7b938faefb6f02f4caR2252) but does not update the row_step to match. pcl::fromPCLPointCloud2() uses row_step to advance through the image, and hence segfaults when fed a message generated with ordered_pc message.

MartyG-RealSense:
Hi @mferenduros  Thanks very much for your report.  I will tag @doronhi the RealSense ROS wrapper developer into this discussion to highlight it to them.

sean-roelofs-ai:
I have been running into segfaults when working with the d435 point cloud in the last few days. This is a real issue. Also I'm not sure this issue is just limited to orderd_pc, because I am also getting problems converting the pointcloud to pcl when it is unordered.

sean-roelofs-ai:
@mferenduros did you find a quick patch for this?

joelbudu:
I am also facing this same issue with segmentation faults pcl::fromPCLPointCloud2() with the ordered_pc

doronhi:
As rviz2 experiences no issues, can you detail a method for reproduction? example node maybe?

joelbudu:
@doronhi 
This is an example callback subscribing to a PointCloud2 topic. The pcl::fromROSMsg(*msg, *cloud) will cause a Segmentation fault.

```
void GraspDetectorGPD::cloud_callback(const sensor_msgs::msg::PointCloud2::SharedPtr msg)
{
      PointCloudPointNormal::Ptr cloud(new PointCloudPointNormal);
      pcl::fromROSMsg(*msg, *cloud);
}
```

Extracted this from the a - [larger codebase](https://github.com/intel/ros2_grasp_library/blob/master/grasp_ros2/src/grasp_detector_gpd.cpp#L178) in this  the [ros2_grasp_library](https://github.com/intel/ros2_grasp_library)
cc: @sharronliu

mferenduros:
> As rviz2 experiences no issues, can you detail a method for reproduction? example node maybe?

[...]

PR #1676, opened 2021-02-07
Fix ordered pointcloud2 format

doronhi:
related to #1664
based on @joelbudu's  #1668, set _msg_pointcloud.width before resize.
I prefer this implementation as it's shorter and contains fewer if-else cases.
If you approve, I'll try to add a unit-test based on the [example](https://github.com/mferenduros/pointcloud2_crash) from @mferenduros before merging.


joelbudu:
@doronhi, I agree this is a much more elegant solution. The unit-test would also be a great addition. Thank you

doronhi:
Couldn't find a wrapper for pcl::fromROSMsg in python and since my current unit tests are in python a unit-test for this will have to wait a little longer. Merging as it is.

joelbudu:
Thanks for this @doronhi . Is it possible to merge into the eloquent branch as well? I know it's EOL but it's what we're using currently for Ubuntu 18.04. Would be grateful.

PR #1682, opened 2021-02-10
Eloquent - Fix ordered pointcloud2 format

doronhi:
Fixed ordered pointcloud: set _msg_pointcloud.width before resize.
fix issue #1664. Based on #1676


doronhi:
In response to #1676 (https://github.com/IntelRealSense/realsense-ros/pull/1676#issuecomment-775286389)

Issue #1491, opened 2020-11-09
URDF and Meshes for D455?


sisaha9:
Where can we find the URDF and Meshes for Intel Realsense D455 camera?

MartyG-RealSense:
Hi @sisaha9   The **urdf** and the **meshes** folders of the RealSense ROS wrapper do not currently list files for the D455.

https://github.com/IntelRealSense/realsense-ros/tree/development/realsense2_description

The official RealSense CAD file archive lists a STEP format model for the D455 though if that is of any help.

https://dev.intelrealsense.com/docs/cad-files

MartyG-RealSense:
Hi @sisaha9   Do you still require assistance with this case, please?   Thanks!

MikeHallettUK:
Yes please ... I also have a D455 ... do we just copy / rename the D435i and IMU .xacro files ?

Or can you please add these D455 equivalents into realsense-ros/realsense2_description/urdf/   ?

Thanks

MartyG-RealSense:
Hi @MikeHallettUK  This request may be best handled as a **Pull Request** on this GitHub to add D455 urdf support, especially as @doronhi the RealSense ROS wrapper developer has just added an urdf PR for the L515 camera model.

https://github.com/IntelRealSense/realsense-ros/pulls

MartyG-RealSense:
Hi @sisaha9  Do you require further assistance with this case, please?   Thanks!

MikeHallettUK:
No thanks ... I did do this as a Pull Request.  Thanks

MartyG-RealSense:
Thanks @MikeHallettUK   The question was to @sisaha9 as they created this discussion.  Thanks very much for your pull   :)

sisaha9:
Yup we can close it. Sorry for the delay and thanks!

MartyG-RealSense:
Thanks very much for the update!

doronhi:
@MikeHallettUK 

> No thanks ... I did do this as a Pull Request. Thanks

Hi @MikeHallettUK,
I'm sorry if I understand your meaning incorrectly, but did you submit a PR? I couldn't find any by your name. If you did, could you share a link please? If not, I'm sorry and please ignore this.

MikeHallettUK:
[...]

PR #1705, opened 2021-02-17
D455 description

LuAPi:
Add a urdf for the d455.

This is  copy-edited version of the d435 urdf. I think I got the correct measurements but you should probably double check.
It currently doesn't have a mesh. I'm working on that but it's a fairly low priority so someone else might want to add that.

Adding both this and a d435 to the same robot will cause problems without also merging https://github.com/IntelRealSense/realsense-ros/pull/1637


doronhi:
Thanks. I'll look into it and into the missing mesh.

felixvd:
Sadly I didn't see this PR in https://github.com/IntelRealSense/realsense-ros/issues/1491, so I just duplicated this work. Please use the simplified mesh from my fork: https://github.com/felixvd/realsense-ros/blob/development/realsense2_description/meshes/d455.stl

doronhi:
Thanks for the added mesh.
Something is still not positioned right:
image
The camera_link coordinate system should not be positioned on the glass but rather 4.55mm from it. See **Table 4-11. Depth Cameras Depth Start Point** in [Intel-RealSense-D400-Series-Datasheet-February-2021](https://www.intelrealsense.com/download/15409/?_ga=2.118400585.1764762426.1616396434-146884591.1608633837)
I didn't go further with my examination. Could you verify and/or fix?

doronhi (changes requested):
[See comment](https://github.com/IntelRealSense/realsense-ros/pull/1705#issuecomment-803952279)

JHPHELAN:
I would think the link/joint should be at the back of the camera at or between the mounting screws?

doronhi:
This PR was fixed and merged in #2137.
Thank you for your contribution.

PR #1752, opened 2021-03-12
Remove wrong dependency

ehddl34:
Hi.

This is this is my first Pull Request. So please understand if I did something wrong.

I tested the foxy branch of realsense-ros . I tried to build source code and there is a build error related dependency.
diagnostic_updater is not exist in foxy. So, it will not need anymore in foxy branch.
Please consider to fix this.


doronhi:
The diagnostic updater feature exists in the ROS1 versions and I hope to migrate it into the ROS2 versions as soon as I manage to. In the meanwhile, I agree, a code should be as clean as possible.
By the way, I faced no building errors related to the diagnostic updater in foxy.

PR #1792, opened 2021-04-06
Update NOTICE

Infinite-Blue-1042:


doronhi:
Could you provide an explanation, please? Why is this update important?

PR #1802, opened 2021-04-09
Add conditional param use_mesh

Teo-Cardoso:
This pull request adds the option to use the param use_mesh to load the 3D files in URDF or not.

So, now we're able to turn on/off the mesh easily:
```
  <xacro:sensor_d435 
          parent="base_link"
          use_nominal_extrinsics="$(arg use_nominal_extrinsics)" 
          add_plug="$(arg add_plug)" 
          use_mesh="$(arg use_mesh)">

    <origin xyz="0 0 0" rpy="0 0 0"/>
  </xacro:sensor_d435>
```
I have been facing performance issues with the meshes then I added this option.


doronhi:
Looking good. Is there a reason why in _d415.urdf.xacro the default value for use_mesh is false while in _d435.urdf.xacro, _l515.urdf.xacro it is true?

Teo-Cardoso:
I set the value wrong :/, but I fixed it now. The value should be true by default to keep compatible.

PR #1806, opened 2021-04-12
ROS2 align pc

doronhi:
modified align_depth option.
if set to true, pointcloud will be build based on the "aligned depth to color" image.
aligning depth image to streams other than color is disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ For the entire list of parameters type `ros2 param list`.
 
 - **rosbag_filename**: Will publish topics from rosbag file.
 - **initial_reset**: On occasions the device was not closed properly and due to firmware issues needs to reset. If set to true, the device will reset prior to usage.
-- **align_depth**: If set to true, will publish additional topics with the all the images aligned to the depth image.</br>
-The topics are of the form: ```/camera/aligned_depth_to_color/image_raw``` etc.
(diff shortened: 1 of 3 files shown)


joelbudu:
Just a quick question about the frame the aligned depth to color pointcloud is published on, both the rgbd.launch package and the deprecated ros2_intel_realsense package publish on the camera_color_optical_frame TF. Is this something the can be applied in this case?

Or rather if I may ask - what does the camera_aligned_depth_to_color_frame represent please?

@doronhi

doronhi:
> Just a quick question about the frame the aligned depth to color pointcloud is published on, both the rgbd.launch package and the deprecated ros2_intel_realsense package publish on the camera_color_optical_frame TF. Is this something the can be applied in this case?
> 
> Or rather if I may ask - what does the camera_aligned_depth_to_color_frame represent please?
> 
> @doronhi

Thank you, that seems like a bug. Fixed.

joelbudu:
@doronhi I found an issue where the realsense2_camera_node sometimes throws a segmentation fault when the node is started and align_depth is set to true.
I'm not sure what causes it but, and seems to be happening randomly, but I'm suspecting it may have something to do with having a subscriber to the /camera/depth/color/points pointcloud topic

doronhi:
> @doronhi I found an issue where the realsense2_camera_node sometimes throws a segmentation fault when the node is started and align_depth is set to true.
> I'm not sure what causes it but, and seems to be happening randomly, but I'm suspecting it may have something to do with having a subscriber to the /camera/depth/color/points pointcloud topic

So far couldn't reproduce.

Issue #1827, opened 2021-04-28
depth_qos parameter not applied to depth pointcloud publisher QoS


anaelle-sw:
Hello,
I found out a problem with a publisher created without taking into account the parameter `<stream_type>_qos` given to the node.

## Setup
- D435
- Ubuntu 20.04, ROS2 Rolling
- realsense-ros repo forked from branch ros2

## Problem description
I am willing realsense-ros node to publish the D435 depth pointcloud not with the default QoS, but with the reliability set to best-effort (which is more adapted to sensor data exchange). So in the launch file, I set the parameter `depth_qos` to `'SENSOR_DATA'` and pass it to the node. Here are the interesting parts of the launch file:
```c
    enable_depth = LaunchConfiguration('enable_depth', default = 'true')
    depth_qos = LaunchConfiguration('depth_qos', default = 'SENSOR_DATA')
    enable_pointcloud = LaunchConfiguration('enable_pointcloud', default = 'true')

[...]

    realsense_node = Node(
        package = 'realsense2_camera',
        executable = 'realsense2_camera_node',
        name = 'realsense_node',
        parameters = [
            {...},
            {'enable_depth': enable_depth},
            {'depth_qos': depth_qos},
            {'enable_pointcloud': enable_pointcloud},
            {...}
           ]
        )

[...]
```
But when launching the node, and doing `ros2 topic info /points -v` (topic "/points" being the topic where realsense node publishes the depth pointcloud), I found out that the topic is advertised with reliable QoS, and not best-effort.

By looking at [realsense2_camera/src/base_realsense_node.cpp](https://github.com/IntelRealSense/realsense-ros/blob/0ea1b212b5e2b5f6ce67b74270c12cba9f126e68/realsense2_camera/src/base_realsense_node.cpp#L1022), I found out that, when setting `enable_depth = true` and `enable_pointcloud = true`, the publisher for the depth pointcloud is created without using the parameter `depth_qos`, but with default QoS (with a queue of 1).

The easy way to fix this just consists in passing the custom QoS for the stream when creating the publisher:
```c
_pointcloud_publisher = _node.create_publisher<sensor_msgs::msg::PointCloud2>(
     "depth/color/points", 
      rclcpp::QoS(
           rclcpp::QoSInitialization(
                qos_string_to_qos(_qos[stream]).history,
                qos_string_to_qos(_qos[stream]).depth),
           qos_string_to_qos(_qos[stream]));
```

## Additional information
[...]

PR #1833, opened 2021-04-29
QoS parameters to be applied for all publishers

anaelle-sw:
Hello,

This pull request is meant to allow the user to set the QoS of all the publishers of realsense-ros nodes, by setting QoS parameters.

It addresses this issue: "depth_qos parameter not applied to depth pointcloud publisher QoS" #1827 
and it follows the logic of this previously merged PR: "Add QoS in realsense parameters list" #1597

It is backward compatible, as the previous values set for the QoS are kept as default (see [here](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L684) and [here](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/include/constants.h#L87)).

Here are the modified publishers:

* [_`pointcloud_publisher`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1083): set by the parameter `pointcloud_qos` (default QoS: `rmw_qos_profile_default`with depth = 1)

* [`_imu_publishers`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1108) and [`_odom_publisher`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1126): set by parameters `gyro_qos`, `accel_qos`, and ~~`odom_qos`~~ `pose_qos`(default QoS: `rmw_qos_profile_default`with depth = 100)

* [`_synced_imu_publisher`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1096): set by parameter `imu_qos` (default QoS: `rmw_qos_profile_default` with depth = 5)

* [`_info_publisher`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1059), [`_depth_aligned_info_publisher`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1074), and [`_imu_info_publisher`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L2365): set by parameters `<stream_type>_info_qos` (default QoS: `rmw_qos_profile_default` with depth = 1)

* [`__depth_to_other_extrinsics_publishers`](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/src/base_realsense_node.cpp#L1136): set by parameters `<stream_type>_extrinsics_qos` (default QoS: `rmw_qos_profile_latched`, which [definition](https://github.com/wyca-robotics/realsense-ros/blob/8d7ab1382366073d9240ed077b7f1dd8c94a6cce/realsense2_camera/include/base_realsense_node.h#L73) has been move to `base_realsense_node.cpp`)

I am not sure of the interest of customizing the "info" and "extrinsic" topics'  QoS. The changes related to these are done is distinct commits to revert them easily if you think there is no reason to change these QoS. Let me know what you think about this.

Also, should I modified the README.md file to match these changes and document the new parameters?

Finally, I haven't been able to test everything yet. I will let you know when it will be done.

Thanks!


anaelle-sw:
The following modified topics have been tested on a D435:
- points
- depth/camera_info
- color/camera_info
- infra1/camera_info
- infra2/camera_info
- align_depth_to_color/camera_info
- align_depth_to_infra/camera_info
- extrinsics/depth_to_color
- extrinsics/depth_to_infra1
- extrinsics/depth_to_infra2

For each, the QoS can be correctly set by passing the corresponding parameter in the launch file, or directly via CLI

I will come back later with the results on a T265 for the topics about the fisheye / IMU streams

anaelle-sw:
Here are the topics that have been tested on a T265:
- accel/imu_info
- gyro/imu_info
- fisheye1/camera_info
- fisheye2/camera_info
- align_depth_to_fisheye/camera_info
- imu
- gyro/sample
- accel/sample
- odom/sample 

The QoS can also be set correctly for each one.

I think all the publishers that are modified by this PR has been tested now. Let me know if I forgot anything. Thanks!

doronhi (changes requested):
Thank you for demonstrating the modification of QoS outside of image_transport.
The changes you suggest brings up second thoughts regarding the many types of QoS used in the code.
Also, Yes, please add usage remarks to the README.md file.

anaelle-sw:
I updated the list of QoS parameters in the README.md. 
The [changes you requested in the review](https://github.com/IntelRealSense/realsense-ros/pull/1833/files/8d7ab1382366073d9240ed077b7f1dd8c94a6cce) seems just to group all the commits of this PR, but I may have missed something. Could you please list the changes you wanted? Thank you!

realsense2_camera/include/base_realsense_node.h
```diff
@@ -70,6 +70,19 @@ namespace realsense2_camera
 
     const std::vector<stream_index_pair> HID_STREAMS = {GYRO, ACCEL, POSE};
 
+    static const rmw_qos_profile_t rmw_qos_profile_latched =
```
doronhi: I think the definition of *rmw_qos_profile_latched* can be in the .cpp file, say before BaseRealSenseNode::BaseRealSenseNode(). It it not related to anything outside the class.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -689,6 +689,32 @@ const rmw_qos_profile_t BaseRealSenseNode::qos_string_to_qos(std::string str)
 #endif
     if (str == "SYSTEM_DEFAULT")
         return rmw_qos_profile_system_default;
+    if (str == "POINTCLOUD_DEFAULT")
```
doronhi: pointcloud and info can use the same QoS : DEFAULT. All they change is the depth, from the original 10 to 1. I prefer, as much as we can, not to add new QoS options.
IMU can use HID_DEFAULT.

doronhi (changes requested):
I must have made a mistake last time I reviewed.
Generally, your PR brought the fact that we use too many variations of QoS to my attention. I think this is a historical choice, not really tested for proving a benefit so I think we can refrain from adding extra fixed QoS options to the system.

anaelle-sw:
I rebased this PR's branch into the newest ROS2 release ([3.2.1](https://github.com/IntelRealSense/realsense-ros/releases/tag/3.2.1))

realsense2_camera/src/base_realsense_node.cpp
anaelle-sw: I agree that too much QoS profiles can be confusing and not useful. The reason I declared distinct default QoS was mainly to ensure backward compatibility, and also to keep the different kind of publishers independent. If someday the default value of depth for the info topics need to be changed, it shouldn't be impacting the pointcloud publisher's depth, for instance. 

But I guess it can be managed when this will be needed (if it is needed one day), and I agree it may be better to keep it simple with this PR.

I did the changes you asked for [here for the info and pointcloud publishers](https://github.com/IntelRealSense/realsense-ros/pull/1833/commits/5b344b9a88445efecf12b5e6d9268f0d3ed19536) and [here for the IMU publisher](https://github.com/IntelRealSense/realsense-ros/pull/1833/commits/773fdafbd58fcdbf599a416ed5e028fd681eb9bb). My main concern is still about the backward compatibility of the publishers' depths but you seemed to say it shouldn't be a problem to change the default depth. Some parameters could be opened for the user to set the publishers' depths. But again, too much settings could be confusing. Could you just confirm it is good for you this way?

realsense2_camera/include/base_realsense_node.h
anaelle-sw: The changes are [here](https://github.com/IntelRealSense/realsense-ros/pull/1833/commits/b8aa13aea7246c09eefa77af578384f8c792f7ac)

> It it not related to anything outside the class.

Actually, in this case, shouldn't be better to declare this profile as an attribute of the class `BaseRealSenseNode::BaseRealSenseNode` in the .h file ?

doronhi: Why do you think it should be better?

README.md
```diff
@@ -158,7 +158,7 @@ The pointcloud, if enabled, will be built based on the aligned_depth_to_color im
...
-- ***<stream_type>*_qos**: <stream_type> can be any of *infra, color, fisheye, depth, gyro, accel, pose, confidence*. Sets the QoS by which the topic is published. Available values are the following strings: SYSTEM_DEFAULT, PARAMETER_EVENTS, SERVICES_DEFAULT, PARAMETERS, DEFAULT, SENSOR_DATA.
+- ***<stream_type>*_qos**: <stream_type> can be any of *infra, color, fisheye, depth, gyro, accel, pose, confidence, pointcloud, gyro, accel, pose, imu_info, infra_info, color_info,fisheye_info, depth_info, gyro_info, accel_info, pose_info, fisheye_extrinsics, color_extrinsics, infra_extrinsics*. Sets the QoS by which the topic is published. Available values are the following strings: SYSTEM_DEFAULT, PARAMETER_EVENTS, SERVICES_DEFAULT, PARAMETERS, DEFAULT, SENSOR_DATA.
```
doronhi: Do you think we can avoid mentioning all the topics and just say that there is a parameter for <stream_type>_ros, <stream_type>_info_ros and <stream_type>_extrinsics?
Also, *HID_DEFAULT* and *EXTRINSICS_DEFAULT* are missing. We should inform the users what are their characteristics:
EXTRINSICS_DEFAULT
{
    RMW_QOS_POLICY_HISTORY_KEEP_LAST,
    1,
    RMW_QOS_POLICY_RELIABILITY_RELIABLE,
    RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
    RMW_QOS_DEADLINE_DEFAULT,
    RMW_QOS_LIFESPAN_DEFAULT,
    RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
    RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT,
    false
};
HID_DEFAULT = DEFAULT with depth of 100.

doronhi (changes requested):
I agree that we may need to give more freedom in the selection of parameters inside the different QoS. However, this is for another, future, PR.
For this one, I just think we need some clarifications in the README.md file.

realsense2_camera/include/base_realsense_node.h
anaelle-sw: Generally, global variables are avoided when possible, especially when there are not related to anything outside some classes (like this one, as you said). So if there is no need to access `rmw_qos_profile_latched` outside from `BaseRealSenseNode`, why should it be declared outside of it, and not be a class member?

README.md
anaelle-sw: I fixed this [here](https://github.com/IntelRealSense/realsense-ros/pull/1833/commits/c59faa7a5a5a62e21b747b7f7f73f633f9e838b8)
Instead of giving all the details for `EXTRINSICS_DEFAULT` (because there is no detail for all the rest of default QoS), the description is just:
*EXTRINSICS_DEFAULT (= DEFAULT with depth of 1 and transient local durability)*.
 Is it ok for you?

doronhi: It looks very good to me. Thanks.

doronhi (approved):

PR #1835, opened 2021-04-29
Facilitate Readme: Use ROS_DISTRO env variable

tkazik:
No more distinction for OS version required for installation.


doronhi:
Thanks.

Issue #1310, opened 2020-07-28
Baseline calculation problem in rtabmap stereo with camera info from D435i


BarzelS:
Hi,
I'm getting this error in rtabmap  odometry stereo:
The stereo baseline (0.0) should be positive (baseline=-Tx/fx). We assume a horizontal left/right stereo "
"setup where the Tx (or P(0,3)) is negative in the right camera info msg."
It takes data from both /camera/infra1/camera_info and /camera/infra2/camera_info to calculate based on this equation:
baseline =  right_.fx()!=0.0 && left_.fx() != 0.0 ? left_.Tx() / left_.fx() - right_.Tx()/right_.fx():0.0;
I found out that both left_.Tx and right_.Tx in the camera info topics are both 0.0, 
How can it be if the baseline should be 0.05m(The distance between the infra camera)?

doronhi:
The information regarding the baseline does not appear in the camera info as each cameraInfo topic describes a single sensor. It appears in the tf messages.
This is changed in #1242 . Does that PR match your needs?

BarzelS:
> The information regarding the baseline does not appear in the camera info as each cameraInfo topic describes a single sensor. It appears in the tf messages.
> This is changed in #1242 . Does that PR match your needs?

If I understand correctly from the PR the data in "/camera/infra1/camera_info" and /camera/infra2/camera_info" is currently not correct(RealSense ROS v2.2.15) and the mentioned PR will solve it?
If that's the case so yes the PR match my needs.

doronhi:
This issue is being discussed for a long time now. The main reason is that I wan't convinced what "correct" is, nor how to check it.
Therefor I will greatly appreciate if you managed to give it a try and see if it works well with rtabmap odometry stereo.

BarzelS:
It's not parameters that should be obtained from the Intel-Realsense hardware team?
I will try to compile the PR and check it

doronhi:
I'd like to know if that PR works for you.
In any case, the information published on the /tf topic is based on the extrinsics given by the device.

BarzelS:
> I'd like to know if that PR works for you.
> In any case, the information published on the /tf topic is based on the extrinsics given by the device.

I've updated to the PR you mentioned, it seems that now the baseline in not 0.0 but I'm still getting this error:
[...]

PR #1840, opened 2021-05-01
Updated Stereo extrinsics to fix incorrect baseline sign

mindThomas:
Follow-up PR to correct the baseline sign of the Stereo extrinsics as discussed in https://github.com/IntelRealSense/realsense-ros/issues/1310
This PR also addresses the issue with the Projection matrix not getting updated if the Intrinsics are changed: https://github.com/IntelRealSense/realsense-ros/pull/1242#issuecomment-785778234

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1952,13 +1952,13 @@ void BaseRealSenseNode::updateExtrinsicsCalibData(const rs2::video_stream_profil
     stream_index_pair right{right_video_profile.stream_type(), right_video_profile.stream_index()};
 
     // Get the relative extrinsics between the left and right camera
-    auto LEFT_T_RIGHT = right_video_profile.get_extrinsics_to(left_video_profile);
+    auto LEFT_T_RIGHT = right_video_profile.get_extrinsics_to(left_video_profile); // transformation to transform a point in RIGHT frame to LEFT frame
+    auto RIGHT_T_LEFT = left_video_profile.get_extrinsics_to(right_video_profile); // transformation to transform a point in LEFT frame to RIGHT frame
 
-    auto R = Eigen::Map<Eigen::Matrix<float,3,3,Eigen::RowMajor>>(LEFT_T_RIGHT.rotation);
-    auto T = Eigen::Map<Eigen::Matrix<float,3,1>>(LEFT_T_RIGHT.translation);
+    auto R = Eigen::Map<Eigen::Matrix<float,3,3,Eigen::RowMajor>>(RIGHT_T_LEFT.rotation);
+    auto T = Eigen::Map<Eigen::Matrix<float,3,1>>(RIGHT_T_LEFT.translation);
 
-    // force y- and z-axis components to be 0   (but do we also need to force P(0,3) and P(1,3) to be 0?)
-    T[1] = 0;
+    // force and z-axis components to be 0 (if it wasn't already)    
     T[2] = 0;
 
     Eigen::Matrix<float,3,4,Eigen::RowMajor> RT;
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1812,9 +1812,6 @@ void BaseRealSenseNode::setupStreams()
 {
 	ROS_INFO("setupStreams...");
     try{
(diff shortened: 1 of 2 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -1952,13 +1949,13 @@ void BaseRealSenseNode::updateExtrinsicsCalibData(const rs2::video_stream_profil
     stream_index_pair right{right_video_profile.stream_type(), right_video_profile.stream_index()};
 
     // Get the relative extrinsics between the left and right camera
-    auto LEFT_T_RIGHT = right_video_profile.get_extrinsics_to(left_video_profile);
+    auto LEFT_T_RIGHT = right_video_profile.get_extrinsics_to(left_video_profile); // transformation to transform a point in RIGHT frame to LEFT frame
+    auto RIGHT_T_LEFT = left_video_profile.get_extrinsics_to(right_video_profile); // transformation to transform a point in LEFT frame to RIGHT frame
```
doronhi, line 1953: Shouldn't RIGHT_TO_LEFT also be used in line 1972-1980 instead of LEFT_TO_RIGHT? It doesn't make an immediate difference as there is no rotation but still?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -1826,14 +1823,14 @@ void BaseRealSenseNode::setupStreams()
                     updateStreamCalibData(video_profile);
 
                     // stream index: 1=left, 2=right
-                    if (video_profile.stream_index() == 1) { left_profile = std::make_shared<rs2::video_stream_profile>(video_profile); }
-                    if (video_profile.stream_index() == 2) { right_profile = std::make_shared<rs2::video_stream_profile>(video_profile);  }
+                    if (video_profile.stream_index() == 1) { _left_stereo_profile = std::make_shared<rs2::video_stream_profile>(std::move(video_profile)); }
```
doronhi, line 1826: Why use std::move? We are creating a copy of the profile here, not removing it from the _enabled_profiles list.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -2453,8 +2450,18 @@ void BaseRealSenseNode::publishFrame(rs2::frame f, const ros::Time& t,
         auto& cam_info = camera_info.at(stream);
         if (cam_info.width != width)
         {
-            updateStreamCalibData(f.get_profile().as<rs2::video_stream_profile>());
+            auto video_profile = f.get_profile().as<rs2::video_stream_profile>();
+            updateStreamCalibData(video_profile);
+
```
doronhi, line 2455: _left_stereo_profile, _right_stereo_profile were already assigned in BaseRealSenseNode::setupStreams(), lines 1826-1827. 
publishFrame() gets called with one frame so if  (cam_info.width != width) only one of the pair _left_stereo_profile, _right_stereo_profile will be updated. It seems to me that both should be reset after the last call to updateExtrinsicsCalibData().

doronhi (commented):
I appreciate the work. I understand the current shortage of a camera to test with although I feel it could help.

doronhi:
I ended up merging a modified #901. I think your conversation with @pavloblindnology made the whole thing possible and appreciate the amount of time and effort you put into this.
Thanks,
doronhi

realsense2_camera/src/base_realsense_node.cpp
mindThomas: I don't think so. But based on the ROS documentation it is not all clear which frames the rotation matrix is defined between.
In this PR the rotation matrix can be applied to rotate a point from the right camera frame to the left camera frame, which is chosen as the common frame.

realsense2_camera/src/base_realsense_node.cpp
mindThomas: The `video_profile` object is created as a temporary stream profile and since we won't be using that beyond line 1826-1827 there is no point in creating a completely new `video_stream_profile` object. Hence we can just move `video_profile` data into the shared object. To summarize, this trick is done to avoid an unnecessary memory allocation which would otherwise be performed by `std::make_shared`.

realsense2_camera/src/base_realsense_node.cpp
mindThomas: That is correct, but `publishFrame()` would be called again right after with the other frame, thereby calling the `updateExtrinsicsCalibData()` again.

Issue #1894, opened 2021-05-25
Using default NodeHandle instead of private NodeHandle for setting parameters


Marenix:
Hi. I've been going over the code for the wrapper, and noticed that in base_realsense_node.cpp in function publishTopics, you're giving function registerDynamicReconfigCb _node_handle (which is a default node handle). This function basically goes on to setting parameters. What I'm wondering is why you used a public node handle? As far as I know, it's recommended to use a private node handle when working with parameters, so I'm curious if there's a reason you used a public node handle specifically?

doronhi:
No reason that I am aware of. This is an old part of the code. If I am convinced there is a good enough reason to change I'll change.

Marenix:
It would make setting default parameters from .launch file like this possible (also, adding the appropriate lines in nodelet.launch.xml):
![rs_camera1](https://user-images.githubusercontent.com/39246140/119614798-00a42680-bdff-11eb-9a17-2e6b7bd5456a.png)
![rs_camera2](https://user-images.githubusercontent.com/39246140/119614806-03068080-bdff-11eb-85ed-4884413ab386.png)

It might not be that important for dynamic parameters (like exposure in my example), since they can be set after the streams start, but the problem is with parameters like inter_cam_sync_mode which can't be set once the streams start. 
I tested the default code with this .launch file and parameters weren't changed (looked at the stream in rviz just to make sure the inter_cam_sync_mode wasn't set):
![rqt1](https://user-images.githubusercontent.com/39246140/119616128-8d031900-be00-11eb-94ba-c7d597628f4a.png)
![rviz1](https://user-images.githubusercontent.com/39246140/119616147-8ffe0980-be00-11eb-9e1d-1415d31217cf.png)

Then I changed to a private node handle in function publishTopics(passing _pnh to functions called in publishTopics instead of _node_handle), and the parameters were set (exposure to 500 and inter_cam_sync_mode to 4, no image in rviz):
![rqt2](https://user-images.githubusercontent.com/39246140/119616697-34804b80-be01-11eb-95b4-fb7d393d8c19.png)
![rviz2](https://user-images.githubusercontent.com/39246140/119616831-5bd71880-be01-11eb-893d-71a7e7d35192.png)

doronhi:
That sounds very interesting. Could you submit a PR so I could test it more thoroughly?
It might take me a little while to get around to it but it sounds important.

Marenix:
[...]

PR #1901, opened 2021-05-27
Changed parameters setting from public to private node handle

Marenix:
Added setting of inter_cam_sync_mode and exposure in .launch file

**(Changes to nodelet.launch.xml and rs_camera.launch are meant only to show the new functionality available with using a private instead of a public node handle for setting parameters. These files should be changed back to their original state before merging)**

EDIT: Related to issue [#1894](https://github.com/IntelRealSense/realsense-ros/issues/1894)

### DIFF
--- a/realsense2_camera/config/my_config_file.yaml
+++ b/realsense2_camera/config/my_config_file.yaml
@@ -1,0 +1,92 @@
+  serial_no: ""
+  usb_port_id: ""
+  device_type: ""
+  rosbag_filename: ""
+  initial_reset: false
+
+  json_file_path: ""
+  rosbag_filename: ""
+  required: false
+
+  fisheye_width: 0
+  fisheye_height: 0
+  enable_fisheye: false
+  enable_fisheye1: false
+  enable_fisheye2: false
+
+  depth_width: 640
+  depth_height: 480
+  enable_depth: true
+
+  infra_width: 640
+  infra_height: 480
+  enable_infra1: false
+  enable_infra2: false
+
+  color_width: 640
+  color_height: 480
+  enable_color: true
+
+  fisheye_fps: 30
+  depth_fps: 30
+  infra_fps: 30
+  color_fps: 30
+  gyro_fps: 0
+  accel_fps: 0
+  enable_gyro: false
+  enable_accel: false
+
+  enable_pointcloud: true
+  pointcloud_texture_stream: "RS2_STREAM_COLOR" #use RS2_STREAM_ANY to avoid using texture
+  pointcloud_texture_index: 0
+
+  enable_sync: false
+  align_depth: false
+
+  base_frame_id: "camera1_link"
+  depth_frame_id: "camera1_depth_frame"
+  infra1_frame_id: "camera1_infra1_frame"
+  infra2_frame_id: "camera1_infra2_frame"
+  color_frame_id: "camera1_color_frame"
+  fisheye_frame_id: "camera1_fisheye_frame"
+  fisheye1_frame_id: "camera1_fisheye1_frame"
+  fisheye2_frame_id: "camera1_fisheye2_frame"
+  accel_frame_id: "camera1_accel_frame"
+  gyro_frame_id: "camera1_gyro_frame"
+  pose_frame_id: "camera1_pose_frame"
+
+  depth_optical_frame_id: "camera1_depth_optical_frame"
+  infra1_optical_frame_id: "camera1_infra1_optical_frame"
+  infra2_optical_frame_id: "camera1_infra2_optical_frame"
+  color_optical_frame_id: "camera1_color_optical_frame"
+  fisheye_optical_frame_id: "camera1_fisheye_optical_frame"
+  fisheye1_optical_frame_id: "camera1_fisheye1_optical_frame"
+  fisheye2_optical_frame_id: "camera1_fisheye2_optical_frame"
+  accel_optical_frame_id: "camera1_accel_optical_frame"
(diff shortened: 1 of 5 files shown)


doronhi:
In several issues we noted that for setting dynamic parameter on startup you could add the following pattern in the rs_camera.launch:

  <rosparam>
      /camera/stereo_module/emitter_on_off: true
  </rosparam>

Your suggestion makes this format impossible and replaces it with asking the users to modify the nodelet.launch.xml, as shown in your example.
My impression is that the current state is more user-friendly. As not all the available parameters are defined in nodelet.launch.xml, it allows users to modify just their own launch files when they wish to set a different parameter.

Marenix:
Thanks for checking this @doronhi 

You're right about adding "/camera/stereo_module/emitter_on_off: true" not working, but this is expected. Using a private node handle changes the namespace a bit, and as you were using the "old" namespace, you couldn't set the parameter. Try using the following pattern in the rs_camera.launch:
/camera/realsense2_camera/stereo_module/emitter_on_off: true

You can also see this change in rqt_reconfigure tool.

Marenix:
Hey @doronhi
Did you get the chance to read my last comment and try the suggestion I wrote there?

doronhi:
Hi @Marenix 
Sorry for the time it took me to come back to you.
I tried changing the parameters while using the existing code and it works by adding the following lines to rs_camera.launch:
```
  <rosparam>
    /camera/stereo_module/inter_cam_sync_mode: 4
  </rosparam>
```
I see that the message: `Hardware Notification:Depth stream start failure` and this is a good sign as 4 in not a valid value.

I then update to private node handler, as you suggest and now I have the parameters names changed so I modify the launch file:
```
  <rosparam>
    /camera/realsense2_camera/stereo_module/inter_cam_sync_mode: 4
  </rosparam>
```
And everything is Ok. Everything works just the same.

It is possible that what you say is true, that it is recommended to use private node handlers (and then I'll ask you to please refer me to where it is recommended) and I may agree that the structure of parameters looks better and more coherent this way,  but it is an interface change.
So far I failed to see what is not working with the current code and I am reluctant to introduce interface changes without a usage benefit. A benefit might be a good reason why it is recommended, a showcase where the current situation causes conflicts or where some parameters cannot be set at the initialization. Any of these or, any other showcase, may help me understand the situation better.

I appreciate it.

Marenix:
Hi @doronhi
I took some time to try a few things again. What you wrote works in both cases, just as you said it. What doesn't work is adding some parameters in launch file the way I did it. I admit, the way you did it is a lot more intuitive and easier to understand. It's also better in a way that you just add it to one place, and you're done.
However, what I found wasn't connected specifically to launch file. I've been using a config file. I've transferred almost all camera parameters there, and am using launch file to load that config file. That's where I first noticed the mentioned problem, where I couldn't set inter_cam_sync_mode if public node handle was used. I just decided to show the problem in launch file, because there is no config file originally in the repository. 

Now, if someone is using a config file, they can't set inter_cam_sync_mode there. The way you changed that parameter works, and that's a good way of setting the parameter when using a launch file, but if that were used while using a config file, it would mean a user would have to modify both the config and launch file to set all the options.

As for whether it's recommended or not using private node handles, I based it on a note I found on one of the official ROS tutorials: http://wiki.ros.org/ROSNodeTutorialC%2B%2B (chapter 3.1). I haven't really come across a case where this would cause me any errors or problems, but I haven't done any extensive testing on it, so I can't say if such a case is likely to happen or not.

doronhi:
Hi @Marenix ,
I appreciate the work you put into this already. I read the comment you referred me to and I'll investigate some more. In the meanwhile, could you please share a config file and a command line you use for reproducing the issue? (Or a shorter version if you prefer)
Thanks

Marenix:
Hi @doronhi,
I added the used config and launch file. I use the following to start it:
`roslaunch realsense2_camera my_rs_camera.launch`
With private node handle, inter_cam_sync_mode is changed, whereas using the public node handle, does not appear to change it. inter_cam_sync_mode is set at the end of the config file, so you can change it there.

doronhi:
Thanks @Marenix ,

I tested with the official version - the one using the public node handles.
In "my_config_file.yaml" file, replacing the lines:
```
  stereo_module:
    inter_cam_sync_mode: 259
```
with this one: 
`/camera/stereo_module/inter_cam_sync_mode: 411`
works.

The comment in the tutorial, chapter 3.1, refers to an issue with running multiple instances of the node. This situation is already dealt with by having all the parameters have a namespace prefix, defined for each instance of the node by the "camera" parameter.
It could be that in retrospect the dynamic parameters should have been written using the private node handle, just as the ROS parameters are written (For instance, "align_depth" uses a private node handle). It could be, and if it came down to it I would investigate the subject some more.
Still, without a failing valid use-case or a functional improvement, I don't see a strong enough reason to introduce an API change.

Marenix:
Hi @doronhi,

Thanks for checking this out. I hadn't thought of trying to add something like:
`/camera/stereo_module/inter_cam_sync_mode: 411`
in config file. You're right, setting the parameter that way work.

If I come across any failing valid use-case in my testing, I'll let you know. But, as you said, parameters already have their own prefixes, so I don't think there should be any clashing of parameter names, and the way you showed me to set parameters works in both cases, so I don't see any problems there now.

Thanks again for taking the time to look at this.

doronhi:
I am closing this PR now.
I really appreciate and thank you for pursuing the subject so and for introducing me to some of the finer details of the node.

Issue #1913, opened 2021-06-04
Passing config_file arg at launch results on failing to parse the argument


TSC21:
Trying `ros2 launch realsense2_camera rs_launch.py config_file:="'<abs_path_to_file>/d435i.yaml'"` results on:

```sh
[WARNING] [launch_ros.actions.node]: Parameter file path is not a file: '<abs_path_to_file>/d435i.yaml'
```

I believe the current logic in `rs_launch.py` using the substitutions with `LaunchConfiguration` is faulty. I tried some different approach but no luck so far. I believe this needs a fix.

TSC21:
Found the solution in https://github.com/IntelRealSense/librealsense/issues/8863#issuecomment-823026235. Issuing a PR for it.

MartyG-RealSense:
Thanks very much @TSC21 - if you create a PR, please post a link to it here when it is submitted.  Thanks again!

TSC21:
> Thanks very much @TSC21 - if you create a PR, please post a link to it here when it is submitted. Thanks again!

@MartyG-RealSense here: https://github.com/IntelRealSense/realsense-ros/pull/1915. Happy to help!

MartyG-RealSense:
Great, thanks @TSC21 - I have added an **Enhancement** tag to this case to highlight that this case https://github.com/IntelRealSense/realsense-ros/issues/1913 should be kept open whilst the associated PR is active.  Thanks again!

doronhi:
Fixed in #1942

PR #1915, opened 2021-06-05
rs_launch: allow to properly set the config_path file as a parameter to load the config file stored in the shared dir

TSC21:
Addresses https://github.com/IntelRealSense/realsense-ros/issues/1913. At the same time, makes sure that one is able to use the yaml files stored at the `config` dir, which by default should be place where these get stored, and removes the need for passing the abs path to the file.

@MartyG-RealSense @doronhi for your review.

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -15,6 +15,7 @@
 """Launch realsense2_camera node."""
 import os
 from launch import LaunchDescription
+from ament_index_python.packages import get_package_share_directory
 import launch_ros.actions
 from launch.actions import DeclareLaunchArgument
 from launch.substitutions import LaunchConfiguration, PythonExpression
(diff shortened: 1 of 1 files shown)


TSC21:
Anything missing here? Can this be reviewed?

MartyG-RealSense:
@TSC21  I will tag @doronhi regarding this check-passed PR.  Thanks very much.

doronhi:
Thanks for looking into the issue and sorry for the long delay.
I think the user must be allowed to give a full path to the config file as it does not necessarily reside in the config directory.

Also, while testing, I noticed the d435i.yaml itself is faulty, with values in emitter_enabled should be set to 1 instead of true and namespace should be camera/camera and not camera1/camera to match the rs_launch.py file. The last thing is not connected to the PR of course.

doronhi:
Thanks for the PR.
I ended up merging #1942 which took your approach to fix the same parsing issue.
For now, I removed the default "config" path. I agree that that's the natural place for the config files to be but I don't want to force users to it. Any suggestion/PR for allowing the "config" directory to be the place to look for the file if the given path is not found is welcome.

TSC21:
> Thanks for the PR.
> I ended up merging #1942 which took your approach to fix the same parsing issue.
> For now, I removed the default "config" path. I agree that that's the natural place for the config files to be but I don't want to force users to it. Any suggestion/PR for allowing the "config" directory to be the place to look for the file if the given path is not found is welcome.

All good I am actually working together with that PR author so I was actually going to close this PR as that one was actually a replacement for this one. Thanks.

PR #1942, opened 2021-06-22
Reduce code in launch file

bastianhjaeger:
Reduce the lines of code. 

So far there are multiple paths depending on the existence of a `config_file` parameter, i.e. if you run

```
ros2 launch realsense2_camera rs_launch.py config_file:="'d435i.yaml'"
```
or
```
ros2 launch realsense2_camera rs_launch.py config_file:="'d435i.yaml'"
```


doronhi:
Thanks for the PR.
The solution to the parsing issue is the same as in #1915 but the "config" parameter was not set so I fixed it.
Hope you are ok with my modifications.

doronhi:
I just found out that this PR does not work with Dashing: The condition `LaunchConfiguration('config_file') != '')` returns true even if config_file is not set and then I get: `Failed to parse parameters from file '.': Error parsing a event near line 0, at /tmp/binarydeb/ros-dashing-rcl-yaml-param-parser-0.7.9/src/parser.c:1213`

Issue #1966, opened 2021-07-02
Unknown Filter. Using multiple filters like pointcloud, decimation etc.


Combinacijus:
I already found the solution but writing it for documentation purposes

So the problem was if I tried using multiple filters in [rs_camera.launch](https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/launch/rs_camera.launch) file the second filter was not recognized:
```
<arg name="filters"                   default="pointcloud, decimation"/>
```
Error:
```
[ERROR] [1625234460.526932960]: Unknown Filter:  decimation
```

SOLUTION is to list filters separated ONLY by commas no spaces allowed:
```
<arg name="filters"                   default="pointcloud,decimation"/>
```

At least for me in programming it's usual to put spaces after commas so that code looks tidier it took some time to find the problem.
I wonder is this intended behavior? Is it some xml list data type that has to be defined without spaces or this is problem with realsense string parsing code?

In [README](https://github.com/IntelRealSense/realsense-ros/blob/development/README.md#launch-parameters) it's told that parameters has to be separated by commas but it's not explicitly said that it has to have no spaces.
```
filters: any of the following options, separated by commas:
```

MartyG-RealSense:
Hi @Combinacijus  Thanks very much for sharing your feedback about defining filters!

MartyG-RealSense:
Hi @doronhi   Could you provide @Combinacijus with advice please about the reason for not having spaces after commas when defining a list of filters? Thanks!

MartyG-RealSense:
Hi @doronhi  Could you provide @Combinacijus with advice please about the reason for not having spaces after commas when defining a list of filters? Thanks!

MartyG-RealSense:
Hi @doronhi  Could you advise @Combinacijus about the reason for not having spaces after commas when defining a list of filters, please? Thanks!

Combinacijus:
Took a look at it myself and it seems it was a problem with how strings were split and iterated. [Made a pull request](https://github.com/IntelRealSense/realsense-ros/pull/1988) which removes spaces before checking which filter is selected. Not sure if it's right and correct way to do it but checked with "pointcloud, decimation" and it worked

MartyG-RealSense:
Excellent, @Combinacijus - thanks for the PR!  I have added the **Enhancement** tag to this case to signify that there is a PR associated with it.  This case should be kept open whilst that PR is active.  Thanks again   :)

[...]

PR #1988, opened 2021-07-20
Removing spaces when iterating filters

Combinacijus:
Issue #1966

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1184,8 +1184,10 @@ void BaseRealSenseNode::setupFilters()
     bool use_colorizer_filter(false);
     bool use_decimation_filter(false);
     bool use_hdr_filter(false);
-    for (std::vector<std::string>::const_iterator s_iter=filters_str.begin(); s_iter!=filters_str.end(); s_iter++)
+    for (std::vector<std::string>::iterator s_iter=filters_str.begin(); s_iter!=filters_str.end(); s_iter++)
     {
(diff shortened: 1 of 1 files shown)


doronhi:
Can you add a use case where it fails before your fix and succeed afterward?

What I mean is that the command line with spaces in the filters' string `roslaunch realsense2_camera rs_camera.launch filters:=spatial, temporal` fails even before the node is started:
```
doronhi@ubuntu:~/catkin_ws$ roslaunch realsense2_camera rs_camera.launch filters:=spatial, temporal
Usage: roslaunch [options] [package] <filename> [arg_name:=value...]
       roslaunch [options] <filename> [<filename>...] [arg_name:=value...]

If <filename> is a single dash ('-'), launch XML is read from standard input.

roslaunch: error: The following input files do not exist: temporal
```

Combinacijus:
If you put quotes around string it works:
```
roslaunch realsense2_camera rs_camera.launch filters:="spatial, temporal"
````
even something as crazy as this will work 
```
roslaunch realsense2_camera rs_camera.launch filters:="s p  a ti al, te mp  o   ral"
```
 because after it split by commas it removes all spaces and then checks what filter was selected

In your case it seems that command line interprets `temporal` as separate argument which is file name that why it doesn't even start. See the error:
> roslaunch: error: The following input files do not exist: temporal

Usually the error should be:
```
Unknown Filter:  temporal
```

What I was originally fixing was that this wouldn't work before in [rs_camera.launch](https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/launch/rs_camera.launch) but works now:
```
<arg name="filters"                   default="pointcloud, decimation"/>
```

PR #2019, opened 2021-08-03
Only send 4 distortion coeffs when using equidistant

### LINKED ISSUE #1989: Using image_proc with t265 camera_info throws openCV assertion.
To kinda preface this I found this issue but I am unsure what the best fix is. I posted this issue here as I found it was easiest to fix here and in my mind the most performant. However arguments could be made this is an issue with image_geometry and/or opencv. 

When trying to use the image proc node to rectify images from the t265 camera I got an assertion from opencv indicating that the undistort call should only have 4 distortion coefficients.
https://github.com/opencv/opencv/blob/master/modules/calib3d/src/fisheye.cpp#L465

Following calls through image geometry it does check that the camera_info distortion_model field is set to equidistant however it doesn't check the the camera_info messages number of distortion coefficients lines up with the model. 
https://github.com/ros-perception/vision_opencv/blob/noetic/image_geometry/src/pinhole_camera_model.cpp#L345
https://github.com/ros-perception/vision_opencv/blob/noetic/image_geometry/src/pinhole_camera_model.cpp#L544

When instantiating the message for the t265 the driver sets the distortion model to equidistant 
https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/src/base_realsense_node.cpp#L1909
However, it still sends 5 parameters even though the model it is representing only needs 4. Which eventually triggers the openCV assertion
https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/src/base_realsense_node.cpp#L1925

Since image geometry uses a const message I found it easiest to send out the number of distortion coefficients the model actually needs as follows.
Could this fix be made to the driver?

int coeff_size = (intrinsic.model == RS2_DISTORTION_KANNALA_BRANDT4) ? 4 : intrinsic.coeffs.size() ;
_camera_info[stream_index].D.resize(coeff_size);
for (int i = 0; i < coeff_size; i++)
{
_camera_info[stream_index].D.at(i) = intrinsic.coeffs[i];
}

CollinAvidano:
Closes #1989. As mentioned in the issue sending more than the parameters needed by the model breaks a number of other rectification packages which often treat the message as a const ptr.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1924,8 +1924,9 @@ void BaseRealSenseNode::updateStreamCalibData(const rs2::video_stream_profile& v
     _camera_info[stream_index].R.at(7) = 0.0;
     _camera_info[stream_index].R.at(8) = 1.0;
 
-    _camera_info[stream_index].D.resize(5);
-    for (int i = 0; i < 5; i++)
+    int coeff_size = (intrinsic.model == RS2_DISTORTION_KANNALA_BRANDT4) ? 4 : 5;
+    _camera_info[stream_index].D.resize(coeff_size);
+    for (int i = 0; i < coeff_size; i++)
     {
         _camera_info[stream_index].D.at(i) = intrinsic.coeffs[i];
     }


doronhi:
I thought its best to support other applications, regardless of who should check what. Thanks!

PR #2027, opened 2021-08-09
Add comment for supporting rosdep for EOL distros

utkarsh867:
The installation steps provided for ROS2 are inaccurate and may cause errors on rosdep dependencies installation.

Since Dashing has reached EOL, in order to install dependencies for Dashing, the  `sudo rosdep init --include-eol-distros` is essential and may not be obvious for first-time users.


doronhi:
Thank for pointing that out.

PR #2031, opened 2021-08-10
chore(realsense2_camera): make reconnect timeout a parameter

JaccovdS:
Currently the reconnect timeout when a device has not been found is hard coded to 6000 ms. This allows for a custom setting through a parameter. Also changed from ms to seconds since this is more convenient for a user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ The following parameters are available by the wrapper:
 
 - **rosbag_filename**: Will publish topics from rosbag file.
 - **initial_reset**: On occasions the device was not closed properly and due to firmware issues needs to reset. If set to true, the device will reset prior to usage.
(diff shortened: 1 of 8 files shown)


JaccovdS:
Can someone maybe take a look?
@doronhi do I have to ping you?

reinzor:
ping @dorodnic

doronhi:
Hi,
Sorry for not paying attention. 
The change seems reasonable. What did you set the parameter to int? why not float? (suppose a user wants to try every 6.5 seconds)

reinzor:
@doronhi thanks for your reply; converting to a float seems reasonable, we will update the PR.

doronhi:
Thanks! I appreciate you didn't neglect even the README.md file :)

PR #2035, opened 2021-08-11
correction

Captain299792458:
can remove .xacro now


doronhi:
I could use a description - what is this PR for?

PR #2095, opened 2021-09-29
[ros2] add wait_for_device_timeout

doisyg:
This PR adds a param `wait_for_device_timeout` for avoiding to wait indefinitely for a device to show up (which can cause some issues in the long run on the usb devices because of the infinite calls to `ctx.query_devices()`)
When the timeout expires, the node closes.
Set to `-1` by default (keep original behavior of waiting indefinitly)

### DIFF
--- a/realsense2_camera/include/realsense_node_factory.h
+++ b/realsense2_camera/include/realsense_node_factory.h
@@ -51,6 +51,7 @@ namespace realsense2_camera
         std::string _serial_no;
         std::string _usb_port_id;
         std::string _device_type;
+        double _wait_for_device_timeout;
         bool _initial_reset;
         std::thread _query_thread;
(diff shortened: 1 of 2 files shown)


doronhi:
Thanks,
I took the liberty to fix the wait to time to match the timeout left (i.e. if wait_for_device_timeout=13 secs and each interval sleep time is 6 - reduce the last sleep to 1 sec so the overall waiting time will match the user's specification - 13 secs)
I hope you approve.
I also added a description in the readme.md file and a line in rs_launch.py.
Merged in commit [3997254a35a9cbba14e9360c59e2d68925068329](https://github.com/IntelRealSense/realsense-ros/commit/3997254a35a9cbba14e9360c59e2d68925068329)

PR #2104, opened 2021-10-07
Fix Windows Usage

Ace314159:
This fixes the build and makes it usable on Windows. I have tested with the D435i.


Ace314159:
@doronhi It would be really nice if you could take a look at this. It's a couple of minor changes and would help my team a lot!

PR #2108, opened 2021-10-10
ROS2 - enabling tests

doronhi:


doronhi:
Migrated to github actions. Issues with tests will be dealt with in another PR.

PR #2123, opened 2021-10-19
Need cv-bridge to build the package.

saratrajput:
This is a small addition to the README. It adds the command to install the ```ros-foxy-cv-bridge``` package. Without it, colcon runs into errors.


doronhi:
I think something went wrong with your installation.
cv_bridge is specified as a dependency in package.xml and should have been installed by the `rosdep install -i --from-path src --rosdistro $ROS_DISTRO -y` command.
I think it's best if you opened an issue and we'll try to figure it out there. Please link this PR and the issue.

I am leaving this PR open for now, until the matter is resolved.

Issue #2001, opened 2021-07-27
ROS2 same frame_ids for multiple cameras


Marenix:
Hey, I have a question regarding frame_ids in ROS2. I might've missed or misinterpreted something, so I'd appreciate any explanation/help. 

I've been testing both ROS and ROS2 wrappers and I've noticed that they have different naming convention for frame_ids. Let's take color_frame_id for example. If we connect two cameras and start rs_multiple_devices.launch, one camera will have color_frame_id=camera1_color_frame while the other will have color_frame_id=camera2_color_frame. On the other hand, using ROS2 to launch rs_multi_camera_launch.py, both camera's will have color_frame_id=camera_color_frame.

My question is why? Is there some reason why this was changed in ROS2?
Thanks for the help!

brennand:
Looking at the code, the `base_realsense_node.cpp` ignores the "camera_name" param.  If you change the camera name, the frame_id in the header doesn't change.

I think we need to patch this to support multi Realsense's.

Marenix:
Yea, I agree. I looked through it too, and saw that frame ID names are made in the following way: camera_STREAM_NAME()_frame which actually makes same named streams of different camera all "blend" with that one name. In ROS wrapper this is solved in launch file/nodelet.launch.xml file where each parameter name is defined as namespace+(rest of the name). This is "convenient" in ROS because in nodelet.launch.xml you can put the prefix of namespace as the parameter that is inputted as namespace for each node. There isn't such an option in ROS2, at least not that I've found, and so, in ROS2, you would have to change the name of each frame by hand. 

I think it would be a good idea to change `base_realsense_node.cpp`, to change the definitions of `FRAME_ID()` `OPTICAL_FRAME_ID` and `ALIGNED_DEPTH_TO_FRAME_ID` so that they don't start with "camera_" but rather the `namespace` of the camera

brennand:
I got it working.  Not really a nice solution.  I added all the frame_id's to rs_camera.launch.py  And then my main launch file passed my new frame ID's as params.:

    {"name": "base_frame_id", "default": "camera_link", "description": ""},
    {"name": "depth_frame_id", "default": "camera_depth_frame", "description": ""},
    {"name": "color_frame_id", "default": "camera_color_frame", "description": ""},

eta

[...]

PR #2128, opened 2021-10-21
Relative naming of coordinate frames

Marenix:
Please see [Issue #2001](https://github.com/IntelRealSense/realsense-ros/issues/2001) for more information.

Basically, this changes naming of frame ID's in base_realsense_node.cpp, so that instead of generic "camera" camera name is used. Additionally, rs_multi_camera launch file starts one more node which publishes transformation information from camera1 to camera2. As cameras no longer have the same coordinate frame names, this transformation is necessary to get streams from both cameras. In final applications, this transformation would contain real world transformation of position of one camera to the other, but in the examples, all the values are set to 0.

Naming coordinate frames this way would make it easier for users to connect multiple cameras and not have to worry about renaming all frame ID's in either launch or config file. Finally, applications which use multiple cameras need to have a set transformation, as this is used in processing information (images), so that position of each pointcloud is set correctly.

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -199,6 +199,7 @@ namespace realsense2_camera
                 double          m_time_ns;
         };
 
+        std::string getNamespaceStr();
         void getParameters();
         void setupDevice();
         void setupErrorCallback();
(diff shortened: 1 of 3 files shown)


doronhi:
Thanks.
I think this solution is indeed better than manipulating the launch files as it doesn't require updating the launch files if the streams ever changed.
I also left the static_transform_publisher node as a good example, although it is not strictly part of realsense2_camera node. It may be removed in the future.

Issue #2139, opened 2021-10-26
No message passing through rosout from rs_launch.py


JamesChooWK:
Hi, I recently discovered that when I launched the `rs_launch.py` file. No logs message are passing into the `/rosout `topic, however, I can see all the messages showing on the terminal like what I show below.
![Screenshot from 2021-10-26 16-38-04](https://user-images.githubusercontent.com/46285293/138901943-ab67f6ec-a4e8-41e3-b131-c1354ba55555.png)

This is wired for me because normally when the node is set in the debug mode, all the logs messages will pass through `/rosout` and I can use `rqt_console` to view the log message.

My question is, is there any method or way that I can make the log message passes through `/rosout`? so that I can bag them using rosbag.

OS:Ubuntu 20.04
ROS: Foxy

Thanks!

MartyG-RealSense:
Hi @JamesChooWK  There was a recent Foxy case at https://github.com/ros-visualization/rqt_console/issues/35 where messages were not getting through to RQT.  The ROS user in that case concluded that it was related to QOS, and they  found that creating a custom QOS enabled RQT to capture all missing messages.

They experienced the problem during high message loads, so it may be worth testing with a lower resolution such as 640x480 instead of 1280x720 to see whether this enables the messages to be captured in rosout.

JamesChooWK:
Hi, here are the updates from my testing. 

1. I have played with the QOS setting at the rqt_console and also on the launch file of Realsense (parameter inside the rs_launch.py) but still get no luck. 
2. I have also tried to lower down the frame rate and frame size, but also still see nothing on rqt_console.

I used the launch file (rs_launch.py) in both the trail. I can see all the topics via the terminal using `ros2 topic list` and able to see the images in the topic message using rqt. However, the message from the logger (ROS_INFO_STREAM etc.) will still only appears on my terminal and not on the rqt console.

MartyG-RealSense:
Are there any messages on the topic **/rosout_agg** that publishes an aggregated feed of the logging messages?

http://wiki.ros.org/rosout#A.2Frosout_agg

JamesChooWK:
Correct if I'm wrong, but I don't think ros2 has the /rosout_gg.

MartyG-RealSense:
You can check whether /rosout_agg is available on ROS2 by listing the topics with **ros2 topic list**

JamesChooWK:
[...]

PR #2154, opened 2021-11-11
Fix no ros logging messages in rosout topic.

JamesChooWK:
This is an attempt to fix issues tickets #2139, where the ros2 logging message is not shown on the /rosout topic.


Bart-van-Ingen:
@doronhi the CI is failing here, because it is unable to install the realsense SDK ...

PR #2183, opened 2021-12-01
Remove empty warning on realsense startup

hello-binit:
Using the the command `roslaunch realsense2_camera rs_camera.launch` to initialize a D435i, a empty ROS warning is printed in the console:

```
[ INFO] [1638390984.478820054]: SELECTED BASE:Depth, 0
[ INFO] [1638390984.513772271]: RealSense Node Is Up!
[ WARN] [1638390984.556584224]: 
```

After applying this patch, the warning is gone:
```
[ INFO] [1638391138.673114422]: SELECTED BASE:Depth, 0
[ INFO] [1638391138.723653877]: RealSense Node Is Up!
```

Let me know if you need me to change anything.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1801,7 +1801,7 @@ void BaseRealSenseNode::multiple_message_callback(rs2::frame frame, imu_sync_met
 
 bool BaseRealSenseNode::setBaseTime(double frame_time, rs2_timestamp_domain time_domain)
 {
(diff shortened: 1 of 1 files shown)


doronhi:
The problem with your suggestion is that now if metadata is not available, you'll get a warning for every frame.
I prefer the once-printed empty message.

doronhi:
Hi @hello-binit ,
I agree that the empty message is not the prettiest thing I encountered but I feel it is better than a complicated-code solution.
I thank you for your suggestion.

PR #2184, opened 2021-12-02
Add namespace without affecting frame names

tonynajjar:
The current implementation does not allow to set a namespace to the realsense node without changing the frame names, causing weird names when propagating a namespace using PushRosNamespace e.g agv4realsense_left_optical_color_frame .This PR proposes to use the node name instead of the namespace to prepend to the frame names.

This is untested, I can test if the idea is accepted

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -96,16 +96,14 @@ def set_configurable_parameters(parameters):
     return dict([(param['name'], LaunchConfiguration(param['name'])) for param in parameters])
 
 def generate_launch_description():
-    log_level = 'info'
     if (os.getenv('ROS_DISTRO') == "dashing") or (os.getenv('ROS_DISTRO') == "eloquent"):
         return LaunchDescription(declare_configurable_parameters(configurable_parameters) + [
             # Realsense
             launch_ros.actions.Node(
                 condition=IfCondition(PythonExpression([LaunchConfiguration('config_file'), " == ''"])),
                 package='realsense2_camera',
-                node_namespace=LaunchConfiguration("camera_name"),
-                node_name=LaunchConfiguration("camera_name"),
-                node_executable='realsense2_camera_node',
+                node=LaunchConfiguration("camera_name"),
+                executable='realsense2_camera_node',
                 prefix=['stdbuf -o L'],
                 parameters=[set_configurable_parameters(configurable_parameters)
                             ],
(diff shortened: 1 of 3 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -177,11 +177,9 @@ size_t SyncedImuPublisher::getNumSubscribers()
...
 }
 
-std::string BaseRealSenseNode::getNamespaceStr()
+std::string BaseRealSenseNode::getNodeNameStr()
 {
-    std::string ns(_node.get_namespace());
-    ns.erase (std::remove(ns.begin(), ns.end(), '/'), ns.end());
-    return ns;
+    return _node.get_name();
 }
```
tonynajjar, line 183: probably doesn't deserve a function anymore

realsense2_camera/launch/rs_launch.py
```diff
@@ -96,16 +96,14 @@ def set_configurable_parameters(parameters):
...
-    log_level = 'info'
     if (os.getenv('ROS_DISTRO') == "dashing") or (os.getenv('ROS_DISTRO') == "eloquent"):
         return LaunchDescription(declare_configurable_parameters(configurable_parameters) + [
             # Realsense
             launch_ros.actions.Node(
                 condition=IfCondition(PythonExpression([LaunchConfiguration('config_file'), " == ''"])),
                 package='realsense2_camera',
-                node_namespace=LaunchConfiguration("camera_name"),
-                node_name=LaunchConfiguration("camera_name"),
-                node_executable='realsense2_camera_node',
```
tonynajjar, line 108: node_name and node_executable are deprecated

doronhi:
Hi @tonynajjar ,
I agree with your idea of using the node's name instead of the namespace. In fact, this is already the way frame names are being set in the "ros2-beta" branch which is intended to be the next version of the realsense-ros wrapper.
Since we already have a working new version with some more interface changes (which you are more than welcome to try out), I will not add an interface-changing modification into the previous "ros2" version.

Regarding the deprecated "node_name" and "node_executable" arguments, they are being used only in the section of the launch file that is meant to support the old "dashing" and "eloquent" releases, where these names were still used.

tonynajjar:
Got it. Would you recommend switching to ROS2-beta or is it too soon? Do you document somewhere the improvements done in ros2 beta?

doronhi:
Version 4.0.1 was released today with many issues fixed.
As far as I and some other teams that I know of have tested, the beta release is at least as stable as the previous ros2 release. 
The improvements and API changes are covered best in the release notes for [version 4.0.0](https://github.com/IntelRealSense/realsense-ros/releases/tag/v4.0.0-beta) and in the [CHANGELOG.srt file for that version](https://github.com/IntelRealSense/realsense-ros/blob/ros2-beta/realsense2_camera/CHANGELOG.rst#440-2021-11-17).

tonynajjar:
I looked at the improvements and they fix many of the bugs we are having, we might consider using it and giving feedback. Thanks

PR #2205, opened 2021-12-30
rename camera_link frame_id

doronhi:
When running rs_multi_camera_launch.py, both nodes publish a "camera_link" static frame, ignoring the names "camera1", "camera2"
This PR fixes the behavior of the "camera_link" frame_id to match the one shared by the rest of the frame_ids - to use the camera_name as a prefix.
In short: rename camera_link frame_id to <camera_name>_link


doronhi:
Replaces the need for #2204

PR #2246, opened 2022-02-07
Allow the catkin package to be included in other projects

scott-robotics:
I have modified the include structure to be consistent with the way other ROS/catkin packages do it. This way, the `realsense2_camera` package can be reused in other projects.


doronhi:
Thanks for the cleanup. It also looks much better this way.

PR #2252, opened 2022-02-12
ROS2-beta: Add frequency monitoring to /diagnostics topic.

doronhi:
Add frequency of enabled streams to the /diagnostics topic.


realsense2_camera/src/ros_sensor.cpp
```diff
@@ -201,6 +216,8 @@ void RosSensor::stop()
     if (get_active_streams().size() == 0)
         return;
     ROS_INFO_STREAM("Stop Sensor: " << get_info(RS2_CAMERA_INFO_NAME));
+    _frequency_diagnistics.clear();
```
remibettan: typo - diagnostics instead of diagnistics

doronhi: Fixed, Thanks.

PR #2255, opened 2022-02-17
ROS2 - Diagnostics added

remibettan:


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -1872,11 +1883,23 @@ void BaseRealSenseNode::pose_callback(rs2::frame frame)
...
+        _frequency_diagnostics.at(sip).Tick();
+}
+
 void BaseRealSenseNode::frame_callback(rs2::frame frame)
 {
     _synced_imu_publisher->Pause();
 
     try{
+        // calling the Tick api for the diagnostics topic
+        frequencyDiagnosticsTick(frame);
```
doronhi, line 1901: At this point "frame" can be either a frame or a frameset. Calling "frequencyDiagnosticsTick" with a frameset causes a Tick only on 1 of the frames in the frameset.
This can be observed if running with `ros2 launch realsense2_camera rs_launch.py diagnostics_period:=1. enable_sync:=true`

doronhi (changes requested):

realsense2_camera/src/base_realsense_node.cpp
doronhi: Also, Please rename the PR to "ROS2-Diagnostics added" - it will help to distinguish between the currently 3 active branches.
Also, please provide a description - What this PR does. What problem does it solve or what additional capabilities it provides. If the PR is a migration from another branch, like this one, you can note that and provided a link to the reference PR.

Nir-Az:
@remibettan , @maloel  .
Too bad for this work  to go to waste.
We need to speak about who is merging PR's..
This PR is not going to merge itself 😉

Nir-Az:
Feature Dropped for old ROS2 branch

PR #2274, opened 2022-03-07
ROS2-beta - Fixing Galactic "deprecated-declarations" compilation warning

SamerKhshiboun:
Duration(rcl_duration_value_t) is deprecated in favor of static Duration::from_nanoseconds(rcl_duration_value_t) starting from v5.0.1

See: [5.1.0/rclcpp/include/rclcpp/duration.hpp](https://github.com/ros2-gbp/rclcpp-release/blob/upstream/5.1.0/rclcpp/include/rclcpp/duration.hpp)

This effects all versions from ROS2 Galactic and above, since older ROS2 distributions are using  rclcpp versions lower than 5.1.0

Tracked by [LRS-390]

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -643,17 +643,17 @@ rclcpp::Time BaseRealSenseNode::frameSystemTimeSec(rs2::frame frame)
     if (frame.get_frame_timestamp_domain() == RS2_TIMESTAMP_DOMAIN_HARDWARE_CLOCK)
     {
         double elapsed_camera_ns = (/*ms*/ frame.get_timestamp() - /*ms*/ _camera_time_base) * 1e6;
-	
-		/*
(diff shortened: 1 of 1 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -643,7 +643,20 @@ rclcpp::Time BaseRealSenseNode::frameSystemTimeSec(rs2::frame frame)
     if (frame.get_frame_timestamp_domain() == RS2_TIMESTAMP_DOMAIN_HARDWARE_CLOCK)
     {
         double elapsed_camera_ns = (/*ms*/ frame.get_timestamp() - /*ms*/ _camera_time_base) * 1e6;
-        return rclcpp::Time(_ros_time_base + rclcpp::Duration(elapsed_camera_ns));
+	
+		/*
```
maloel: Looks like there are tab characters here instead of spaces

maloel (approved):

PR #2275, opened 2022-03-07
ROS2-beta - Update README.md and package.xml

SamerKhshiboun:
- Update maintainer and authors in realsense2_camera/package.xml
- Fix/Reformat installation steps in README.md

Tracker by [LRS-392] and [LRS-396]


README.md
```diff
@@ -14,62 +14,61 @@ This version supports ROS2 Dashing, Eloquent, Foxy, Galactic and Rolling.
...
-  colcon build
-  ```
+   colcon build
+   ```
 
-  ### Step 6: Source (on each new terminal):
-  ```bash
-  . install/local_setup.bash
-  ```
+### Step 6: Source (on each new terminal):
```
maloel: What does this mean? `on each new terminal`

SamerKhshiboun: These source commands set env variables to run ROS2 with the chosen distro.
You can add these source commands to your ~/.bashrc and then you will not need to do it on each terminal. 
**It's up to the user.**
I chose this way because I have multiple ROS2 distros on my machine, so I need to config each terminal with the right ROS2 distro setup

maloel (approved):

PR #2293, opened 2022-03-20
Add pre-release testings to ci

Nir-Az:
Add ROS build farm local build to CI for galactic/foxy (only supported distro's GH actions)


maloel:
Maybe just add an explanation to the .yml so readers can understand what it is we're trying to add?

maloel (approved):

PR #2324, opened 2022-04-19
ROS2-beta  - Fix align_depth enable/disable

SamerKhshiboun:
Tracked on [LRS-407] and [LRS-406]

### DIFF
--- a/realsense2_camera/include/named_filter.h
+++ b/realsense2_camera/include/named_filter.h
@@ -57,4 +57,11 @@ namespace realsense2_camera
             rclcpp::Publisher<sensor_msgs::msg::PointCloud2>::SharedPtr _pointcloud_publisher;
             std::string _pointcloud_qos;
     };
+
+    class AlignDepthFilter : public NamedFilter
+    {
+        public:
+            AlignDepthFilter(std::shared_ptr<rs2::filter> filter, std::function<void(const rclcpp::Parameter&)> update_sensor_func,
+                std::shared_ptr<Parameters> parameters, rclcpp::Logger logger, bool is_enabled = false);
+    };
 }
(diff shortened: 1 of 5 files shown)


SamerKhshiboun:
Tracked by LRS-407

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -157,8 +158,11 @@ void BaseRealSenseNode::setupFilters()
     _filters.push_back(std::make_shared<NamedFilter>(std::make_shared<rs2::temporal_filter>(), _parameters, _logger));
     _filters.push_back(std::make_shared<NamedFilter>(std::make_shared<rs2::hole_filling_filter>(), _parameters, _logger));
     _filters.push_back(std::make_shared<NamedFilter>(std::make_shared<rs2::disparity_transform>(false), _parameters, _logger));
-    _align_depth_filter = std::make_shared<NamedFilter>(std::make_shared<rs2::align>(RS2_STREAM_COLOR), _parameters, _logger);
+
+    std::function<void(const rclcpp::Parameter&)> update_sensor_func = [this](const rclcpp::Parameter&){_is_align_depth_changed = true; _cv_mpc.notify_one();};
```
Nir-Az, line 162: I know it wasn't implemented like that on `is_profile_changed` but I think it was missed and the behavior can be different than expected sometimes.

When using conditional variables between threads the thread that change the shared variable has to lock (mutex) the variable once he is changing it with the same lock that used to lock the CV.

You can read about it here:
1. https://en.cppreference.com/w/cpp/thread/condition_variable
2. https://www.modernescpp.com/index.php/c-core-guidelines-be-aware-of-the-traps-of-condition-variables

realsense2_camera/include/named_filter.h
```diff
@@ -57,4 +57,11 @@ namespace realsense2_camera
             rclcpp::Publisher<sensor_msgs::msg::PointCloud2>::SharedPtr _pointcloud_publisher;
             std::string _pointcloud_qos;
     };
+
```
Nir-Az, line 61: Please add comment about why do we need a special handling for the align filter

SamerKhshiboun:
Tracked on [LRS-407] and [LRS-406]

SamerKhshiboun:
Closing this PR and opening two PRs since we are fixing two different issues
LRS-406: PR 2336
LRS-407: PR 2332

PR #2325, opened 2022-04-20
[Bugfix] Commented materials urdf and defined each material aluminum …

Giggus:
Fix  "Error: material 'aluminum' is not unique" for multiple type camera setup


SamerKhshiboun:
Hi @Giggus ,

As I see there are several PRs regarding this issue:
- #2325 (This PR)
- #1637 
- #2674 

I think that all of them solve the problem, but as I see #1637 is the one with minimal changes.
Do you agree with me ?
If so, I can close other PRs and approve that PR

Thanks,
Samer

SamerKhshiboun:
Fixed in #1637 for ROS1 and in #2689 for ROS2

PR #2332, opened 2022-04-27
Fix align_depth enable/disable

SamerKhshiboun:
Tracked on [LRS-407]

### DIFF
--- a/realsense2_camera/src/named_filter.cpp
+++ b/realsense2_camera/src/named_filter.cpp
@@ -263,3 +263,14 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
             _pointcloud_publisher->publish(_msg_pointcloud);
     }
 }
+
+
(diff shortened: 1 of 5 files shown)


Nir-Az (approved):
LGTM

realsense2_camera/src/named_filter.cpp
```diff
@@ -263,3 +263,14 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
             _pointcloud_publisher->publish(_msg_pointcloud);
     }
 }
+
+
+AlignDepthFilter::AlignDepthFilter(std::shared_ptr<rs2::filter> filter,
```
maloel, line 268: Shouldn't this be in its own .cpp? `named_filter.cpp` should be only for the `NamedFilter` class, no?

maloel: I'll ignore for now :)

maloel (approved):

PR #2404, opened 2022-07-08
fix ros bag playback

gwen2018:
Changes:
Fix ros2_beta can not playback from ros bag

Tracked on: LRS-445


SamerKhshiboun:
Fixed in #2409 
This PR can be closed.

PR #2406, opened 2022-07-11
Support ROS2 Humble

SamerKhshiboun:
Tracked by [LRS-449]


README.md
```diff
@@ -28,18 +28,20 @@ LibRealSense supported version: v2.50.0 (see [realsense2_camera release notes](h
...
    ### Step 1: Install the ROS2 distribution
- - #### Ubuntu 18.04 : 
-   - [ROS2 Dashing](https://docs.ros.org/en/dashing/Installation/Ubuntu-Install-Debians.html)
-   - [ROS2 Eloquent](https://docs.ros.org/en/eloquent/Installation/Linux-Install-Debians.html)
+ - #### Ubuntu 22.04:
+   - [ROS2 Humble](https://docs.ros.org/en/humble/Installation/Ubuntu-Install-Debians.html)
  - #### Ubuntu 20.04: 
    - [ROS2 Foxy](https://docs.ros.org/en/foxy/Installation/Ubuntu-Install-Debians.html)
    - [ROS2 Galactic](https://docs.ros.org/en/galactic/Installation/Ubuntu-Install-Debians.html)
    - [ROS2 Rolling](https://docs.ros.org/en/rolling/Installation/Ubuntu-Install-Debians.html)
```
Nir-Az: I suggest moving Rolling out of the Ubuntu distribution title, Rolling should be above as the ROS2 up to date code (development)

realsense2_camera/src/dynamic_params.cpp
```diff
@@ -103,7 +103,7 @@ namespace realsense2_camera
         try
         {
             ROS_DEBUG_STREAM("setParam::Setting parameter: " << param_name);
-#if defined(GALACTIC) || defined(ROLLING)
```
Nir-Az, line 106: What do you think about reversing this condition? sounds like a good idea?

README.md
SamerKhshiboun: I will remove it at all from here, because we are not supporting Rolling distros really.
What do you think ?

realsense2_camera/src/dynamic_params.cpp
SamerKhshiboun: done

Nir-Az (approved):
Great

Nir-Az:
Verified as working!
This image is from RVIZ2
image

maloel (approved):

PR #2408, opened 2022-07-12
Publish real frame rate of realsense camera node topics/publishers

SamerKhshiboun:
- Catch backend exceptions and don't tick frequency diagnostics if so
- Tick the frequency diagnostics only if topic publishes the frame

All changes in base_realsense_node are removing the try/catch and moving them outside to the ros_sensor file

Tracked on [LRS-434] and [LRS-448]


SamerKhshiboun:
all changes in base_realsense_node are removing the try/catch and moving them outside to the ros_sensor file

Nir-Az (approved):
LGTM

PR #2409, opened 2022-07-13
fix rosbag device loading by preventing set_option to HDR/Gain/Exposure

SamerKhshiboun:
Tracked by [LRS-445]

If rosbag file was loaded, a virtual device cannot set options of HDR/Exposure/Gain
Prevent that by asking if we are on a rosbag file

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -97,7 +97,8 @@ namespace realsense2_camera
         BaseRealSenseNode(rclcpp::Node& node,
                           rs2::device dev,
                           std::shared_ptr<Parameters> parameters,
-                          bool use_intra_process = false);
+                          bool use_intra_process = false,
+                          bool is_rosbag_file = false);
         ~BaseRealSenseNode();
         void publishTopics();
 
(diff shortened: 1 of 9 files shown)


realsense2_camera/include/base_realsense_node.h
```diff
@@ -97,7 +97,8 @@ namespace realsense2_camera
         BaseRealSenseNode(rclcpp::Node& node,
                           rs2::device dev,
                           std::shared_ptr<Parameters> parameters,
-                          bool use_intra_process = false);
+                          bool use_intra_process = false,
+                          bool is_rosbag_file = false);
```
Nir-Az: Since you are getting the rs2::device as well, try to drop the new parameter (all over the code) and use this API, see if it's possible:
```
if (dev.is<playback>())
```

Nir-Az:
Have you seen this PR #2404 ?

realsense2_camera/include/base_realsense_node.h
SamerKhshiboun: Thanks. 
It make sense and make the code easier.
I saw the PR you talked about, but still I wanted to do this without asking the primary node about rosbag_filename parameter value, since I don't thinks this is the best way to do this.

Nir-Az: Great, Lets also check if this API works too:
```
sensor.is<playback_sensor>()
```

If this works we can drop more changes.
Worth checking

Nir-Az: Not sure it's possible, I don't see `playback_sensor` in the API

maloel (approved):

PR #2436, opened 2022-08-02
Publish static tf for IMU frames

Gilaadb:
{camera}_imu_frame and {camera}_imu_optical_frame were not published.
The tf is the same as the Gyro's and Accelerometer's. Now publishing
IMU frames tf whenever publishing gyro's tf.


SamerKhshiboun (approved):
LGTM

PR #2460, opened 2022-08-26
Expose angular_velocity_cov arg

hello-binit:
The README documents `angular_velocity_cov` as an argument for the `rs_camera.launch` file, but it's not available. This PR would make it available.


SamerKhshiboun:
Hi @hello-binit,
Thanks for contribution, but we are not maintaining ros1 branches anymore.
I'm closing the PR, but users can still take these fixes manually to their repos in the future.

PR #2468, opened 2022-09-04
force cmake version to 3.23.0

SamerKhshiboun:
Tracked by LRS-504


Nir-Az (approved):
Great until it will be fixed on their side.

Issue #2432, opened 2022-07-30
Possibly deceptive extrinsics topic name?


john-maidbot:
Realsense-ros version: ros-galactic-realsense2-camera/focal,focal,now 4.0.4-1focal.20220430.140243 amd64
librealsense version: ros-galactic-librealsense2/focal,focal,now 2.50.0-2focal.20220222.210235 amd64
Camera: d435
OS: Ubuntu 20.04

Please correct me if I am wrong, but it appears that the .../extrinsics/depth_to_color is actually the extrinsics going from the color camera to the depth camera. If that is the case, I think the topic should be .../extrinsics/color_to_depth to clarify this point OR there should be a note about this in the readme.

Here is where you lookup the extrinsic information:
https://github.com/IntelRealSense/realsense-ros/blob/442d6d9a19149cb9cb40ffb351593c3fa8b9eb9f/realsense2_camera/src/base_realsense_node.cpp#L2260

```
ex = getAProfile(stream).get_extrinsics_to(base_profile);
```

if,
base_profile == depth profile
getAProfile(stream) == color profile
then ex will be the extrinsic from the color to the depth camera, right?

I think this is the case, since when I get the extrinsic from realsense-ros, I get:

```
ros2 topic echo /d435/extrinsics/depth_to_color
rotation:
- 0.9999857544898987
- 0.0021478114649653435
- -0.0048892986960709095
- -0.0021304551046341658
- 0.9999914169311523
- 0.003552317386493087
- 0.004896886646747589
- -0.0035418502520769835
- 0.9999817609786987
translation:
- -0.014618603512644768
- -4.5008506276644766e-05
- -0.0005783233791589737
---

```

Which I have confirmed is in fact the extrinsic from color to depth, not "depth_to_color":
```
rs-enumerate-devices -c
Extrinsic from "Color"	  To	  "Depth" :
 Rotation Matrix:
   0.999986        -0.00213046       0.00489689    
   0.00214781       0.999991        -0.00354185    
  -0.0048893        0.00355232       0.999982      

 Translation Vector: -0.0146186035126448  -4.50085062766448e-05  -0.000578323379158974  
```

Thanks for your time! (-:

MartyG-RealSense:
Hi @john-maidbot  I will enquire to the RealSense ROS developer team about your question.  Thanks for your patience!

SamerKhshiboun:
Hi @john-maidbot,
[...]

PR #2477, opened 2022-09-13
fix extrinsics calculation

SamerKhshiboun:
Tracked by LRS-476

Issue #2432

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -850,7 +850,7 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
     rs2_extrinsics ex;
     try
     {
-        ex = profile.get_extrinsics_to(base_profile);
+        ex = base_profile.get_extrinsics_to(profile);
     }
     catch (std::exception& e)
     {
--- a/realsense2_camera/src/t265_realsense_node.cpp
+++ b/realsense2_camera/src/t265_realsense_node.cpp
@@ -89,7 +89,7 @@ void T265RealsenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
     rs2_extrinsics ex;
     try
     {
-        ex = profile.get_extrinsics_to(base_profile);
+        ex = base_profile.get_extrinsics_to(profile);
     }
     catch (std::exception& e)
     {


remibettan (approved):
LGTM

Nir-Az (approved):
LGTM

schmidtp1:
i don't agree with this fix. the way "from - to" is used is ambiguous. in librealsense "extrinsics from A to B" refer to the transformation of a coordinates **from/ expressed in frame A _to_ coordinates expressed in frame B** (which is the same as coordinate frame **A expressed _with respect to_ frame B**). ROS, on the other hand, seems to refer with "to" rather to the kinematic tree (from "parent frame" to "child frame") which is exactly the opposite direction.

i think the documentation here is a bit misleading: http://docs.ros.org/en/noetic/api/geometry_msgs/html/msg/TransformStamped.html but you can easily test it by publishing:
$ ros2 run tf2_ros static_transform_publisher 0 0 1 0 0 0 1 frame_id child_frame_id
$ ros2 topic echo /tf_static
transforms:
- header:
    stamp:
      sec: 1665789714
      nanosec: 273191209
    frame_id: frame_id
  child_frame_id: child_frame_id
  transform:
    translation:
      x: 0.0
      y: 0.0
      z: 1.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0

and possibly inspecting in rviz.
Screen Shot 2022-10-14 at 4 36 34 PM
to transform from "child frame" to "frame" you would add 1. from "frame" to "child frame" subtract 1. (as a simple example.)
please let me know if this makes sense, @SamerKhshiboun, @Nir-Az.

Nir-Az:
> i don't agree with this fix. the way "from - to" is used is ambiguous. in librealsense "extrinsics from A to B" refer to the transformation of a coordinates **from/ expressed in frame A _to_ coordinates expressed in frame B** (which is the same as coordinate frame **A expressed _with respect to_ frame B**). ROS, on the other hand, seems to refer with "to" rather to the kinematic tree (from "parent frame" to "child frame") which is exactly the opposite direction.
> 
> i think the documentation here is a bit misleading: http://docs.ros.org/en/noetic/api/geometry_msgs/html/msg/TransformStamped.html but you can easily test it by publishing: $ ros2 run tf2_ros static_transform_publisher 0 0 1 0 0 0 1 frame_id child_frame_id $ ros2 topic echo /tf_static transforms:
> 
> * header:
>   stamp:
>   sec: 1665789714
>   nanosec: 273191209
>   frame_id: frame_id
>   child_frame_id: child_frame_id
>   transform:
>   translation:
>   x: 0.0
>   y: 0.0
>   z: 1.0
>   rotation:
>   x: 0.0
>   y: 0.0
>   z: 0.0
>   w: 1.0
> 
> and possibly inspecting in rviz. Screen Shot 2022-10-14 at 4 36 34 PM to transform from "child frame" to "frame" you would add 1. from "frame" to "child frame" subtract 1. (as a simple example.) please let me know if this makes sense, @SamerKhshiboun, @Nir-Az.

Thanks for the feedback @schmidtp1 .
We will take a look at your perspective and investigate this issue further.
Thanks.

SamerKhshiboun:
Hi @schmidtp1 and thanks for you informative comment

I think you are right.
The documentation of [TransformStamped](http://docs.ros.org/en/noetic/api/geometry_msgs/html/msg/TransformStamped.html) is "This expresses a transform from coordinate frame header.frame_id to the coordinate frame child_frame_id"

And as I see in one of our static TFs message:

`header:
    stamp:
      sec: 1680527160
      nanosec: 457847686
    frame_id: camera_depth_frame
  child_frame_id: camera_depth_optical_frame
  transform:
    translation:
      x: 0.0
      y: -0.0
      z: -0.0
    rotation:
      x: -0.5
      y: 0.4999999999999999
      z: -0.5
      w: 0.5000000000000001
`

while frame_id (parent) let it be the **ROS2 depth coordinates**
and child_frame_id (child) let it be the **OPTICAL depth coordinates**

image


so if we want to move from parent to child, which means from ROS2 coordinates to OPTICAL coordinates in this example, 
we should use this rotation (+0.5, -0.5, +0.5) which is the opposite values of what we get today in our static TF published message.

I think that now we calculate the extrinsic in a right way (after the PR is merged) and publish them correctly to the corresponding topics (e.g., /camera/extrinsics/depth_to_color).
But as I see, we need to manipulate (turn around) these values before we publish them to the static TF topic, and that can be done easily in our code.

Adding also @Gilaadb, @benlev since they are using TFs and Extrinsic topic so they may can advice more.

Issue #2339, opened 2022-05-05
How to have /camera/rgb_camera/parameter_descriptions topic?


miku54:
I use a D455 camera, which is connected to my device via USB3.0,my system is foxy,and I am using the ROS2-beta branch.
ros2 launch realsense2_camera rs_launch.py

Tried to start like this, there is no /camera/rgb_camera/parameter_descriptions topic
：ros2 run realsense2_camera realsense2_camera_node

miku54:
When I start the camera with this command：
ros2 launch realsense2_camera rs_launch.py​​ align_depth:=true pointcloud.enable:=true
Still no topic of pointcloud2；
https://github.com/IntelRealSense/realsense-ros/issues/2295
This thread has a similar situation to mine, but I can't use this workaround to solve my problem

miku54:
![lADPJxuMQwRvJGHNC9DND8A_4032_3024](https://user-images.githubusercontent.com/81612531/166915479-8c75208e-01b7-45c3-8f35-f46f461ac2a9.jpg)

MartyG-RealSense:
Hi @miku54  The **parameter_descriptions** topic was listed as supported in the previous ROS2 wrapper (up to version 3.2.3) but is not included in the topics list of the ros2_beta (4.x versions) wrapper documentation.  Instead it adds a topic called **parameter_events**

https://github.com/IntelRealSense/realsense-ros/tree/ros2-beta#published-topics

Are you able to achieve a point cloud with the launch instruction below?

`ros2 launch realsense2_camera rs_launch.py depth_module.profile:=1280x720x30 pointcloud.enable:=true`

miku54:
Hi, @MartyG-RealSense <https://github.com/MartyG-RealSense> thanks for your
reply!
I used to download the realsense package of ros2 using binary instructions,
and now I use ros2_beta to start the camera;
ros2 launch realsense2_camera rs_launch.py
depth_module.profile:=1280x720x30 pointcloud.enable:=true
It is not possible to implement point clouds, but I found that there is a
launch (
https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/launch/demo_pointcloud.launch)
file in the ROS2 branch that can open point clouds and get topics data, but
it only runs for a few seconds after opening the file, and then the image
freezes (it looks more like the node died) but the terminal doesn't report
any errors, which is too weird

MartyG-RealSense ***@***.***> 于2022年5月6日周五 16:29写道：

[...]

PR #2487, opened 2022-09-21
fix ordered_pc arg prefix

SamerKhshiboun:
Fixing Pointcloud ordered_pc issue #2339 

Tracked on [LRS-486]


Nir-Az (approved):
Good catch!

PR #2536, opened 2022-11-08
Update stale comment to reflect new branch name

jaiveersinghNV:
Seems like there's a stale comment referring to the old, `ros2-beta` branch name, when it should instead reference the new, `ros2-development` branch.


Nir-Az (approved):
Great, nice catch!

PR #2548, opened 2022-11-16
add rolling distro build to github actions

Nir-Az:
Tracked on [LRS-556]

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,8 +23,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


Nir-Az:
@SamerKhshiboun See this notes:
1. The 3.23 fixed version cannot be removes as some of the distros are still failing.
2. I tried building all distros with librealsense2 dependency (except rolling) but eloquent/dashing could not build since we didn't release it and the package.xml librealsense2 version is higher than the latest release, so I left the old logic

SamerKhshiboun (approved):

PR #2572, opened 2022-12-07
Properly read camera config files in rs_launch.py

Gilaadb:
Modified rs_launch.py to properly read parameters from yaml config file. It did not read them before.
Now we can set camera paramters that are not listen in rs_launch.py "configurable_parameters" list by specifying them in a yaml config file.

Example for such parameters could be (among many others): depth_module.enable_auto_exposure
depth_module.inter_cam_sync_mode

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -3,12 +3,11 @@
 
 """Launch realsense2_camera node."""
 import os
+import yaml
 from launch import LaunchDescription
-from ament_index_python.packages import get_package_share_directory
 import launch_ros.actions
-from launch.actions import DeclareLaunchArgument
-from launch.substitutions import LaunchConfiguration, PythonExpression
-from launch.conditions import IfCondition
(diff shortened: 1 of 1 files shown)


Nir-Az:
Thanks for the PR Gilad, The changes looks very good.
Can you please elaborate how this was tested?
With config file + without? with dynamic changes on CLI?
ROS2 distro?

Gilaadb:
Hi Nir. I tested it on ROS Foxy, with and without a config file along with parameters that were already supported by the launch file (such as streams selection, resolution, IMU enablement etc...). Meaning the parameters from the yaml and from the launch arguments do merged to one list of parameters for the node.
It shouldn't change what is possible or not possible to change with dynamic changes on CLI - I tested it to some extent by tweaking some of the parameters via 'rqt' after a camera node start.

The main advantage you gain from applying this change, is to be able to set camera node parameters that you will see if you will do a "ros2 param dump /camera/camera" but are not included in the subset of parameters that are listed in the "configurable_parameters" variable in the rs_launch.py file.

Nir-Az (approved):
Works like a charm :)

PR #2579, opened 2022-12-14
Adds the use of type adaptation to the image publisher

Voldivh:
## Summary
This PR, as the tittle states, adds the use of type adaptation for the image publisher. Basically, it takes the [type adapter](https://github.com/ros-perception/vision_opencv/blob/rolling/cv_bridge/include/cv_bridge/cv_mat_sensor_msgs_image_type_adapter.hpp) from the `cv_bridge` package and uses it on the publisher in order to improve efficiency of the communication of the image. This modification is only available when the realsense_camera_node is being used with intra-process communication enabled since that is the only way we can take advantage of the use of [type adaptation](https://ros.org/reps/rep-2007.html). 

The modification made on this PR, does not break any current use of the driver since the type adaptation on the publisher side only avoids the serialization and conversion of the image to a ros message. On the other hand, if the subscriber is expecting a `sensor_msgs::msg::Image` that is exactly what it will receive and everything will work exactly the same as before. However, if the subscriber uses the same type adapter as the publisher, the serialization and conversion of the image to a ros message will not occur and therefor improve the communication efficiency.

## Dependencies
The type adapter used on this PR comes from the `cv_bridge` [package](https://github.com/ros-perception/vision_opencv/tree/rolling/cv_bridge). However, the type adapter was added after the humble release, in particular, on [this PR](https://github.com/ros-perception/vision_opencv/pull/441). This creates a dependency to the package version identified with the tag `3.3.0` and onwards.


Nir-Az:
@Voldivh thanks you for this work,
We will investigate it once we can and will update the results here.

Voldivh:
I'm closing this PR in order to do some refactors on the implementation of the type adapter. I'll be creating another PR in the near future with all the changes and improvements.

PR #2381, opened 2022-06-17
Publish static transforms when intra porocess communication is enabled

Yadunund:
Hi @MartyG-RealSense ,

In the existing implementation of the ROS 2 driver, if users enable `intra_process_communication`, the driver will not publish static tf transforms due to this [condition](https://github.com/IntelRealSense/realsense-ros/blob/451147de3aaa9c5b05b1fb1f7a70f3f3f43c66f6/realsense2_camera/src/base_realsense_node.cpp#L94-L98)

However there is really no need to sacrifice static tf publishing at the expense on enabling `intra_process_communication` as `rclcpp` is smart enough to selectively disable the intra process publishing mechanism for publishers (or subscribers) by suitably overwriting the `PublisherOptions` for a given publisher.

This PR enables users to benefit from both `intra_process_comminication` **as well** as static tf transforms published by this driver. It short, it overcomes the `static_tf` limitation described in the README [here](https://github.com/IntelRealSense/realsense-ros/tree/ros2-beta#limitations) 😄 

It has been tested on `foxy` and `galactic` on `Ubuntu 20.04` with a `D455` camera.

Once you rebuild the package, and load the driver component into a component container with `use_intra_process_comms:=True`, you should still be able to `ros2 topic echo /tf_static --qos-reliability reliable --qos-durability transient_local` to see the static transforms.

Signed-off-by: Yadunund <yadunund@openrobotics.org>

PR #2599, opened 2023-01-17
enable intra-process communication for point clouds

christian-rauch:
The node currently publishes point cloud messages directly and does not use a unique shared pointer, which prevents using efficient zero-copy intra-process communication as it requires copying the data between nodes in the same process.

This PR fixes this by publishing the point cloud via a unique shared pointer. I verified by the address of the `data` vector that the camera node and a point cloud subscriber indeed share that message without copying data.

The PR also contains some cleanup commits and one commit to guard different method signatures behind a version check to enable compilation on Ubuntu 20.04 with ROS2 rolling. Note that those checks should be done by library version and not ROS2 distribution names, as the same ROS2 distribution (e.g. `rolling` in this case) can have different versions of a library on different Ubuntu versions.

### DIFF
--- a/realsense2_camera/src/named_filter.cpp
+++ b/realsense2_camera/src/named_filter.cpp
@@ -240,7 +240,6 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
     }
     else
     {
-        std::string format_str = "intensity";
         msg_pointcloud->row_step = msg_pointcloud->width * msg_pointcloud->point_step;
         msg_pointcloud->data.resize(msg_pointcloud->height * msg_pointcloud->row_step);
 
--- a/realsense2_camera/src/named_filter.cpp
+++ b/realsense2_camera/src/named_filter.cpp
@@ -167,14 +167,16 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
 
     rs2_intrinsics depth_intrin = pc.get_profile().as<rs2::video_stream_profile>().get_intrinsics();
 
-    sensor_msgs::PointCloud2Modifier modifier(_msg_pointcloud);
+    sensor_msgs::msg::PointCloud2::UniquePtr msg_pointcloud = std::make_unique<sensor_msgs::msg::PointCloud2>();
+
+    sensor_msgs::PointCloud2Modifier modifier(*msg_pointcloud);
     modifier.setPointCloud2FieldsByString(1, "xyz");    
     modifier.resize(pc.size());
     if (_ordered_pc)
     {
-        _msg_pointcloud.width = depth_intrin.width;
-        _msg_pointcloud.height = depth_intrin.height;
-        _msg_pointcloud.is_dense = false;
+        msg_pointcloud->width = depth_intrin.width;
+        msg_pointcloud->height = depth_intrin.height;
+        msg_pointcloud->is_dense = false;
     }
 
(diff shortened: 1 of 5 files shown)


christian-rauch:
@Nir-Az The CI is failing for `galactic` for the same reasons as in https://github.com/IntelRealSense/realsense-ros/pull/2381#issuecomment-1383347745 (`Could not resolve the rosdep key 'ros_workspace'`). Are there plans to fix the CI for `galactic` or can we merge this PR with those pipelines failing?

Nir-Az:
> @Nir-Az The CI is failing for `galactic` for the same reasons as in [#2381 (comment)](https://github.com/IntelRealSense/realsense-ros/pull/2381#issuecomment-1383347745) (`Could not resolve the rosdep key 'ros_workspace'`). Are there plans to fix the CI for `galactic` or can we merge this PR with those pipelines failing?

Thanks for the PR, looks like a great enhancement.
I agree about the versions vs distro check, thanks for that too.
We will review the PR and test it.
The galactic fail of pre-release can be ignored, looks like a regression from `industrial-ci`
The other failure seems like a network glitch in GHA, I reran it now lets see.. should be OK.

realsense2_camera/src/named_filter.cpp
```diff
@@ -167,14 +167,16 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
 
     rs2_intrinsics depth_intrin = pc.get_profile().as<rs2::video_stream_profile>().get_intrinsics();
 
-    sensor_msgs::PointCloud2Modifier modifier(_msg_pointcloud);
+    sensor_msgs::msg::PointCloud2::UniquePtr _msg_pointcloud = std::make_unique<sensor_msgs::msg::PointCloud2>();
```
Nir-Az: Do you think that maybe in the initial implemantation reusing the member was more efficiant than creating a new one every frame?
(I am not sure..)

Nir-Az: BTW, on our naming convention a `_` prefix for a variable indicates a class member.

christian-rauch: > BTW, on our naming convention a `_` prefix for a variable indicates a class member.

That makes sense. I kept it to keep the variable name but now I removed the prefix.

> Do you think that maybe in the initial implemantation reusing the member was more efficiant than creating a new one every frame?

You may save some memory allocation with this, but the majority of the memory is taken by the dynamically sized `data`. I think it would reallocate memory anyway in some situation, e.g. if the point cloud becomes much larger.

The unique pointer has to be created for every point cloud since it is "moved" and becomes `nullptr` afterwards.

christian-rauch:
@Nir-Az Can you approve the CI pipeline for this PR?

Nir-Az (approved):
Great enhancement, thanks!

tonynajjar:
With this PR should the readme be updated? It still says that pointclouds [are not supported](https://github.com/IntelRealSense/realsense-ros/tree/ros2-development#:~:text=not%20point%2Dcloud%20or%20others) for ipc

Nir-Az:
> With this PR should the readme be updated? It still says that pointclouds [are not supported](https://github.com/IntelRealSense/realsense-ros/tree/ros2-development#:~:text=not%20point%2Dcloud%20or%20others) for ipc

You are right, thanks.
If you can PR it, it would be even awesome  :)

PR #2601, opened 2023-01-18
if default profile is not defined, take the first available profile as default

SamerKhshiboun:
Tracked by LRS-638

### DIFF
--- a/realsense2_camera/src/profile_manager.cpp
+++ b/realsense2_camera/src/profile_manager.cpp
@@ -113,7 +113,10 @@ rs2::stream_profile ProfilesManager::getDefaultProfile()
         }
     }
     if (!(default_profile.get()))
-        throw std::runtime_error("No default profile found");
+    {
+        ROS_INFO_STREAM("No default profile found. Setting the first available profile as the default one.")
+        default_profile = _all_profiles.front();
+    }
 
     return default_profile;
 }
--- a/realsense2_camera/src/profile_manager.cpp
+++ b/realsense2_camera/src/profile_manager.cpp
@@ -114,7 +114,7 @@ rs2::stream_profile ProfilesManager::getDefaultProfile()
     }
     if (!(default_profile.get()))
     {
-        ROS_INFO_STREAM("No default profile found. Setting the first available profile as the default one.")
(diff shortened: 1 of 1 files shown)


realsense2_camera/src/profile_manager.cpp
```diff
@@ -113,7 +113,10 @@ rs2::stream_profile ProfilesManager::getDefaultProfile()
         }
     }
     if (!(default_profile.get()))
-        throw std::runtime_error("No default profile found");
+    {
+        ROS_INFO_STREAM("No default profile found. Setting the first available profile as the default one.");
```
Nir-Az, line 117: I suggest adding the stream information to the log info.
Maybe something like `"No default profile found for sensor <sensor_name>, setting first profile: <stream_type, stream_index, fps, res?>
might be helpful..

SamerKhshiboun: I could do that, but chose this way, because we already print these messages today:

```
[realsense2_camera_node-1] [INFO] [1674639570.888621178] [camera.camera]: Starting Sensor: RGB Camera
[realsense2_camera_node-1] [INFO] [1674639570.918054720] [camera.camera]: Open profile: stream_type: Color(0), Format: RGB8, Width: 1280, Height: 720, FPS: 30
```

so the line I've added ("no default profile found..") will set in the middle between these two lines.

If you still prefer duplicating this info in the "no default..." line, I will add it.

Nir-Az: > I could do that, but chose this way, because we already print these messages today:
> 
> ```
> [realsense2_camera_node-1] [INFO] [1674639570.888621178] [camera.camera]: Starting Sensor: RGB Camera
> [realsense2_camera_node-1] [INFO] [1674639570.918054720] [camera.camera]: Open profile: stream_type: Color(0), Format: RGB8, Width: 1280, Height: 720, FPS: 30
> ```
> 
> so the line I've added ("no default profile found..") will set in the middle between these two lines.
> 
> If you still prefer duplicating this info in the "no default..." line, I will add it.

Great, I was not aware of it, good enough

Nir-Az (approved):
LGTM

Issue #2598, opened 2023-01-16
D405 how to turn auto exposure on while using ros2 launch realsense2_camera rs_launch.py


EkamBehl:
HI,
I've got a d405 camera .I  subscribed to /camera/color/image_rect_raw and couldn't do detection(was trying to detect chessboard corners  .I opened the image feed on realsense-viewer and it is all white due to lighting condition..The code works on the screenshot of the image after i turn auto exposure on in realsense-viewer.
Is the auto exposure on while running the launch file.If not,how do i turn it on?
i am running ros2 humble .
Thanks for any help

MartyG-RealSense:
Hi @EkamBehl  Auto-exposure should be enabled by default in both the realsense-viewer tool and the ROS wrapper.

EkamBehl:
Hi @MartyG-RealSense ,
For some reason the auto exposure is not on by default.
Is there a way to turn it on?
I've attached a few images one from rviz on the topic camera/colour/image_rect_raw
And one on real sense with auto exposure on:
![16739255126592560896709862844584](https://user-images.githubusercontent.com/62633992/212802373-bf0edbe0-14f2-428c-bc3c-33f0207b5a40.jpg)
![16739255373838255031301006107102](https://user-images.githubusercontent.com/62633992/212802465-a153a999-dff9-4e18-9068-5c3efa001eb2.jpg)

Adum888:
as a preliminary workaround you can use rqt. The check-mark for auto-exposure will be active. Deactivate it and reactivate it. Looking at the infrar image (in Rviz) should confirm the change instantly.

MartyG-RealSense:
Please try enabling the auto-exposure after launch during runtime with the ROS2 command below.

`ros2 param set /camera/camera depth_module.enable_auto_exposure true`

EkamBehl:
@MartyG-RealSense thanks it worked!!

MartyG-RealSense:
That's excellent to hear, @EkamBehl - thanks very much for the update!

MartyG-RealSense:
Case closed due to solution achieved and no further comments received.

Adum888:
How is this a sufficient solution?
a parameter not working when set during launch
Has to be set after launch (or depending on the initial configuration firstly turned false and then true again)
This is a mediocre workaround at best and not documented at all.
Edit: @EkamBehl  thank you for your tipp. It works and has to do until there is an actual fix

MartyG-RealSense:
[...]

Issue #2530, opened 2022-11-01
D455 cannot run depth at 30 HZ under ROS2


patrickpoirier51:
Hello,

Just received a D455 as replacement for D435i for a SLAM system under ROS2

Launching with this command
ros2 launch realsense2_camera rs_launch.py align_depth:=true depth_width:=640 depth_height:=480 depth_fps:=30.0 color_width:=640 color_height:=480 color_fps:=30.0

**I have 30 hz with the D435i**
[realsense2_camera_node-1] [INFO] [1667324597.517171040] [camera.camera]: Open profile: stream_type: Depth(0), Format: Z16, Width: 848, Height: 480, FPS: 30
[realsense2_camera_node-1] [INFO] [1667324597.517365971] [camera.camera]: Stopping Sensor: RGB Camera
[realsense2_camera_node-1] [INFO] [1667324597.518914167] [camera.camera]: Starting Sensor: RGB Camera
[realsense2_camera_node-1] [INFO] [1667324597.523300218] [camera.camera]: Open profile: stream_type: Color(0), Format: RGB8, Width: 1280, Height: 720, FPS: 30
[realsense2_camera_node-1] [INFO] [1667324597.523956618] [camera.camera]: RealSense Node Is Up!

![image](https://user-images.githubusercontent.com/12484190/199302369-00e47560-4b27-4544-a2ee-e763ec656836.png)

**I have 15 hz with the D455**
[realsense2_camera_node-1] [INFO] [1667324924.660251453] [camera.camera]: Open profile: stream_type: Depth(0), Format: Z16, Width: 848, Height: 480, FPS: 30
[realsense2_camera_node-1] [INFO] [1667324924.660682043] [camera.camera]: Stopping Sensor: RGB Camera
[realsense2_camera_node-1] [INFO] [1667324924.711995876] [camera.camera]: Starting Sensor: RGB Camera
[realsense2_camera_node-1] [INFO] [1667324924.721774605] [camera.camera]: Open profile: stream_type: Color(0), Format: RGB8, Width: 1280, Height: 720, FPS: 30
[realsense2_camera_node-1] [INFO] [1667324924.723208133] [camera.camera]: RealSense Node Is Up!

![image](https://user-images.githubusercontent.com/12484190/199303377-960c4659-f82b-44bf-9170-093aeec1bcf3.png)

MartyG-RealSense:
[...]

Issue #2507, opened 2022-10-19
Ros2 depth_module.auto_exposure_enabled via launch file doesnt do anything


geoeo:
System: Ubuntu 20.04 Ros2 Galactic

If I start the camera node via `ros2 launch realsense2_camera rs_record.py config_file:="'../my_config.yaml'"` with `depth_module.enable_auto_exposure: true`, the auto exposure is not enabled.

I can check the state of the parameter with `ros2 param get /camera/camera depth_module.enable_auto_exposure` and it returns true!

If I set it again via `ros2 param get /camera/camera depth_module.enable_auto_exposure true` then it works and the infra1 images changes exposure

MartyG-RealSense:
Hi @geoeo  Other RealSense ROS users have tried setting the true / false status of auto_exposure or auto_exposure_priority from roslaunch instruction, launch file or yaml.  It only works when set during runtime after launch.

**ROS1**
rosrun dynamic_reconfigure dynparam set /camera/stereo_module enable_auto_exposure 1

**ROS2-DEVELOPMENT**
ros2 param set /camera/camera depth_module.enable_auto_exposure true

geoeo:
Ok thanks for the confirmation

geoeo:
@MartyG-RealSense Btw. This seems to be a regression as it used to work under v. 2.50

MartyG-RealSense:
I have referred this issue to the RealSense ROS team.

SamerKhshiboun:
Hi @geoeo,

My name is Samer, and I'm from the RealSense ROS Team.
Can you please provide me the full command of the ros2 launch cmd ?
Also, if you can provide the launch file (or paste it here).

Thanks,
Samer

geoeo:
`ros2 launch realsense2_camera rs_launch.py config_file:="'path/to/file/launch.yaml'"`

```
camera:
    camera:
      ros__parameters:
        device_type: ''
        serial_no: ''
        usb_port_id: ''

        depth_qos: "SYSTEM_DEFAULT"
        infra_qos: "SYSTEM_DEFAULT"

        gyro_fps: 200
        accel_fps: 100

        enable_accel: true
        enable_gyro: true
        enable_depth: true
        enable_infra1: true
        enable_sync: true
        enable_color: false
        enable_infra2: false
        enable_pointcloud: false

        unite_imu_method: 1
        filters: ''

        depth_module:
            profile: 640x480x5
            emitter_enabled: 1
[...]

Issue #2503, opened 2022-10-13
ROS2-beta: RGB and depth FPS is lower than configured 


guoqingzh:
launch ROS2 realsense node and echo rgb/depth topic with ros2 topic hz. The report FPS is much lower than configurated FPS

MartyG-RealSense:
Hi @guoqingzh  If you are launching in ros2_beta with **rs_launch.py**, do you get improved FPS if you launch with **ros2 run**

`ros2 run realsense2_camera realsense2_camera_node`

guoqingzh:
> Hi @guoqingzh If you are launching in ros2_beta with **rs_launch.py**, do you get improved FPS if you launch with **ros2 run**
> 
> `ros2 run realsense2_camera realsense2_camera_node`

this is the command I used:

ros2 launch realsense2_camera rs_launch.py camera_name:='camera2' serial_no:="'_xxxxxxxxx'" rgb_camera.profile :='1280,720,30'

MartyG-RealSense:
What is the actual FPS that is being reported?

Also, can you confirm whether the camera is being detected as being on a USB 3 connection.  If it was being detected as USB 2.1 (either because the connection was mis-detected or the camera is plugged into a USB 2 port / attached to a USB 2 cable) then the only FPS speed supported for 1280x720 resolution in USB 2.1 mode would be 6 FPS (5 on the RealSense D455 camera model).

guoqingzh:
> What is the actual FPS that is being reported?
> 
> Also, can you confirm whether the camera is being detected as being on a USB 3 connection. If it was being detected as USB 2.1 (either because the connection was mis-detected or the camera is plugged into a USB 2 port / attached to a USB 2 cable) then the only FPS speed supported for 1280x720 resolution in USB 2.1 mode would be 6 FPS (5 on the RealSense D455 camera model).

It is USB3. The reported FPG is ~ 23 FPS when configured 30 FPS.

MartyG-RealSense:
Thank you.  And what is the FPS for color when launching with **ros2 run realsense2_camera realsense2_camera_node**

guoqingzh:
> Thank you. And what is the FPS for color when launching with **ros2 run realsense2_camera realsense2_camera_node**

it is the same

MartyG-RealSense:
As you are using **camera_name:='camera2'** do you have more than one camera enabled at the same time?

[...]

PR #2612, opened 2023-02-01
Fix lower FPS in D405, D455

SamerKhshiboun:
Tracked by [LRS-527], [DSO-18534]
- Added missing params in rs_launch.py: 
  - depth_module.enable_auto_exposure
  - depth_module.exposure (for manual exposure) 
  - rgb_camera.enable_auto_exposure: true

* Fixing #2507 #2598 #2503 #2530 #2486 

Description:
D405, D455 models, have default manual depth exposure value of 33000
If enable_auto_exposure is off, the depth FPS of 30 will be reduced to 15, since the manual exposure is high, the camera will try to get the first frame within this 33000 [μs], but since it very close to the 30FPS high limit (33333.33333μs) and due to other FW syncs inside the camera, we are missing this 30 FPS Goal, then we lose a cycle of frames, so the FPS is cut by half. 

So, if user want to turn off depth_module.enable_auto_exposure or use HDR features (which disable auto_exposure), he must set the depth_module.exposure to a value that is lower than ~32200 in order to bypass the problem I described above.

Updated the launch file to set depth_module.exposure to be 8500, to be aligned with D435 default manual exposure. User can override this in the launch file or in the ros2 run/launch cmd line


Nir-Az (approved):
Great investigation work!

lifeisgood-ai:
Thanks @SamerKhshiboun

PR #2619, opened 2023-02-09
Use latched QoS for Extrinsic topic when intra-process is used

Nir-Az:
Tracked on [DSO-18782]


SamerKhshiboun (approved):
LGTM.
I think we need to start documenting these things in some internal pages.
In general about RS ROS intra-process and specially for these issues.

Nir-Az:
The comment inside the code should do it and the public documentations..
It's the best place to find answers regarding the code.
If you think something is missing let me know and I will add, I want it to be self explained or documented inside `intra-process` readme that we will have :)

PR #2625, opened 2023-02-13
Modify sensor plug data

SoichiTaniguchi:
- Fixed d430 plug position
- Added plug to d435i


SamerKhshiboun:
Hi @SoichiTaniguchi,
Thanks for contribution, but we are not maintaining ros1 branches anymore.
I'm closing the PR, but users can still take these fixes manually to their repos in the future.

Issue #2617, opened 2023-02-08
Harcoded IMU frame and unpublished IMU TF


arslogavitabrevis:
ROS Version: Humble

Actual behavior:
The imu optical frame does not include the "camera_name" parameter and it seam to include the default camera_name "camera". Also, the TF for the IMU frame is not published as we can see using tf2 view_frames.

Additional remark:
The IMU_OPTICAL_FRAME_ID is defined at the line 64 of the file base_realsense_node.h and used at the line 920 of base_realsense_node.cpp. It seam that contrary to the other frames ID defined at the beginning of base_realsense_node.h the IMU_OPTICAL_FRAME_ID does not take the sip parameter.

MartyG-RealSense:
Hi @arslogavitabrevis  It sounds as though your query may be similar to another issue from a few weeks ago at https://github.com/IntelRealSense/realsense-ros/issues/2602

arslogavitabrevis:
Hi @MartyG-RealSense , thank you for following up with me on this issue. I thinks as you this issue is similar to #2602 , but it has some differences. From my understanding here are those differences:

The #2602 is noting discordance between the Readme file and the the real behavior of the robot as the parameter "base_frame_id" is supposed to drive the odom frame from what is wroten in the readme, but in reality this is the "camera_parameter" that is used to build the odom frames.

This issue is about the "camera_name" parameter kept to default value specifically for the IMU_FRAME_ID and IMU_OPTICAL_FRAME_ID. From the code, it seem that this was not the intended behavior and it sound more like a bug that a enhancement as the #2602 was tagged.

Here is some line from the lines 61-65 "base_realsense_node.h" that I think may be related to the bug:
```
#define FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_frame")).str()
#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
#define IMU_OPTICAL_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
#define OPTICAL_FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_optical_frame")).str()
[...]

PR #2627, opened 2023-02-15
Correct  IMU_FRAME and IMU_OPTICAL_FRAME definition from camera_name

arslogavitabrevis:
Related to the issue  #2617.

Now the imu topic header is based on camera_name:
```
header:
  stamp:
    sec: 1676470898
    nanosec: 797342208
  frame_id: front_camera_imu_optical_frame
orientation:
  x: 0.0
  y: 0.0
  z: 0.0
```
And the tf tree is complete:
Screenshot from 2023-02-15 09-22-43

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -59,8 +59,8 @@ using realsense2_camera_msgs::msg::Extrinsics;
 using realsense2_camera_msgs::msg::IMUInfo;
 
 #define FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_frame")).str()
-#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
-#define IMU_OPTICAL_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
+#define IMU_FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
+#define IMU_OPTICAL_FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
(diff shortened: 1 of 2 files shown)


realsense2_camera/include/base_realsense_node.h
```diff
@@ -59,8 +59,8 @@ using realsense2_camera_msgs::msg::Extrinsics;
 using realsense2_camera_msgs::msg::IMUInfo;
 
 #define FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_frame")).str()
-#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
-#define IMU_OPTICAL_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
+#define IMU_FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
+#define IMU_OPTICAL_FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
```
Nir-Az, line 63: I don't see usage of the new parameter..
Am I missing something?

arslogavitabrevis: From my test this parameter allow usage of the updated values of the variable "_camera_name". As you can see in #2617, before this modification, the default value of "_camera_name" was used. After this modification, the value of "_camera_name" used was the one defined in the parameters. I did this modification based on the others lines around that are using a similar pattern.

Nir-Az: What I meant is that your change added a macro parameter that is not used.
If we look on other macro's i.e.
```
#define FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_frame")).str()
```
You can notice the macro usage of the parameter here:
```
STREAM_NAME(sip) 
```

arslogavitabrevis: You are right this does not seem to be doing the difference. Thanks you for following with me.

PR #2658, opened 2023-03-14
fix timestamps

SamerKhshiboun:
Fix timestamp calculation metadata header to be aligned with metadata json timestamp
Tracked by [DSO-18855]

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -708,17 +708,35 @@ bool BaseRealSenseNode::setBaseTime(double frame_time, rs2_timestamp_domain time
     return false;
 }
 
+uint64_t BaseRealSenseNode::millisecondsToNanoseconds(double timestamp_ms)
+{
+        // modf breaks input into an integral and fractional part
+        double int_part_ms, fract_part_ms;
+        fract_part_ms = modf(timestamp_ms, &int_part_ms);
+
+        //convert to ns
+        uint64_t int_part_ns = static_cast<uint64_t>(int_part_ms) * 1000000;
+        uint64_t fract_part_ns = static_cast<uint64_t>(fract_part_ms * 10000000);
+
+        // Convert fract_parts_ns into 6 digits and round it
+        // to be aligned with librealsense get_timestamp API
+        fract_part_ns = (fract_part_ns % 10 > 4) ? (fract_part_ns/10 + 1) : (fract_part_ns/10);
+
+        return int_part_ns + fract_part_ns;
+}
+
 rclcpp::Time BaseRealSenseNode::frameSystemTimeSec(rs2::frame frame)
 {
+    double timestamp_ms = frame.get_timestamp();
     if (frame.get_frame_timestamp_domain() == RS2_TIMESTAMP_DOMAIN_HARDWARE_CLOCK)
     {
-        double elapsed_camera_ns = (/*ms*/ frame.get_timestamp() - /*ms*/ _camera_time_base) * 1e6;
+        double elapsed_camera_ns = millisecondsToNanoseconds(timestamp_ms - _camera_time_base);
 
         /*
         Fixing deprecated-declarations compilation warning.
         Duration(rcl_duration_value_t) is deprecated in favor of 
         static Duration::from_nanoseconds(rcl_duration_value_t)
-        starting from GALAXY.
+        starting from GALACTIC.
         */
 #if defined(FOXY) || defined(ELOQUENT) || defined(DASHING)
         auto duration = rclcpp::Duration(elapsed_camera_ns);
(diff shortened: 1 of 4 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -708,17 +708,35 @@ bool BaseRealSenseNode::setBaseTime(double frame_time, rs2_timestamp_domain time
...
 
+uint64_t BaseRealSenseNode::millisecondsToNanoseconds(double timestamp_ms)
+{
+        // modf breaks input into an integral and fractional part
+        double int_part_ms, fract_part_ms;
+        fract_part_ms = modf(timestamp_ms, &int_part_ms);
+
+        //convert to ns
+        uint64_t int_part_ns = static_cast<uint64_t>(int_part_ms) * 1000000;
+        uint64_t fract_part_ns = static_cast<uint64_t>(fract_part_ms * 10000000);
```
OhadMeir: Too many zeros.
It would be better to use a constant `static constexpr uint64_t milli_to_nano = 1000000`.
Can also use C++14 digit separator `1'000'000`   :-)

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -708,17 +708,35 @@ bool BaseRealSenseNode::setBaseTime(double frame_time, rs2_timestamp_domain time
...
+
+        // Convert fract_parts_ns into 6 digits and round it
+        // to be aligned with librealsense get_timestamp API
+        fract_part_ns = (fract_part_ns % 10 > 4) ? (fract_part_ns/10 + 1) : (fract_part_ns/10);
```
OhadMeir: More efficient `( fract_part_ns + 5 ) / 10`. No need to use `? :`.
Or use `round` function while converting to `uint64_t`

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -708,17 +708,35 @@ bool BaseRealSenseNode::setBaseTime(double frame_time, rs2_timestamp_domain time
...
+        return int_part_ns + fract_part_ns;
+}
+
 rclcpp::Time BaseRealSenseNode::frameSystemTimeSec(rs2::frame frame)
 {
+    double timestamp_ms = frame.get_timestamp();
     if (frame.get_frame_timestamp_domain() == RS2_TIMESTAMP_DOMAIN_HARDWARE_CLOCK)
     {
-        double elapsed_camera_ns = (/*ms*/ frame.get_timestamp() - /*ms*/ _camera_time_base) * 1e6;
+        double elapsed_camera_ns = millisecondsToNanoseconds(timestamp_ms - _camera_time_base);
```
OhadMeir, line 730: Converting `uint64_t` to double might loose precision.
`double` can exactly represent numbers up to 2^51 which is around 26 days in ns. Is it OK? If timestamp is since power-up then probably yes. If since epoch then surely not.

SamerKhshiboun: I think this question in more related to Librealsense API, since frame.get_timestamp is returning a double timestamp.
For this PR, I think this discussion can be skipped, but it is really worth taking this issue into account.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -708,17 +708,35 @@ bool BaseRealSenseNode::setBaseTime(double frame_time, rs2_timestamp_domain time
...
+        double int_part_ms, fract_part_ms;
+        fract_part_ms = modf(timestamp_ms, &int_part_ms);
+
+        //convert both parts to ns
+        static constexpr uint64_t milli_to_nano = 1000000;
+        uint64_t int_part_ns = static_cast<uint64_t>(int_part_ms) * milli_to_nano;
+        // for fract_part_ns, multiplie ns * 10, then divide by 10 while rounding
+        // to be aligned with librealsense get_timestamp API
+        uint64_t fract_part_ns = static_cast<uint64_t>(fract_part_ms * milli_to_nano * 10);
+        fract_part_ns = (fract_part_ns + 5)/10;
```
OhadMeir: Looking at this now it seems like a redundant division.
Use `static_cast<uint64_t>( ( fract_part_ms + 0.5 ) * milli_to_nano )` to eliminate it.

OhadMeir: Or just `std::round` that does the same. It is more readable.

OhadMeir: After consulting with StackOverflow I suggest you use `round` as the `+0.5` method does not handle all corner cases (Inf, NaN and such).

SamerKhshiboun: done

OhadMeir (approved):
LGTM

PR #2669, opened 2023-03-26
use playback device for rosbags

SamerKhshiboun:
when running a playback device, from a rosbag file, use the playback API, instead of pipe->start() which caused losing frames


Nir-Az (approved):
LGTM

PR #2674, opened 2023-03-30
fix: import materials only once

hashb:
This fixes the error `Error: material 'aluminum' is not unique`

This PR takes a different approach to fixing this issue compared to https://github.com/IntelRealSense/realsense-ros/pull/1637 and https://github.com/IntelRealSense/realsense-ros/pull/2325

### DIFF
--- a/realsense2_description/urdf/_d415.urdf.xacro
+++ b/realsense2_description/urdf/_d415.urdf.xacro
@@ -70,7 +70,7 @@ aluminum peripherial evaluation case.
           <geometry>
             <mesh filename="package://realsense2_description/meshes/d415.stl" />
           </geometry>
-          <material name="aluminum"/>
+          <xacro:material_aluminum/>
         </xacro:if>
         <xacro:unless value="${use_mesh}">
           <origin xyz="0 ${-d415_cam_depth_py} 0" rpy="0 0 0"/>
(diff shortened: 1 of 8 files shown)


SamerKhshiboun:
Hi @hashb 

and thanks for you PR.

As you mentioned there are several PRs regarding this issue:
- #2325 
- #1637 
- #2674 (This PR)

I think that all of them solve the problem, but as I see #1637 is the one with minimal changes.
Do you agree with me ?
If so, I can close other PRs and approve that PR

Thanks,
Samer

hashb:
#1637 doesn't seem to work on ROS2. Thats why I created this PR

SamerKhshiboun:
> #1637 doesn't seem to work on ROS2. Thats why I created this PR

I tried to insert that code manually to ros2-development branch, and it did work for me.
I'm thinking to duplicate that #1637 to ros2-development also.
Did you try it manually ?
https://github.com/IntelRealSense/realsense-ros/pull/1637/files

hashb:
I think I know why #1637 was failing for my use case. It appears that #1637 works when you have multiple cameras of the same type but in my use case, I have different types realsense cameras. I will add a test that verifies this scenario.

SamerKhshiboun:
Thanks @hashb,
Please update me asap so I can figure out how to fix this issue.

On my side, it worked good in both scenarios: 
with two types of same camera, and with two different cameras.

hashb:
I can confirm #1637 works for my use case. Issue I had earlier was likely a bug in my workspace. Closing this PR. Thanks @SamerKhshiboun

Issue #2522, opened 2022-10-28
[ros2-beta] Missing TFs in "/tf" topic when tf_publish_rate > 0


Ilia-Kovalev:
When parameter `tf_publish_rate`  > 0, the node [publishes](https://github.com/IntelRealSense/realsense-ros/blob/1ab947b1c1afa10199d896c11fa544ff02631925/realsense2_camera/src/base_realsense_node.cpp#L979) all TFs from`_static_tf_msgs`.
This [line](https://github.com/IntelRealSense/realsense-ros/blob/1ab947b1c1afa10199d896c11fa544ff02631925/realsense2_camera/src/rs_node_setup.cpp#L322) clears static transforms saved in the RS node before adding transforms for each sensor. After all sensors initialization, `_static_tf_msgs` has only TFs for the last sensor. 
So, dynamic transform publisher doesn't publish all TFs as static transform publisher does, but only TFs of the last initialized sensor.

It looks like a mistake, because `_static_tf_msgs`, as its name suggests, should have all static TFs inside it for the dynamic transform publisher.

MartyG-RealSense:
Hi @Tristis116  In Intel's ros2_beta wrapper documentation, it states for **tf_publish_rate** that "double, positive values mean dynamic transform publication with specified rate, all other values mean static transform publication.  Default is 0".

The documentation also states that if the **/static_tf** topic is disabled then transformations will be published if tf_publish_rate is set to 1.0 in the launch file or on the command-line using `-p tf_publish_rate:=1.0` **and** the topics `/tf` and `/extrinsic/<stream>_to_<stream>` are activated and read.

https://github.com/IntelRealSense/realsense-ros/tree/ros2-beta#limitations

Ilia-Kovalev:
Hi @MartyG-RealSense
I read about limitations. My point is: `/tf` doesn't have all transformations - only for a sensor initialized the last. And the documentation doesn't say anything about which transformations should be published and which shouldn't.
According to the code, it looks like it should duplicate `/tf_static`. But sensor initialization clears the list of transformations when it initializes every sensor.

MartyG-RealSense:
I will highlight your query to the RealSense ROS team.  Thanks very much for your patience!

MartyG-RealSense:
Hi @Tristis116  Just an update to let you know that I have been discussing your case with the RealSense ROS team.  At this time it is not known why you are experiencing the reported issue with /tf, unfortunately.

[...]

PR #2676, opened 2023-03-31
Fix: /tf and /static_tf topics' inconsistencies

Arun-Prasad-V:
Fixing #2522 

**Issues Fixed**:
1. `/static_tf` topic is getting listed even when `publish_tf:=false`
2. When the `tf_publish_rate` param is dynamically changed from 0.0 to >0.0, `/tf` topic is not getting published
3. `/tf` topic has only the TFs of last started sensor and not the TFs of previously started sensors


SamerKhshiboun (commented):
Nice catch @Arun-Prasad-V ! Working good also on my environment.
As we fix this, I think we can add a warning also when disabling the dynamic tf
so, as we have in line 989:
`ROS_WARN("Publishing dynamic camera transforms (/tf) at %g Hz", _tf_publish_rate);`
We can add something similar at line ~1002, something like:
`ROS_WARN("Publishing dynamic camera transforms to topic (/tf) has been stopped");`

so, when user do
`ros2 param set /camera/camera tf_publish_rate 0.0`
he will get warning about that

realsense2_camera/src/parameters.cpp
```diff
@@ -37,7 +37,6 @@ void BaseRealSenseNode::getParameters()
                 startDynamicTf();
             });
     _parameters_names.push_back(param_name);
-    startDynamicTf();
```
Nir-Az, line 40: Does this change the behavior from default start publishing to not?

Arun-Prasad-V: When the node is started, "**startDynamicTf()**" is getting called twice. So, the prints are also coming twice as shown below:
image

With this change, it will be called only once. 

Basic testing is done. Will test for corner cases as well.

SamerKhshiboun:
As I tested it now, if we put ```pulish_tf:=false```,
and in another terminal run```ros2 topic list```, we will still see /tf_static as topic (it is an empty one, not publishing anything)
I would prefer to not starting this topic at all.
We can do this by moving `_static_tf_broadcaster` init to be inside the publishStaticTransform
It will look something like this:

```
void BaseRealSenseNode::publishStaticTransforms(std::vector<rs2::stream_profile> profiles)
{
    // Publish static transforms
    if (_publish_tf)
    {
        if(!_static_tf_broadcaster)
        {
            // intra-process do not support latched QoS, so we need to disable intra-process for this topic
            rclcpp::PublisherOptionsWithAllocator<std::allocator<void>> options;
            options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;

#ifndef DASHING
            _static_tf_broadcaster = std::make_shared<tf2_ros::StaticTransformBroadcaster>(_node, tf2_ros::StaticBroadcasterQoS(), std::move(options));
#else
            _static_tf_broadcaster = std::make_shared<tf2_ros::StaticTransformBroadcaster>(_node, rclcpp::QoS(100), std::move(options));
#endif
        }
        
        for (auto &profile : profiles)
        {
            calcAndPublishStaticTransform(profile, _base_profile);
        }
.
.
.
```

This way, we will not see any of `/tf_static` and `/tf` topics if `publish_tf:=false`

realsense2_camera/launch/rs_launch.py
```diff
@@ -58,6 +58,7 @@
                            {'name': 'initial_reset',                'default': 'false', 'description': "''"},                           
                            {'name': 'allow_no_texture_points',      'default': 'false', 'description': "''"},                           
                            {'name': 'pointcloud.ordered_pc',        'default': 'false', 'description': ''},
+                           {'name': 'publish_tf',                   'default': 'true', 'description': 'enable/disable publishing tf'},
```
SamerKhshiboun: I would prefer to explain more in the description field, specially for these two parameters since they can confuse users.
- `publish_tf`: [bool] enable/disable publishing static or dynamic TF
- `tf_publish_rate`: [double] rate in HZ for publishing dynamic TF


Also, I would prefer to explanation and elaborate more in the README.md file, where we explain about `publish_rate `and `tf_publish_rate`
1- to explain that `publish_tf:=false` will not publish any static/dynamic TFs, even if `tf_publish_rate `is bigger than 0.0
2- to explain that `publish_tf:=true`, will automatically publish static TFs. If dynamic TFs are needed, user should set the `tf_publish_rate` for it, and it should be > 0.0
3- to mention that the default value is `publish_tf:=true` with tf_`publish_rate:=0.0` which means publishing static TFs as default behavior.

SamerKhshiboun:
Take into mind that this fix is not covering the whole problem from issue #2522 
@Tristis116 rightly claims that not all transformations are transferred when publish Dynamic TFs (/tf topic), since we clear the TF messages. Reference: look for  `_static_tf_msgs.clear();`
Every time we restarts a sensor, we clear these messages. 
If we put some debug messages, it looks like something like this:

image

The static broadcaster (`tf2_ros::StaticTransformBroadcaster`), as a Latched Topic, works like it accumulates inside him all the messages. So, clear of messages is not harming it. Every new sensor we turn on during runtime, the broadcaster will add it's TF to the rest of TFs already in his hand.
Turning off Sensors, will not change anything, and will not remove any TF msg from the published info, and this is okay.

But the dynamic broadcaster (`tf2_ros::TransformBroadcaster` )is not Latched topic, and it works like normal topics. It looks every time on the _static_tf_msgs we have, and tries to broadcast it. So usually, you will see in `/tf` topic only the last started sensor TFs, and that is a wrong behavior.

We need to figure out how to solve it, maybe to use a HashMap instead of vector of tf_msgs, (see `std::vector<geometry_msgs::msg::TransformStamped> _static_tf_msgs;`)  where the KEY is the wanted profile or something like a string TF_FROM_TO, and then we clear/fill only by this key.



P.S. 
I don't know why we publish dynamic TFs at all, since all the camera parts are statically connected.
So, in our Camera modules, dynamic TFs are the same as static TFs, but they are being published in a specific rate, while the static TFs are being published only when needed (usually once..)

Arun-Prasad-V:
Added fix for the issue "_**`/tf` topic has only the TFs of last started sensors**_"

**Root cause**:
- The TFs are stored in `static_tf_msgs` vector
- This vector is getting cleared everytime a sensor is started - that's why /tf has only TFs of last started sensor 

**Fix details**:
- Not clearing `static_tf_msgs` vector everytime
- When a sensor is started, add its corresponding TF frames to that vector
- When a sensor is stopped, erase its corresponding TF frames alone in that vector. So that the TFs of other sensors will be alive
    - The corresponding TFs to be erased are searched and identified in the vector using `frame_id` & `child_frame_id`

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -855,6 +846,37 @@ tf2::Quaternion BaseRealSenseNode::rotationMatrixToQuaternion(const float rotati
     return tf2::Quaternion(q.x(), q.y(), q.z(), q.w());
 }
 
+void BaseRealSenseNode::unpublish_static_tf(const std::string& frame_id,
+                                            const std::string& child_frame_id)
+{
+    std::vector<geometry_msgs::msg::TransformStamped>::iterator it;
+    struct find_tf
```
SamerKhshiboun: I think it is better to move this struct outside this method.
and maybe name it with another name since using find_tf with find_if might be confusing while reading the code.

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -351,6 +374,11 @@ void BaseRealSenseNode::updateSensors()
                 }
             }
         }
+        if (_publish_tf)
+        {
+            std::lock_guard<std::mutex> lock_guard(_publish_tf_mutex);
+            publishStaticTransforms();
```
SamerKhshiboun, line 365: if we publish static transforms only at this point, then I think we should change `calcAndPublishStaticTransforms` name to be just `calculcateStaticTransforms`
or even `calcualteTransforms` since we use it for static and dynamic TFs
This is affecting line 358

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -195,6 +195,26 @@ void BaseRealSenseNode::stopPublishers(const std::vector<stream_profile>& profil
         }
         _metadata_publishers.erase(sip);
         _extrinsics_publishers.erase(sip);
+
+        if (_publish_tf)
+        {
+            std::lock_guard<std::mutex> lock_guard(_publish_tf_mutex);
```
SamerKhshiboun: I think this part from line 201 to line 217 should be in a separate function.
Like we have `PublishStaticTransforms()` which uses `publish_static_tf` inside it, we can have `UnpublishStaticTransforms(<params>)` which will use `unpublish_static_tf` inside it.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -1016,12 +1052,22 @@ void BaseRealSenseNode::startDynamicTf()
             _tf_t->join();
             _tf_t.reset();
             _dynamic_tf_broadcaster.reset();
+            ROS_WARN("Stopped publishing dynamic camera transforms (/tf)");
+        }
+        else
+        {
+            ROS_WARN("Currently not publishing dynamic camera transforms (/tf)");
```
SamerKhshiboun: Just a question, can you remind me when we will get to here ? In which scenario ?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -981,28 +1003,42 @@ void BaseRealSenseNode::SetBaseStream()
...
-        for (auto &profile : profiles)
-        {
-            calcAndPublishStaticTransform(profile, _base_profile);
-        }
-        if (_static_tf_broadcaster)
-            _static_tf_broadcaster->sendTransform(_static_tf_msgs);
+        _static_tf_broadcaster.reset();
     }
+
+    // intra-process do not support latched QoS, so we need to disable intra-process for this topic
```
SamerKhshiboun: I would move this part from line 1008 to line 1024 into a seperate function
`resetStaticTransformBroadcaster(_static_tf_broadcaster)`

SamerKhshiboun:
Great job @Arun-Prasad-V.
I think now we cover all flows and it is working good.
I gave you some comments, and I think we can move all `base_realsense::<TFs functions>` to a new cpp file `tfs.cpp` to be dedicated to all TF methods (no need for header file. see `parameters.cpp` for example).
If you do that, remember to add this cpp to the cmake file

SamerKhshiboun (approved):
LGTM

PR #2680, opened 2023-04-02
fix imu topic frame_id with came_name ns

brsbilgin:
The imu topic frame was published  with default name space which is "camera" even i gave different name to the camera_name parameter. The frame name was always "camera_imu_optical_frame" never changed according the camera_name

### DIFF
--- a/realsense2_camera/include/base_realsense_node.h
+++ b/realsense2_camera/include/base_realsense_node.h
@@ -59,8 +59,8 @@ using realsense2_camera_msgs::msg::Extrinsics;
 using realsense2_camera_msgs::msg::IMUInfo;
 
 #define FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_frame")).str()
-#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
-#define IMU_OPTICAL_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
+#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name <<  "_" << "imu_frame")).str()
+#define IMU_OPTICAL_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << "imu_optical_frame")).str()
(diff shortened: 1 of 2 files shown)


SamerKhshiboun:
Hi @brsbilgin 

I think that the current code is behaving as expected, and no need for these changes
The camera name is working good, as also the axes both in ROS CS and in Optical CS.
with PR #2684 applied (should be merged soon), these are the screenshots from rviz2 after running with D435i:
```
ros2 launch realsense2_camera rs_launch.py camera_name:=test_name enable_accel:=true enable_gyro:=true unite_imu_method:=2
```

image
image

realsense2_camera/include/base_realsense_node.h
```diff
@@ -59,8 +59,8 @@ using realsense2_camera_msgs::msg::Extrinsics;
 using realsense2_camera_msgs::msg::IMUInfo;
 
 #define FRAME_ID(sip) (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << STREAM_NAME(sip) << "_frame")).str()
-#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_frame")).str()
-#define IMU_OPTICAL_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_imu_optical_frame")).str()
+#define IMU_FRAME_ID (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name <<  "_" << "imu_frame")).str()
```
SamerKhshiboun, line 62: already working good.
why to split "_" from the rest of the const string ?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -286,13 +286,13 @@ sensor_msgs::msg::Imu BaseRealSenseNode::CreateUnitedMessage(const CimuData acce
     rclcpp::Time t(gyro_data.m_time_ns);  //rclcpp::Time(uint64_t nanoseconds)
     imu_msg.header.stamp = t;
 
-    imu_msg.angular_velocity.x = gyro_data.m_data.x();
-    imu_msg.angular_velocity.y = gyro_data.m_data.y();
-    imu_msg.angular_velocity.z = gyro_data.m_data.z();
+    imu_msg.angular_velocity.x = -gyro_data.m_data.z();
```
SamerKhshiboun, line 289: no need to swap axes, these gyro frame is in ROS CS, and IMU frame should be in same coordinates.
this is not the optical one, which we publish later in other function.

SamerKhshiboun:
Closing this PR.
See https://github.com/IntelRealSense/realsense-ros/pull/2684 for more details

Issue #2637, opened 2023-02-28
Static transforms are inverted


MasonLilly:
We recently updated our system from using the ROS 1 realsense node (commit `44a703ed`) to the ROS 2 version (commit `839d8378`). Since the upgrade, we've found that the static transforms published by the camera node have their translations inverted compared to the old version.

From a backup still running ROS 1:
```bash
root# rosrun tf tf_echo camera_front_link camera_front_color_frame
At time 0.000
- Translation: [-0.000, -0.059, 0.000]
- Rotation: in Quaternion [-0.000, -0.002, -0.001, 1.000]
            in RPY (radian) [-0.000, -0.005, -0.001]
            in RPY (degree) [-0.026, -0.259, -0.073]
root# rosrun tf tf_echo camera_front_color_frame camera_front_color_optical_frame

At time 0.000
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [-0.500, 0.500, -0.500, 0.500]
            in RPY (radian) [-1.571, -0.000, -1.571]
            in RPY (degree) [-90.000, -0.000, -90.000]
root# rosrun tf tf_echo camera_front_depth_frame camera_front_depth_optical_frame

At time 0.000
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [-0.500, 0.500, -0.500, 0.500]
            in RPY (radian) [-1.571, -0.000, -1.571]
            in RPY (degree) [-90.000, -0.000, -90.000]
```

From the new system using ROS 2:
```bash
~$ ros2 run tf2_ros tf2_echo camera_front_link camera_front_color_frame
At time 0.0
- Translation: [0.000, 0.059, 0.000]
- Rotation: in Quaternion [0.000, 0.002, -0.000, 1.000]
~$ ros2 run tf2_ros tf2_echo camera_front_color_frame camera_front_color_optical_frame
At time 0.0
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [-0.500, 0.500, -0.500, 0.500]
~$ ros2 run tf2_ros tf2_echo camera_front_depth_frame camera_front_depth_optical_frame
At time 0.0
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [-0.500, 0.500, -0.500, 0.500]
```

[...]

Issue #2583, opened 2022-12-21
Static TFs appear inverted


alexmillane:
When I look at the static transforms published by the `realsense2_camera_node` (after running `ros2 run realsense2_camera realsense2_camera_node`) the translation between the camera optical frames appears to be inverted.

For example in the image below the infra1 image comes from the left camera but appears on the right in the tf tree.

Am I missing something here?

![d455_frames](https://user-images.githubusercontent.com/671701/208834125-5e0d3181-2b39-482b-a7be-c3669ac7cdc4.png)

MartyG-RealSense:
Hi @alexmillane  On RealSense 400 Series cameras, the left and right positions of the infrared sensors are from the perspective of standing behind the camera looking forwards.  Because of this, if you look at the camera from the front then the left infrared sensor is on the right side of the camera.  This is illustrated at https://github.com/IntelRealSense/librealsense/issues/3662

alexmillane:
Hi Marty. Thank you very much for the reply.

So when I say left in my original post, I mean from the perspective of behind the camera. But that is good to clarify, thank you.

In any case, if I look at the infra1 image above, it appears to me to be taken from a perspective slightly to the left of the perspective of the infra2 image. But this left-side positioning of the infra1 image is inconsistent with the TF tree, which puts infra1 on the right.

Moreover, the reason I actually started looking at this is that when I project the depth pointcloud between the cameras on the realsense body, I get incorrect results unless I invert the cam-to-cam transforms I get from the tf tree. After inversion, everything works perfectly. This leads me to believe that the tf tree is incorrect.

Do you have another interpretation of the image above?

MartyG-RealSense:
The only other relevant discussion I could find on this subject was at https://github.com/introlab/rtabmap_ros/issues/517 where infra1 and infra2 were on the opposite sides to your image, but the RealSense user in that case was receiving an error message about the positions.

![image](https://user-images.githubusercontent.com/41145062/208869715-7db40582-97b1-45c6-97a0-4da735c76205.png)

alexmillane:
Thank you for the link. That picture looks correct to me. Perhaps a mistake has been introduced since that issue was posted? 

I might have a deeper look in January and make a pull request if I find a solution.

In the meantime, if anyone else believes that the tf tree published by the realsense node is correct, I'm interested in hearing about it :).

MartyG-RealSense:
[...]

Issue #2500, opened 2022-10-12
Why are the axes swapped when publishing static tf?


schmidtp1:
https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/src/base_realsense_node.cpp#L1991 ff.
seems like it shouldn't be there/necessary if the transformations are done correctly. can anyone provide insight why this is currently required? thanks

MartyG-RealSense:
Hi @schmidtp1  Looking at that section of code, it may be taking account of ROS using X-forward, Y-left, Z-up as its standard frame_id, whilst in the RealSense coordinate system Z is forward and Y is up.  This subject is discussed at https://github.com/IntelRealSense/realsense-ros/issues/1099

schmidtp1:
> Hi @schmidtp1 Looking at that section of code, it may be taking account of ROS using X-forward, Y-left, Z-up as its standard frame_id, whilst in the RealSense coordinate system Z is forward and Y is up. This subject is discussed at #1099

but as it looks this is happening for all the static/internal transformations (that should be given by the CAD model /calibration) and not just for the reference frame.
also why is the orientation (quaternion) not rotated then?
also doesn't it seem the wrong place to do this (or at least unexpected) when one would just expect "publishing" (based on the function name)?

MartyG-RealSense:
I will discuss this section of base_realsense_node.cpp code with my Intel RealSense colleagues.  Thanks very much for your patience!

schmidtp1:
> I will discuss this section of base_realsense_node.cpp code with my Intel RealSense colleagues. Thanks very much for your patience!

thank you for your quick response, your insight and for following up on this

schmidtp1:
> I will discuss this section of base_realsense_node.cpp code with my Intel RealSense colleagues. Thanks very much for your patience!

to provide a bit more context, i suspect a bug in the transformation(s) there as well as in this function: https://github.com/IntelRealSense/realsense-ros/blob/development/realsense2_camera/src/t265_realsense_node.cpp#L94
based on that when i'm getting the transformation between the sensors from ROS, it is (slightly) different than what I'm getting from LRS, see below.

if it is agreed that this is not the expected behavior, can this issue be promoted to a bug? (@MartyG-RealSense)
[...]

PR #2684, opened 2023-04-05
Fix Swapped TFs Axes

SamerKhshiboun:
Fixing issues #2500 #2637 #2583 
These issues related to static TFs and Extrinsic

### Background and Terminology:
- Point of view: Imagine we are standing behind of the camera, and looking forward.
- ROS2 Coordinates System (X: Forward, Y:Left, Z: Up). Everything related to TFs should be calculated in this coordination system.
- Camera Optical Coordinates System (X: Right, Y: Down, Z: Forward). Everything related to extrinsics should be calculated in this coordination system.
image
- Static TF msg expresses a transform from coordinate frame header.frame_id (parent) to the coordinate frame child_frame_id [Reference](http://docs.ros.org/en/noetic/api/geometry_msgs/html/msg/Transform.html)
- Extrinsic from sensor A to sensor B means the position and orientation of sensor A relative to sensor B
- As you see, TF(A->B) translation should be the inversion of Extrinsic(A->B)
- In a previous [PR](https://github.com/IntelRealSense/realsense-ros/pull/2477), we fixed the Extrinsic calculation, which made the /camera/extrinsics/<A_TO_B> correct.
**For example, depth_to_color, in D435i:**
If we look from behind of the D435i, extrinsic from depth to color, means, where is the depth in relative to the color.
If we just look at X coordinates (again, from behind), we can say that Depth is on the right of RGB (color) sensor by 0.0148m (1.48cm), which is correct.

```
administrator@perclnx466 ~/ros2_humble $ ros2 topic echo /camera/extrinsics/depth_to_color
rotation:
...
translation:
- 0.01485931035131216
- 0.0010161789832636714
- 0.0005317096947692335
---
```
image

The problem we are trying to fix in this PR, is that when publishing this data as TF message, in the ROS2 coordinates system, we forgot to invert the extrinsic data. Please read the top of this description again if you still confused why we need to invert extrinsic data when we are moving from Optical coordinates system to ROS2 coordinates system.

Attaching rviz2 screenshots that describe the change.

### Before the fix:
You can see that INFRA1 is in the right side, and INFRA2 is in the left, while we are looking from behind of the camera, and this is wrong !
before_the_fix

### After the fix:
after_fix_camera
after_fix_2

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -865,9 +865,13 @@ void BaseRealSenseNode::publish_static_tf(const rclcpp::Time& t,
     msg.header.stamp = t;
     msg.header.frame_id = from;
     msg.child_frame_id = to;
+
+    // Convert x,y,z (taken from camera extrinsics)
+    // from optical cooridnates to ros coordinates
     msg.transform.translation.x = trans.z;
     msg.transform.translation.y = -trans.x;
     msg.transform.translation.z = -trans.y;
+
     msg.transform.rotation.x = q.getX();
     msg.transform.rotation.y = q.getY();
     msg.transform.rotation.z = q.getZ();
@@ -917,7 +921,10 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
     auto Q = rotationMatrixToQuaternion(ex.rotation);
     Q = quaternion_optical * Q * quaternion_optical.inverse();
 
-    float3 trans{ex.translation[0], ex.translation[1], ex.translation[2]};
+    // Invert x,y,z of extrinsic translation before sending this as TF translation
+    // because extrinsic from A to B is the position of A relative to B
+    // while TF from A to B is the transformation to be done on A to get to B
+    float3 trans{-ex.translation[0], -ex.translation[1], -ex.translation[2]};
     publish_static_tf(transform_ts_, trans, Q, _base_frame_id, FRAME_ID(sip));
 
     // Transform stream frame to stream optical frame
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -888,30 +888,6 @@ void BaseRealSenseNode::publishExtrinsicsTopic(const stream_index_pair& sip, con
     }
 }
 
-
-rs2_extrinsics BaseRealSenseNode::invertExtrinsics(const rs2_extrinsics& ex) const
-{
-    rs2_extrinsics result;
-
-    // invert translation vector
-    result.translation[0] = -ex.translation[0];
-    result.translation[1] = -ex.translation[1];
-    result.translation[2] = -ex.translation[2];
-
-    // invert rotation matrix
-    result.rotation[0] = ex.rotation[0];
-    result.rotation[1] = ex.rotation[3];
-    result.rotation[2] = ex.rotation[6];
-    result.rotation[3] = ex.rotation[1];
-    result.rotation[4] = ex.rotation[4];
-    result.rotation[5] = ex.rotation[7];
-    result.rotation[6] = ex.rotation[2];
-    result.rotation[7] = ex.rotation[5];
-    result.rotation[8] = ex.rotation[8];
-
-    return result;
-}
-
 void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile& profile, const rs2::stream_profile& base_profile)
 {
     // Transform base to stream
@@ -924,17 +900,25 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
 
     rclcpp::Time transform_ts_ = _node.now();
 
-    rs2_extrinsics ex;
+    // extrinsic from A to B is the position of A relative to B
+    // TF from A to B is the transformation to be done on A to get to B
+    // so, we need to calculate extrinsics in two opposite ways, one for extrinsic topic
+    // and the second is for transformation topic (TF)
+    rs2_extrinsics normal_ex;  // used to for extrinsics topic
+    rs2_extrinsics tf_ex; // used for TF
+
     try
     {
-        ex = base_profile.get_extrinsics_to(profile);
+        normal_ex = base_profile.get_extrinsics_to(profile);
+        tf_ex = profile.get_extrinsics_to(base_profile);
     }
     catch (std::exception& e)
     {
         if (!strcmp(e.what(), "Requested extrinsics are not available!"))
         {
-            ROS_WARN_STREAM("(" << rs2_stream_to_string(profile.stream_type()) << ", " << profile.stream_index() << ") -> (" << rs2_stream_to_string(base_profile.stream_type()) << ", " << base_profile.stream_index() << "): " << e.what() << " : using unity as default.");
-            ex = rs2_extrinsics({{1, 0, 0, 0, 1, 0, 0, 0, 1}, {0,0,0}});
+            ROS_WARN_STREAM("(" << rs2_stream_to_string(base_profile.stream_type()) << ", " << base_profile.stream_index() << ") -> (" << rs2_stream_to_string(profile.stream_type()) << ", " << profile.stream_index() << "): " << e.what() << " : using unity as default.");
+            normal_ex = rs2_extrinsics({{1, 0, 0, 0, 1, 0, 0, 0, 1}, {0,0,0}});
+            tf_ex = normal_ex;
         }
         else
         {
@@ -942,20 +926,16 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
         }
     }
 
-    publishExtrinsicsTopic(sip, ex);
+    // publish normal extrinsics e.g. /camera/extrinsics/depth_to_color
+    publishExtrinsicsTopic(sip, normal_ex);
 
-    // Invert extrinsic translation vector and rotation matrix before sending this as TF translation
-    // because extrinsic from A to B is the position of A relative to B
-    // while TF from A to B is the transformation to be done on A to get to B
-    ex = invertExtrinsics(ex);
-
-    auto Q = rotationMatrixToQuaternion(ex.rotation);
+    // publish static TF
+    auto Q = rotationMatrixToQuaternion(tf_ex.rotation);
     Q = quaternion_optical * Q * quaternion_optical.inverse();
-    float3 trans{ex.translation[0], ex.translation[1], ex.translation[2]};
-
+    float3 trans{tf_ex.translation[0], tf_ex.translation[1], tf_ex.translation[2]};
     publish_static_tf(transform_ts_, trans, Q, _base_frame_id, FRAME_ID(sip));
 
-    // Transform stream frame to stream optical frame
+    // Transform stream frame to stream optical frame and publish it
     publish_static_tf(transform_ts_, zero_trans, quaternion_optical, FRAME_ID(sip), OPTICAL_FRAME_ID(sip));
 
     if (profile.is<rs2::video_stream_profile>() && profile.stream_type() != RS2_STREAM_DEPTH && profile.stream_index() == 1)
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -888,6 +888,30 @@ void BaseRealSenseNode::publishExtrinsicsTopic(const stream_index_pair& sip, con
     }
 }
 
+
+rs2_extrinsics BaseRealSenseNode::invertExtrinsics(const rs2_extrinsics& ex) const
+{
+    rs2_extrinsics result;
+
+    // invert translation vector
+    result.translation[0] = -ex.translation[0];
+    result.translation[1] = -ex.translation[1];
+    result.translation[2] = -ex.translation[2];
+
+    // invert rotation matrix
+    result.rotation[0] = ex.rotation[0];
+    result.rotation[1] = ex.rotation[3];
+    result.rotation[2] = ex.rotation[6];
+    result.rotation[3] = ex.rotation[1];
+    result.rotation[4] = ex.rotation[4];
+    result.rotation[5] = ex.rotation[7];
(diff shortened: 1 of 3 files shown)


guoqingzh:
@SamerKhshiboun  how about IMU extrinsic ? Are they following the Camera Optical Coordinate System as well ?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -917,7 +921,10 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
     auto Q = rotationMatrixToQuaternion(ex.rotation);
     Q = quaternion_optical * Q * quaternion_optical.inverse();
 
-    float3 trans{ex.translation[0], ex.translation[1], ex.translation[2]};
+    // Invert x,y,z of extrinsic translation before sending this as TF translation
+    // because extrinsic from A to B is the position of A relative to B
+    // while TF from A to B is the transformation to be done on A to get to B
+    float3 trans{-ex.translation[0], -ex.translation[1], -ex.translation[2]};
```
Gilaadb: What about the rotation?
I think a more correct way to handle it would be to replace line #906: ex = base_profile.get_extrinsics_to(profile);
with: ex = profile.get_extrinsics_to(base_profile);

In fact, I see it was that way before a PR from september 2022: https://github.com/IntelRealSense/realsense-ros/commit/61406afa1e32b024992606626d29e57258590640

Maybe that PR was a mistake or did it solve a real issue?

Gilaadb: After discussing this offline, I understood why the change was introduced in the first place. Samer will check if the "get_extrinsics_to()" function call is "heavy" and if not, we might have one for publishing to the extrinsic topic and another one with opposite order of frames to publish to the tf topic.

Nir-Az: We will have to explain it well in the code, hard to understand why extrinsic msg and static_tf should behave differently.

SamerKhshiboun: A quick example on D435i why we needed the change of #2477:

**Before the fix of PR #2477:**

```administrator@perclnx466 ~/ros2_humble $ ros2 topic echo /camera/extrinsics/depth_to_color
rotation:
- 0.9999583959579468
- -0.008895229548215866
- 0.0020131953060626984
- 0.008895332925021648
- 0.9999604225158691
- -4.254872692399658e-05
- -0.0020127370953559875
- 6.045500049367547e-05
- 0.9999979734420776
translation:
- -0.0148666612803936
- -0.0008839939837343991
- -0.0005615801201201975
```

**After the fix of PR #2477:**

```administrator@perclnx466 ~/ros2_humble $ ros2 topic echo /camera/extrinsics/depth_to_color
rotation:
- 0.9999583959579468
- 0.008895332925021648
- -0.0020127370953559875
- -0.008895229548215866
- 0.9999604225158691
- 6.045500049367547e-05
- 0.0020131953060626984
- -4.254872692399658e-05
- 0.9999979734420776
translation:
- 0.01485931035131216
- 0.0010161789832636714
- 0.0005317096947692335
```


To make sure which result is the right one, all we need is to compare it to `rs-enumare-devices -c `while running the same camera. Somewhere in the output we get this:

```
Extrinsic from "Depth"    To      "Color" :
 Rotation Matrix:
   0.999958        -0.00889523       0.0020132
   0.00889533       0.99996         -4.25487e-05
  -0.00201274       6.0455e-05       0.999998

 Translation Vector: 0.0148593103513122  0.00101617898326367  0.000531709694769233
```


And that makes sense with the new result after the fix.
If we "imagine" that color sensor is (0,0,0), then Depth is on the right of this sensor by 0.01485 m (~1.5 cm), and this is the result that should be published in the relevant topic of the 
**/camera/extrinsics/depth_to_color**

SamerKhshiboun: Now, the next explanation is to elaborate more why we need in addition to calculate extrinsic in the opposite way as @Gilaadb suggest.

The first extrinsic we calculated above, was used to publish the depth_to_color topic.
As I described, and as you see the results, it shows where is "Depth" in the eyes of the "Color".

To calculate TF from depth to color, we need the something opposite: how to go from Depth To Color ?
We can manipulate the Extrinsic result we already calculated above, and do some inversion, to get the right direction for TFs, but as Gilad suggested, it may also be complicated a little bit, especially with the ROTATION matrix, so he suggested to calculated two things, instead of inverting the rotation-matrix/translation:

ex1 = base_profile.get_extrinsics_to(profile);  // will be used for depth_to_color topic
ex2 = profile.get_extrinsics_to(base_profile); // will be used for the  Depth To Color TF information

SamerKhshiboun: @Gilaadb ,

After a deeper investigation and testing, I would prefer calculating the extrinsic only once, publish it to the extrinsic topic, and then to manipulate it (invert Matrix and Translation-Vector) for TF topics.

I tested with ex1 and ex2, and it looks like it add some time to fetch the extrinsic in both ways, as it goes and enumerate on all extrinsic of the camera (twice).
But I learned from that, that also the Rotation Matrix should be flipped (inverted), so I will fix tomorrow and update the code.

so for example, we get today:

```rotation:
      x: 0.004447687417268753
      y: -2.575120197434444e-05
      z: -0.0010064936941489577
      w: 0.9999895691871643
```

but it should be:
```
 rotation:
      x: -0.004447687417268753
      y: 2.575120197434444e-05
      z: 0.0010064936941489577
      w: 0.9999895691871643
```

so, we have to invert X,Y,Z also in the Rotation, and I'm on it to find the best way for doing that.

SamerKhshiboun:
Pushed a new commit that fixes the Matrix rotation also.

I just saw something weird to me... maybe related to calibration info inside the camera
from rs-enumerate-device 


```
Extrinsic from "Color"    To      "Depth" :
 Rotation Matrix:
   0.999958         0.00889533      -0.00201274
  -0.00889523       0.99996          6.0455e-05
   0.0020132       -4.25487e-05      0.999998

 Translation Vector: -**0.0148666612803936**  **-0.000883993983734399**  **-0.000561580120120198**


Extrinsic from "Depth"    To      "Color" :
 Rotation Matrix:
   0.999958        -0.00889523       0.0020132
   0.00889533       0.99996         -4.25487e-05
  -0.00201274       6.0455e-05       0.999998

 Translation Vector: **0.0148593103513122**  **0.00101617898326367**  **0.000531709694769233**
```

why does there are small differences between some translation vector values ?  (except they are inverted)
for example TranslationVector1.y = -0.000883993983734399
while TranslationVector2.y = 0.00101617898326367

Gilaadb:
> why does there are small differences between some translation vector values ? (except they are inverted) for example TranslationVector1.y = -0.000883993983734399 while TranslationVector2.y = 0.00101617898326367

That's because we also need to account for the rotation. The correct way to calculate it is to invert the matrix (calculated in matlab):
A =

   0.999958000000000   0.008895330000000  -0.002012740000000  -0.014866661280394
  -0.008895230000000   0.999960000000000   0.000060455000000  -0.000883993983734
   0.002013200000000  -0.000042548700000   0.999998000000000  -0.000561580120120
                   0                   0                   0   1.000000000000000

>> B = inv(A)

B =

   0.999958820147486  -0.008895233841980   0.002013198903423   0.014859316312756
   0.008895341189744   0.999960869629048  -0.000042548710445   0.001016179522653
  -0.002012742637003   0.000060455040734   0.999997945213466   0.000531709645056
                   0                   0                   0   1.000000000000000

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -865,9 +865,13 @@ void BaseRealSenseNode::publish_static_tf(const rclcpp::Time& t,
     msg.header.stamp = t;
     msg.header.frame_id = from;
     msg.child_frame_id = to;
+
+    // Convert x,y,z (taken from camera extrinsics)
+    // from optical cooridnates to ros coordinates
     msg.transform.translation.x = trans.z;
```
Nir-Az, line 871: Let's verify if IMU coordinates need to change to match this (like in PR  #2680  )

SamerKhshiboun:
Ok, so I used two extrinsic at the end, as it being called only once for each stream in our code.
This saves dealing with rotation matrix calculations/inversion

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: After a deeper check, and after review with @Gilaadb, no need for the PR #2680, since all IMU frames and IMU optical frames are looking good.

SamerKhshiboun:
waiting for @guoqingzh review to confirm this PR code.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -896,31 +900,42 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
 
     rclcpp::Time transform_ts_ = _node.now();
 
-    rs2_extrinsics ex;
+    // extrinsic from A to B is the position of A relative to B
+    // TF from A to B is the transformation to be done on A to get to B
+    // so, we need to calculate extrinsics in two opposite ways, one for extrinsic topic
+    // and the second is for transformation topic (TF)
+    rs2_extrinsics normal_ex;  // used to for extrinsics topic
+    rs2_extrinsics tf_ex; // used for TF
```
guoqingzh, line 908: Feel like a proper way is to define a class called rs2_ros_tf , the class should have a constructor which takes rs2_extrinsics as input , the conversion should be done inside the constructor. Current fix defines both normal_ex and tf_ex as rs2_extrinsics but the tf_ex is actually ros tf

guoqingzh (changes requested):

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: The problem is that doing the conversion/transformation inside the class you are suggesting, requires inverting the rotation matrix, and I found it complex to do that. 
Instead, I called the extrinsic API of librealsense twice, in two opposite ways, and that solved the problem.

Nir-Az: @guoqingzh can you please first verify you agree with the outputs?
We focus on Extrinsics + static tf of the video + IMU streams.
Any coding suggestions are always welcome but should not block this PR.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -934,8 +949,6 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
         publish_static_tf(transform_ts_, zero_trans, zero_rot_quaternions, FRAME_ID(sip), IMU_FRAME_ID);
         publish_static_tf(transform_ts_, zero_trans, quaternion_optical, IMU_FRAME_ID, IMU_OPTICAL_FRAME_ID);
```
guoqingzh, line 950: I think this is wrong. The IMU data published by libRS is already aligned with depth CS (x-right, y-down, z-front) according to the code-> https://github.com/IntelRealSense/librealsense/blob/master/src/proc/motion-transform.cpp#L91
The rotation from IMU_FRAME_ID to IMU_OPTICAL_FRAME_ID should be identity matrix IMO

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -896,31 +900,42 @@ void BaseRealSenseNode::calcAndPublishStaticTransform(const rs2::stream_profile&
...
 
-    auto Q = rotationMatrixToQuaternion(ex.rotation);
-    Q = quaternion_optical * Q * quaternion_optical.inverse();
+    // publish normal extrinsics e.g. /camera/extrinsics/depth_to_color
+    publishExtrinsicsTopic(sip, normal_ex);
 
-    float3 trans{ex.translation[0], ex.translation[1], ex.translation[2]};
+    // publish static TF
+    auto Q = rotationMatrixToQuaternion(tf_ex.rotation);
+    Q = quaternion_optical * Q * quaternion_optical.inverse();
```
guoqingzh, line 934: what are you trying to achieve with this line ? q*p*q-1 is meant for transformation a vector from CS A to CS B but you have Q in the middle, which is a rotation quaternion. I don't understand the math here. You have relative rotation from tf_ex already, why you need to rotate it again with the quaternion_optical ? Isn't that the tf_ex already assume optical frame ?

guoqingzh (changes requested):
have to block the PR, because I think the static_tf for IMU is wrong. And also I don't understand the transformation in line 934

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: Actually I did not add this line, it was there for a couple of years.
see this commit:
https://github.com/IntelRealSense/realsense-ros/commit/1cd63b675548b4603b78990de3422fd53e59b39c

and it was working all the time.
I don't know yet if it necessary or not.

Nir-Az: @guoqingzh This added by PR #383 , please review it

Could be that Q here is acctually P from the original equation 
Screenshot_20230410_235742

guoqingzh: @Nir-Az As your screenshot told, the expression rotates vector quaternion (not rotation quaternion). I read the original PR#383, I think the original author is trying to do a rotation combination here. In our code, Q is optical to optical , which can't be used directly in ros tf so he has to rotate first from ros to optical (quaternion_optical inverse), then apply Q, after that apply quaternion_optical again to rotate from optical back to ros, i think it is ok

realsense2_camera/src/base_realsense_node.cpp
guoqingzh: @Nir-Az @SamerKhshiboun notified you set IMU frame_id to IMU_OPTICAL_FRAME_ID by default (BaseRealSenseNode::ImuMessage_AddDefaultValues()), so i think it is ok as well

guoqingzh (approved):
IMU messgae is associated with IMU optical frame and quaternion_optical*Q*quaternion_optical.inv() is a rotation combination rather than vector transformation.  I will approve this

Nir-Az (approved):
Excellent work

PR #2709, opened 2023-04-25
Fixing TF frame links b/w multi camera nodes when using custom names

Arun-Prasad-V:
When two camera nodes are launched using *multi_camera_launch.py*, 
- if default camera names are used, the TF frames of those two cameras are linked as below:
image

- if customs camera names are used, they are not linked
image

### DIFF
--- a/realsense2_camera/launch/rs_multi_camera_launch.py
+++ b/realsense2_camera/launch/rs_multi_camera_launch.py
@@ -23,9 +23,9 @@
 
 """Launch realsense2_camera node."""
 import copy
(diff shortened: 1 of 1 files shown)


Arun-Prasad-V:
Root cause:
- The TF frame IDs are hardcoded in multi_camera_launch.py with default camera names

Fix:
- Get the camera names during run time from LaunchConfiguration and update the TF frame IDs

SamerKhshiboun (approved):
Great, Works perfectly.

PR #2726, opened 2023-05-05
Integration test template

PrasRsRos:
Update the test infra


SamerKhshiboun (approved):
LGTM

Nir-Az (approved):
Great work!

Nir-Az:
@PrasRsRos , Note that industrial_ci is also running the tests and they failed.

This means that if we bloom release with the PR merged we will be blocked.

Please investigate the failures.

We need it "green" in industrial_co too.

SamerKhshiboun:
> @PrasRsRos , Note that industrial_ci is also running the tests and they failed.
> 
> This means that if we bloom release with the PR merged we will be blocked.
> 
> Please investigate the failures.
> 
> We need it "green" in industrial_co too.

Both Iron and Humble industrial_ci are failing on:
**ModuleNotFoundError: No module named 'launch_pytest'**

We need to tell industrial_ci that we want to install some modules before running.
I think this link explains how to use "BEFORE_INIT" parameter to do that:
https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#customize-within-the-ci-process

realsense2_camera/package.xml
```diff
@@ -28,6 +28,9 @@
   <depend>tf2</depend>
   <depend>tf2_ros</depend>
   <depend>diagnostic_updater</depend>
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>launch_testing</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
```
Nir-Az, line 33: Maybe missing `launch_pytest`?
if we have it here, it will be installed automatically as part of our building flow IMO.

BTW, whats the difference between `launch_testing` and `launch_pytest`?

.travis.yml
```diff
@@ -3,12 +3,14 @@ matrix:
     include:
         - dist: bionic
         - dist: focal
+        - dist: jammy
```
SamerKhshiboun, line 6: is this file in use anywhere ?

SamerKhshiboun:
Hi @PrasRsRos,

Please review/answer two comments/questions on this PR.
Great work! waiting for GitHub actions to finish.

@Nir-Az, you can merge after Prasanth answers the questions and GHA success. I think you can squash commits before merging right ?

realsense2_camera/package.xml
PrasRsRos: @Nir-Az,  luanch_pytest is a relative new package that helps us filter tests based on markers. This is good for RS test grouping, as we can have test markers based on rosbag, devices (d435, d455 etc.). The package was already added as a dependency to the test.

.travis.yml
PrasRsRos: @SamerKhshiboun, I am not aware of any usage of it in the repo. May be, we can check if it's used in the dailies/weeklies? If not, we can remove this.

PR #2735, opened 2023-05-10
<material name="aluminum"/> is already defined as _materials.urdf.xacro

YUKINA-3252:
`<material name="aluminum"/>` is already defined in   `<xacro:include filename="$(find realsense2_description)/urdf/_materials.urdf.xacro" />`


YUKINA-3252:
sorry, this has been solved by https://github.com/IntelRealSense/realsense-ros/pull/1637, but not been released as deb. You need to check out ros1-legacy branch

PR #2757, opened 2023-05-29
Support Depth Mapping Streams

SamerKhshiboun:


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -785,6 +803,59 @@ void BaseRealSenseNode::publishPointCloud(rs2::points pc, const rclcpp::Time& t,
...
+    modifier.resize(pc.size());
+
+    // Fill the PointCloud message with data
+    sensor_msgs::PointCloud2Iterator<float> iter_x(*msg_pointcloud, "x");
+    sensor_msgs::PointCloud2Iterator<float> iter_y(*msg_pointcloud, "y");
+    sensor_msgs::PointCloud2Iterator<float> iter_z(*msg_pointcloud, "z");
+    sensor_msgs::PointCloud2Iterator<float> iter_label(*msg_pointcloud, "label");
+    const rs2::vertex* vertex = pc.get_vertices();
+    const uint8_t* label = pc.get_labels();
+    msg_pointcloud->width = 320;
```
SamerKhshiboun, line 838: @remibettan 
should we add get_width / get_height for rs2::labeleld_points? 
because in the future, labeled point cloud might support different heights/widths/resolutions

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -785,6 +803,59 @@ void BaseRealSenseNode::publishPointCloud(rs2::points pc, const rclcpp::Time& t,
...
+    {
+        *iter_x = vertex->x;
+        *iter_y = vertex->y;
+        *iter_z = vertex->z;
+        *iter_label = *label;
+        ++iter_x; ++iter_y; ++iter_z; ++iter_label;
+    }
+
+    msg_pointcloud->header.stamp = t;
+    msg_pointcloud->header.frame_id = "camera_link";
```
SamerKhshiboun, line 853: @benlev , I'm not sure this is the right name.
but other names might corrupt the TF tree. let me know what do you think here

realsense2_camera/include/base_realsense_node.h
```diff
@@ -271,7 +273,7 @@ namespace realsense2_camera
 
         bool _use_intra_process;      
         std::map<stream_index_pair, std::shared_ptr<image_publisher>> _image_publishers;
-        
+        rclcpp::Publisher<sensor_msgs::msg::PointCloud2>::SharedPtr _labeled_pointcloud_publisher;
```
remibettan, line 276: Can you explain why you have a specific publisher for lpc and not for pc?

realsense2_camera/src/base_realsense_node.cpp
remibettan: The API size() should be used instead of width and height

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -861,50 +932,56 @@ void BaseRealSenseNode::publishFrame(rs2::frame f, const rclcpp::Time& t,
         image = fix_depth_scale(image, _depth_scaled_image[stream]);
     }
 
-    if (info_publishers.find(stream) == info_publishers.end() ||
-        image_publishers.find(stream) == image_publishers.end())
+    bool is_sc_stream = isSCStream(stream);
+
+    // if stream is on, and is not a SC stream, publish camereaInfo
```
remibettan: typo in the comment

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -861,50 +932,56 @@ void BaseRealSenseNode::publishFrame(rs2::frame f, const rclcpp::Time& t,
...
+    if(!is_sc_stream && info_publishers.find(stream) != info_publishers.end())
```
remibettan: I would have renamed the is_sc_stream to should_publish_camera_info, since this is the aim of it

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -188,6 +189,11 @@ void BaseRealSenseNode::stopPublishers(const std::vector<stream_profile>& profil
             _info_publisher.erase(sip);
             _depth_aligned_image_publishers.erase(sip);
             _depth_aligned_info_publisher.erase(sip);
+
+            if(_labeled_pointcloud_publisher)
```
remibettan, line 193: In which case would this condition be false?

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -215,7 +221,13 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
         rmw_qos_profile_t qos = sensor.getQOS(sip);
         rmw_qos_profile_t info_qos = sensor.getInfoQOS(sip);
 
-        if (profile.is<rs2::video_stream_profile>())
+        // special handling for labeled point cloud stream
+        if (profile.is<rs2::video_stream_profile>() && profile.stream_type() == RS2_STREAM_LABELED_POINT_CLOUD)
+        {
+            _labeled_pointcloud_publisher = _node.create_publisher<sensor_msgs::msg::PointCloud2>("labeled_point_cloud/points",
+                rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(qos),qos));
```
remibettan: Please explain what is QoS?

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -215,7 +221,13 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
...
+        }
+        else if (profile.is<rs2::video_stream_profile>())
```
remibettan: Since the video_stream_profile happens most of the time, it should be the first condition in the code.
Please pass it back to the if condition, and move the lpc condition to some else if after it (even maybe last one)

remibettan (commented):
Few comments

remibettan (changes requested):
Changes requested

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: but I used size to get the size of the whole PC.
what if I want to get width/height to fill the structure of point cloud message ?
or should I assume this is unordered Point Cloud and set the height 1 and width = frame.size() ?
Reference: http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/PointCloud2.html
**2D structure of the point cloud. If the cloud is unordered, height is 1 and width is the length of the point cloud.**

realsense2_camera/include/base_realsense_node.h
SamerKhshiboun: Because we can turn on both Depth point cloud and Labeled Point cloud at the same time
These are two different publishers

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: if it was not initialized at all (e.g., the camera doesn't support Depth Mapping Sensor)

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: Quality Of Service.

Each publisher should define quality of service, regarding its Durability, Reliability, Depth, History, Queue Size, etc..
Reference: https://docs.ros.org/en/rolling/Concepts/About-Quality-of-Service-Settings.html

Read both parts:
1- https://docs.ros.org/en/rolling/Concepts/About-Quality-of-Service-Settings.html#qos-policies
2- https://docs.ros.org/en/rolling/Concepts/About-Quality-of-Service-Settings.html#qos-profiles

realsense2_camera/src/base_realsense_node.cpp
remibettan: Well, you may this. Because, as for point cloud, there is no meaning to height and width there, since vertices are passed, and not a frame

SamerKhshiboun: I understand we are getting unordered vertices, not an ordered ones.
But, its better to wait for @benlev inputs here, since he might want this meta info in the msg itself.

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: fixed

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: changed the function name from isSCStream to shouldPublishCameraInfo, and changed the implementation inside it.

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: Changed the flow.
Combined it to one  "if (profile.is<rs2::video_stream_profile>()) "
then asked about Labeled PC stream inside it.

Nir-Az:
1 side comment, we did some work with the community on supporting zero copy for PC, please make sure you are aligning L-PC to that.
Reference: https://github.com/IntelRealSense/realsense-ros/pull/2599

SamerKhshiboun:
> 1 side comment, we did some work with the community on supporting zero copy for PC, please make sure you are aligning L-PC to that. Reference: #2599

Already aligned in the current implementation

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -520,6 +522,12 @@ void BaseRealSenseNode::frame_callback(rs2::frame frame)
             if (f.is<rs2::video_frame>())
                 ROS_DEBUG_STREAM("frame: " << f.as<rs2::video_frame>().get_width() << " x " << f.as<rs2::video_frame>().get_height());
 
+            if (f.is<rs2::labeled_points>())
```
remibettan, line 525: better change this if statement to else if - do you agree?

remibettan (approved):
One last comment

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: done

realsense2_camera/src/base_realsense_node.cpp
Nir-Az: This was a valid question :)
LPC resolution is now changed and I am phasing the same question..

Issue #2595, opened 2023-01-12
setting align_depth causes misaligned coloured point cloud


christian-rauch:
When I set the parameter `align_depth.enable` to `True`, the coloured point cloud will have misaligned colour, i.e. the colour will not be correctly registered to the points.

Consider the following ROS2 launch file:
```yaml
launch:
- node_container:
    pkg: rclcpp_components
    exec: component_container
    name: camera_container
    namespace: realsense
    composable_node:
    -   pkg: realsense2_camera
        plugin: realsense2_camera::RealSenseNodeFactory
        name: camera
        param:
        -   name: rgb_camera.profile
            value: 640x480x30
        -   name: "align_depth.enable"
            value: True
        -   name: pointcloud.enable
            value: True
        -   name: pointcloud.stream_filter
            value: 2 # RS2_STREAM_COLOR
```
The launch file will start the node, set VGA resolution @ 30 fps (`rgb_camera.profile`), align the colour with the depth (`align_depth.enable`) and publish a coloured point cloud (`pointcloud`).

With
```yaml
        -   name: "align_depth.enable"
            value: True
```
the point cloud is shown in RViz2 with misaligned colour: [rs_not_aligned.zip](https://github.com/IntelRealSense/realsense-ros/files/10402071/rs_not_aligned.zip). The image shows the point cloud of a RealSense box. It is clearly visible that the white background is registered to the majority of the box points in the foreground.

When set to `False` or removed, the colour is correctly aligned with the points: [rs_aligned.zip](https://github.com/IntelRealSense/realsense-ros/files/10402079/rs_aligned.zip)

It looks like the colour is registered twice before it is assigned to the points.

MartyG-RealSense:
Hi @christian-rauch  If the pointcloud filter is enabled then alignment should automatically be used without having to set align_depth to true.

christian-rauch:
> Hi @christian-rauch If the pointcloud filter is enabled then alignment should automatically be used without having to set align_depth to true.

I am sure that in the default configuration, the node will align depth and colour. If I don't use `align_depth`, then the alignment in the point cloud is correct. But I am additionally setting `align_depth` because I also need the aligned depth images on topic `/aligned_depth_to_color/image_raw`. Enabling both somehow registers the images twice as if the point cloud registration would use the already registered image from `/aligned_depth_to_color/image_raw` instead of the original depth image.

[...]

PR #2775, opened 2023-06-12
Applying AlignDepth filter after Pointcloud

Arun-Prasad-V:
Fixing #2595. 

**Issue**:
- When both PointCloud and AlignDepth filters are enabled, the point cloud output has misaligned colored point cloud

**Root cause**:
- First Align Depth filter process the frame and then Point Cloud filter uses the same processed frame and internally align again
- That's why color is getting registered twice

**Proposed Solution**:
- Apply the Point Cloud filter first
    - So, it will get the unprocessed frame and internally do the alignment
    - The alignment will only happen on point cloud's data and the original depth frame will be untouched 
- Then, apply the Align Depth filter
    - It will get the original depth frame and does the alignment

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -195,14 +195,15 @@ void BaseRealSenseNode::setupFilters()
         }
         _cv_mpc.notify_one();
     };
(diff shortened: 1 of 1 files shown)


SamerKhshiboun (approved):
LGTM.
@Nir-Az should we wait for user confirmation? or should we merge it ?

Nir-Az:
> LGTM. @Nir-Az should we wait for user confirmation? or should we merge it ?

Since the last comment was 2 weeks ago and it was tested by us then we will merge and monitor the issue for user feedback :)

PR #2787, opened 2023-06-21
Support Iron distro

SamerKhshiboun:
Tracked by RSDEV-353

* Iron to be added to pre-release when we release librealsense package in ROS build farm, till then we will get error in the script


.github/workflows/pre-release.yml
```diff
@@ -31,12 +31,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [humble, foxy]
+        ros_distro: [humble]
```
SamerKhshiboun: will add iron here only when it builds and be ready on the ROS2 public package index
meanwhile, if we add it here, it will fail, since it will not find librealsense2 package for iron.

Nir-Az: I say let's add it and ignore Iron failure, it should be fixed once it's merged , right?

SamerKhshiboun: done

Nir-Az (approved):

PR #2793, opened 2023-06-27
Publish /aligned_depth_to_color topic only when color frame present

Arun-Prasad-V:
**Issue:**
- /aligned_depth_to_color topic has unaligned images as well published randomly
- Happens especially when rgb module frame rate is less than depth module frame rate

**Root cause:**
- When a particular frameset have only depth frame and not color frame, alignment will not happen. But still, the unaligned image is getting published 

**Fix:**
- Publish aligned image only when both color and depth frames are present in the frameset


SamerKhshiboun (approved):
LGTM

PR #2802, opened 2023-07-03
add new RGBD topic

SamerKhshiboun:
Tracked in [RSDEV-419]

RGBD new topic, publishing [RGB + Depth] in the same message.
For now, works only with depth aligned to color images.

These are the minimum required parameters when launching the node:

**enable_rgbd**: to enable/disable rgbd topic, changeable during runtime
**align_depth.enable**: in order to work with depth aligned images
**enable_sync**: in order to let librealsense sync between frames, and get the frameset with color and depth images combined
And, you should enable both color and depth sensors to get that, and this is done by **enable_color** and **enable_depth** paramteres.

a run example:
`ros2 launch realsense2_camera rs_launch.py enable_rgbd:=true enable_sync:=true align_depth.enable:=true enable_color:=true enable_depth:=true
`
The current QoS of the topic itself, is the same as Depth and Color streams (SYSTEM_DEFAULT)

The message structure:

```
RGBD Message
std_msgs/Header header
sensor_msgs/Image rgb
sensor_msgs/CameraInfo rgb_camera_info
sensor_msgs/Image depth
sensor_msgs/CameraInfo depth_camera_info

```

Although for now the **depth_camera_info** and the **rgb_camera_info** looks the same (because of the alignment), I chose to keep both camera_infos since we can support also un-aligned depth images in the future.


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -886,6 +893,51 @@ IMUInfo BaseRealSenseNode::getImuInfo(const rs2::stream_profile& profile)
...
+cv::Mat& BaseRealSenseNode::getCVMatImage(
+    rs2::frame& frame,
+    std::map<stream_index_pair, cv::Mat>& images,
+    unsigned int width,
+    unsigned int height,
+    unsigned int bpp,
+    const stream_index_pair& stream)
+{
+    auto& image = images[stream];
+    if (image.size() != cv::Size(width, height) || image.depth() != _image_formats[bpp])
```
Nir-Az, line 926: `image.depth()` return the format?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -886,6 +893,51 @@ IMUInfo BaseRealSenseNode::getImuInfo(const rs2::stream_profile& profile)
...
+    {
+        image.create(height, width, _image_formats[bpp]);
+    }
+
+    image.data = (uint8_t*)frame.get_data();
+
+    if (frame.is<rs2::depth_frame>())
+    {
+        image = fix_depth_scale(image, _depth_scaled_image[stream]);
```
Nir-Az, line 935: What is the operation here?
Does it copy?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -907,66 +959,61 @@ void BaseRealSenseNode::publishFrame(
...
-        image.create(height, width, _image_formats[bpp]);
-    }
-    image.data = (uint8_t*)f.get_data();
+        auto &image_publisher = image_publishers.at(stream);
+        cv::Mat image_cv_matrix;
+        if (_rgbd_publisher && 0 != _rgbd_publisher->get_subscription_count())
+        {
+            image_cv_matrix = getCVMatImage(f, images, width, height, bpp, stream);
+        }
+        if (0 != image_publisher->get_subscription_count())
```
Nir-Az, line 976: Missing some comments here.
Can we separate `rgbd` to a dedicated function?

`image_publisher` is depth image publisher?

Id we asked for rgbd we allocate the first, if we ask only depth we allocate the second?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -907,66 +959,61 @@ void BaseRealSenseNode::publishFrame(
...
+        {
+            if(image_cv_matrix.empty())
+            {
+                image_cv_matrix = getCVMatImage(f, images, width, height, bpp, stream);
+            }
 
-    if (f.is<rs2::depth_frame>())
-    {
-        image = fix_depth_scale(image, _depth_scaled_image[stream]);
```
Nir-Az, line 920: No need to fix the scale here?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -907,66 +959,61 @@ void BaseRealSenseNode::publishFrame(
...
+            }
+            
+            // Transfer the unique pointer ownership to the RMW
+            sensor_msgs::msg::Image *msg_address = img_msg_ptr.get();
+            image_publisher->publish(std::move(img_msg_ptr));
 
+            ROS_DEBUG_STREAM(rs2_stream_to_string(f.get_profile().stream_type()) << " stream published, message address: " << std::hex << msg_address);
+        }
+    }
     // if stream is on, and is not a SC stream, publish cameraInfo
```
Nir-Az, line 1000: This code needs dome refactoring IMO..
can we try splitting the camera info publish to a function?
Something like
```
if (should_publish_camera_info())
{
    publish_camera_info(...);
}
```

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -907,66 +959,61 @@ void BaseRealSenseNode::publishFrame(
...
     if(shouldPublishCameraInfo(stream) && info_publishers.find(stream) != info_publishers.end())
     {
         auto& info_publisher = info_publishers.at(stream);
-        if(0 != info_publisher->get_subscription_count())
+        if (_rgbd_publisher && 0 != _rgbd_publisher->get_subscription_count())
         {
             auto& cam_info = _camera_info.at(stream);
             if (cam_info.width != width)
```
Nir-Az, line 1014: Why do we test width? and if so why width only?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -907,66 +959,61 @@ void BaseRealSenseNode::publishFrame(
...
             {
                 updateStreamCalibData(f.get_profile().as<rs2::video_stream_profile>());
             }
             cam_info.header.stamp = t;
```
Nir-Az, line 1018: We set a field in `cam_info` here, and override it on line 1010.
Why is that?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -975,6 +1022,37 @@ void BaseRealSenseNode::publishFrame(
...
 
+
+void BaseRealSenseNode::publishRGBD(cv::Mat& rgb_cv_matrix, cv::Mat& depth_cv_matrix, const rclcpp::Time& t)
+{
+    if (_rgbd_publisher && 0 != _rgbd_publisher->get_subscription_count())
+    {
+        ROS_DEBUG_STREAM("Publishing RGBD message");
+        unsigned int rgb_width = rgb_cv_matrix.size().width;
+        unsigned int rgb_height = rgb_cv_matrix.size().height;
+        unsigned int rgb_bpp = rgb_cv_matrix.elemSize() ? rgb_cv_matrix.elemSize() : 3;
```
Nir-Az: `rgb_cv_matrix.elemSize()` can be 0?

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -223,6 +223,9 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
 
         if (profile.is<rs2::video_stream_profile>())
         {
+            _is_color_enabled |= profile.stream_type() == RS2_STREAM_COLOR;
```
Nir-Az: Those parameters are bool right?
Better use logical operators over bitwise operators IMO

realsense2_camera_msgs/CMakeLists.txt
```diff
@@ -28,16 +28,18 @@ find_package(ament_cmake REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 find_package(builtin_interfaces REQUIRED)
 find_package(std_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
```
Nir-Az, line 31: Why do we need to add this?

realsense2_camera_msgs/msg/RGBD.msg
```diff
@@ -0,0 +1,6 @@
+# RGBD Message
+std_msgs/Header header
+sensor_msgs/Image rgb
```
Nir-Az: Don't you prefer: 
```
sensor_msgs/CameraInfo rgb_camera_info
sensor_msgs/Image rgb
sensor_msgs/CameraInfo depth_camera_info
sensor_msgs/Image depth
```

First metadata then data?

realsense2_camera_msgs/package.xml
```diff
@@ -16,10 +16,12 @@
   <buildtool_depend>rosidl_default_generators</buildtool_depend>
   <build_depend>builtin_interfaces</build_depend>
   <build_depend>std_msgs</build_depend>
+  <build_depend>sensor_msgs</build_depend>
```
Nir-Az, line 19: But rgbd is a custom message, right?
I think the camera depend on camera_msgs
and it may have this depends?

realsense2_camera_msgs/msg/RGBD.msg
SamerKhshiboun: what is the point of the order you chose ? anyway the user will fetch each field by it's id

realsense2_camera_msgs/CMakeLists.txt
SamerKhshiboun: the colcon build fails for this package without this line
since we are using sensor_msgs/Image and sensor_msg/CameraInfo types inside the RGBD.msg

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: Almost.
It was before, just moved the location of this line
Reference:
https://docs.opencv.org/3.4/d3/d63/classcv_1_1Mat.html#a8da9f853b6f3a29d738572fd1ffc44c0

realsense2_camera_msgs/msg/RGBD.msg
Nir-Az: From the concept of first header / metadata then data.

Lets say you need to debug and view the raw data, what will you prefer? fir to see 10000000 bytes of image and then ~50 bytes of timetag / id and stuff like that?

I would even change it to 
```
sensor_msgs/CameraInfo rgb_camera_info
sensor_msgs/CameraInfo depth_camera_info
sensor_msgs/Image rgb
sensor_msgs/Image depth
```

But it's a manner of style, your call.

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: image_publisher is used for depth and for rgb publishers
image is used to store all matrix of depth and rgb
that what was before.
I tried to reorder the code with minimal changes, creating small functions, to reuse this in different places, specially for the RGBD.
in order to make sure the RGBD can be published, you need to make sure that both images of depth and color exist.
they can be created in two ways:
- if there are subscribers for the depth and for the color, then the two images are for sure inside the images map.
- if there are no subscribers for these topics, but the user enabled both sensors, we create these images and store them.
We can sync on this over teams.

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: Fixed
It was only for debugging.

realsense2_camera_msgs/package.xml
SamerKhshiboun: this is the camera_msgs package.xml, so it should be here.
there was another duplication inside "realsense2_camera" package.xml that I removed

Nir-Az: Got it.

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: this fix_depth_scale just converts the depth pixels to a specific unit (mm)
only when creating a matrix for the first time (see the if inside the function) we adjust this matrix cells
then, every time we use this matrix to fill images, we are using the same scaled image

realsense2_camera_msgs/CMakeLists.txt
Nir-Az: OK

realsense2_camera/src/base_realsense_node.cpp
Nir-Az: Thanks

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -111,6 +111,9 @@ BaseRealSenseNode::BaseRealSenseNode(rclcpp::Node& node,
     _is_initialized_time_base(false),
     _camera_time_base(0),
     _sync_frames(SYNC_FRAMES),
+    _enable_rgbd(ENABLE_RGBD),
```
Gilaadb, line 114: I didn't see where you use the boolean value given by the user and replaces this constant, but I assume it exists somewhere. please show me.

realsense2_camera/include/base_realsense_node.h
```diff
@@ -208,14 +210,34 @@ namespace realsense2_camera
...
+        void publishFrame(
+            rs2::frame f,
             const rclcpp::Time& t,
             const stream_index_pair& stream,
             std::map<stream_index_pair, cv::Mat>& images,
             const std::map<stream_index_pair, rclcpp::Publisher<sensor_msgs::msg::CameraInfo>::SharedPtr>& info_publishers,
             const std::map<stream_index_pair, std::shared_ptr<image_publisher>>& image_publishers,
             const bool is_publishMetadata = true);
 
+        void publishRGBD(cv::Mat& rgb_cv_matrix, cv::Mat& depth_cv_matrix, const rclcpp::Time& t);
```
Gilaadb: Can you declare the cv::Mats as consts?

realsense2_camera/include/base_realsense_node.h
```diff
@@ -238,6 +260,7 @@ namespace realsense2_camera
         void updateSensors();
         void publishServices();
         void startPublishers(const std::vector<rs2::stream_profile>& profiles, const RosSensor& sensor);
+        void startRGBDPublisherIfNeeded();
```
Gilaadb, line 263: why is this publisher different and has its own function?

realsense2_camera/src/base_realsense_node.cpp
Gilaadb: but doesn't bpp stand for bits per pixel? To me it looks confusing that it means an index to array of image formats.

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -324,6 +329,27 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
...
+        {
+            rmw_qos_profile_t qos = rmw_qos_profile_system_default;
+
+            rclcpp::PublisherOptionsWithAllocator<std::allocator<void>> options;
+            options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
+            _rgbd_publisher = _node.create_publisher<realsense2_camera_msgs::msg::RGBD>("rgbd",
+                rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(qos), qos), std::move(options));
+        }
+        else {
+            ROS_ERROR("In order to get rgbd topican enabled, "\
```
Gilaadb: topican? :)

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: https://github.com/IntelRealSense/realsense-ros/blob/0f2aa7c1b68bba626d9e1fcf8854238f797c5047/realsense2_camera/src/parameters.cpp#L49

realsense2_camera/include/base_realsense_node.h
SamerKhshiboun: Because each publisher was related to sensor (at least in the current architecture)
This publisher is unique and depends on two sensors (and on their ON/OFF status)
Redesigning the current logic is something we want to do, but not now, not for this current high priority requirements.

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: fixed

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: you are commenting on the old file

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: I prefer to do this refactoring later, with a bigger redesign to all topics/publishers architecture, after finishing this high priority feature.

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: Fixed

realsense2_camera_msgs/msg/RGBD.msg
SamerKhshiboun: Updated.

realsense2_camera/include/base_realsense_node.h
SamerKhshiboun: Done

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: line 999, if there any subscribers to our RGBD topic, we should make sure that cam_info msg is ready, but still not publishing it, since we will do that when the whole message RGBD message is ready (two images, and two cam infos)
*notice, that there might be no subscribers to the depth and color topics, although the user enabled these streams.

in line 1010, we are sure that there are subscribers to info topics of depth/color camera info, so we are publishing the cam info there.
here, we can take the cam_info that was already calculated above (if there any subscriber to RGBD) or calculate it from scratch.

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: This was before, I don't know exactly why this was the criteria for checking if we need to fix the camera_info calib data or not.
Maybe who implemented this chose this field (from a big Intrinsic struct) as the minimal check to compare against, so if the width is not equal, it means that the camera info should be changed.

Actually, if you test this, we get inside the if only in the first time, when the cam_info.width is 0 and need to be updated.
Since that, the camera info of each stream is not being updated, and we keep fetching it from the std::map of _camera_info

realsense2_camera/src/base_realsense_node.cpp
Nir-Az: Yes, I saw it was removed so I asked, But I see it in another place now I will look closer..

realsense2_camera/src/base_realsense_node.cpp
Nir-Az: NP

realsense2_camera/src/base_realsense_node.cpp
Nir-Az: I think it's worth to add some comment on lines that there is some assumption for the flow.

Otherwise we can easily change it in the future without knowing that the one who wrote it took in mind some special scenarios.

Currently its hard to understand the flow, so if we postpone the refactor, lets add comments..

SamerKhshiboun: Sure, will update this.

SamerKhshiboun: Updated.

README.md
```diff
@@ -407,6 +408,28 @@ Enabling stream adds matching topics. For instance, enabling the gyro and accel
 
 <hr>
 
+<h3 id="rgbd">
+  RGBD Topic
+</h3>
+
+RGBD new topic, publishing [RGB + Depth] in the same message (see RGBD.msg for reference). For now, works only with depth aligned to color images.
```
Nir-Az: I think it worth mentioning that the RGB + Depth are synchronized by frame time tag, right?

SamerKhshiboun: done

Nir-Az (approved):
Great!

PR #2815, opened 2023-07-18
fix labeled point cloud publisher reset condition

SamerKhshiboun:
Tracked by [RSDEV-423]


SamerKhshiboun:
@Nir-Az , the fix was validated by validation team.

Nir-Az (approved):

PR #2817, opened 2023-07-19
Replaced Deprecated function mktemp to TemporaryFile

karina-ranadive:
The mktemp function in launch launch_utils.py presents a security vulnerability for my team that utilizes realsense-ros. To fix this, I utilized the function TemporaryFile which has the same purpose as mktemp but is a safe function to use whereas mktemp is a call to a deprecated function, which can be insecure.


SamerKhshiboun:
Hi @karina-ranadive 
Thanks for the fix.

I'm adding a reference link describing the differences 
https://docs.python.org/3/library/tempfile.html

Approving the PR.

SamerKhshiboun (approved):

karina-ranadive:
Awesome, thank you so much!

PR #2822, opened 2023-07-25
Updated rs_launch configurations

Arun-Prasad-V:
**Issue:**
- In `rs_multi_camera_launch.py`, '1' and '2' are appended as suffix of actual param names listed in `rs_launch.py`. For example: `enable_infra` param name from `rs_launch.py` becomes `enable_infra1` and `enable_infra2`. But, there are already params with those names in `rs_launch.py`. This leads to misconfiguration of params.

**Root cause:**
- Misconfiguration happens because while running `rs_multi_camera_launch.py` both actual and appended param names are getting registered
- No need to register actual param names from `rs_launch.py`
- Should register only appended param names

**Fix:**
- Updating `rs_launch.py` to support getting only required param names dynamically and register them

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -85,43 +85,42 @@ def yaml_to_dict(path_to_yaml):
     with open(path_to_yaml, "r") as f:
         return yaml.load(f, Loader=yaml.SafeLoader)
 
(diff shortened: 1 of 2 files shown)


SamerKhshiboun (approved):
LGTM, thanks for the fix Arun !

PR #2836, opened 2023-08-03
Reduce changes between ros2-hkr and ros2-development - stage 2

SamerKhshiboun:
Merge ros2-development into ros2-hkr:
- support formats
- fix RGBD intra-process


Nir-Az (approved):
CI pass offline

PR #2842, opened 2023-08-10
Get camera_name from YAML if provided

Arun-Prasad-V:


realsense2_camera/launch/rs_launch.py
```diff
@@ -30,7 +30,6 @@
                            {'name': 'unite_imu_method',             'default': "0", 'description': '[0-None, 1-copy, 2-linear_interpolation]'},
                            {'name': 'json_file_path',               'default': "''", 'description': 'allows advanced configuration'},
                            {'name': 'log_level',                    'default': 'info', 'description': 'debug log level [DEBUG|INFO|WARN|ERROR|FATAL]'},
-                           {'name': 'output',                       'default': 'screen', 'description': 'pipe node output [screen|log]'},
```
Arun-Prasad-V, line 33: 'Output' param is not used. So, deleted.

realsense2_camera/launch/rs_launch.py
```diff
@@ -95,35 +94,35 @@ def yaml_to_dict(path_to_yaml):
...
-                node_executable='realsense2_camera_node',
-                prefix=['stdbuf -o L'],
-                parameters=[params
-                            , params_from_file
-                            ],
-                output='screen',
-                arguments=['--ros-args', '--log-level', LaunchConfiguration('log_level' + param_name_suffix)],
+                package         = 'realsense2_camera',
+                node_namespace  = params['camera_namespace' + param_name_suffix],
+                node_name       = params['camera_name' + param_name_suffix],
```
Arun-Prasad-V, line 107: Camera name and namespace values got from LaunchConfiguration are already stored in 'params'. Also, those values will get overwritten if they are defined in YAML config file.

realsense2_camera/launch/rs_launch.py
```diff
@@ -95,35 +94,35 @@ def yaml_to_dict(path_to_yaml):
...
             ]
     else:
         return [
             launch_ros.actions.Node(
-                package='realsense2_camera',
-                namespace=LaunchConfiguration('camera_namespace' + param_name_suffix),
-                name=LaunchConfiguration('camera_name' + param_name_suffix),
-                executable='realsense2_camera_node',
-                parameters=[params
-                            , params_from_file
```
Arun-Prasad-V, line 122: 'params' is already overwritten by 'params_from_file' in line 99.

PR #2845, opened 2023-08-13
Add trigger-private-ci.yaml

timor-raiman:


.github/workflows/trigger-private-ci.yaml
```diff
@@ -0,0 +1,45 @@
+name: Trigger Private CI
+on:
+  workflow_dispatch:
+  issue_comment:
+    types: [created, edited]
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+    branches: [ros2-development]
```
SamerKhshiboun: replace this to "ros2-hkr"

timor-raiman: do you mean you want different ci processes for different branches, or do you want 2 kinds of ci to work for every PR?

SamerKhshiboun: for ros2-development, keep the same behavior (3 scripts inside .workflow - main_ci, cppchecker, prerelase)
for ros2-hkr, we will have cppcheck and your new script.

PR #2849, opened 2023-08-18
Create /imu topic only when motion streams enabled

Arun-Prasad-V:
Tracked by RSDEV-550


realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -305,6 +307,11 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
                 rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(extrinsics_qos), extrinsics_qos), std::move(options));
         }
     }
+    if (_is_accel_enabled && _is_gyro_enabled && (_imu_sync_method > imu_sync_method::NONE))
+    {
+        _synced_imu_publisher = std::make_shared<SyncedImuPublisher>(_node.create_publisher<sensor_msgs::msg::Imu>("~/imu", 5));
```
SamerKhshiboun: I know this was before, but do you think 5 is fair enough ? or should we use here QoS in the same format as we used in other topics ?

SamerKhshiboun:
I think we need to edit this part in the Readme
1 - add "-" to the first mark
2- replaces ? I don't think it replace, I think it is an addition, or am I wrong ?

image

SamerKhshiboun (approved):

PR #2861, opened 2023-08-29
fix readme and nodefactory for ros2 run

SamerKhshiboun:
Tracked in RSDEV-472

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@
   * [Installation](#installation)
   * [Usage](#usage)
      * [Starting the camera node](#start-camera-node)
+     * [Camera name and namespace](#camera-name-and-namespace)
      * [Parameters](#parameters)
      * [ROS2-vs-Optical Coordination Systems](#coordination)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -183,6 +184,74 @@
 
 <hr>
 
+<h3 id="camera-name-and-namespace">
```
Nir-Az, line 187: Please remember we will have to notify this change.
To validation to get ready for it, and other internal users as well
What will happen to users that doesn't change there launch file?
Maybe we need a red notice here about this change?
Let's speak and see how we can make a smooth transition as possible.

SamerKhshiboun: I can use these kinds of notifications in the top of the README page
https://github.com/orgs/community/discussions/16925
We can tell validation and internal users about this before we merge this into SC branches.

Nir-Az: Looks like a good option..

Nir-Az (approved):

Issue #2810, opened 2023-07-12
Misaligned rgbd point cloud when align_depth is enabled


hello-chintan:
Hey team,

I am using a D435i camera on my robot to accomplish various tasks. I am using source built up to date version of realsenser-ros drivers along with ROS 2 Galactic and Humble. Upon enabling the align_depth parameter while launching the camera driver in ROS 2 and visualizing in RViz, I am observing a peculiar behavior which results in the /camera/depth/color/points topic to have an offset with respect to the actual object. This behavior is not observed when the align_depth parameter is disabled.

I am attaching images to make things more clear:
![align_depth_enabled](https://github.com/IntelRealSense/realsense-ros/assets/97639181/ff4250d0-1444-403c-9992-d0c6b69dd974)
Figure 1: Point cloud offset with respect to robot model when align_depth parameter is enabled

![align_depth_disabled](https://github.com/IntelRealSense/realsense-ros/assets/97639181/448a3a2e-af19-471d-903a-6ca23541434a)
Figure 2: No offset between point cloud and robot model when align_depth parameter is disabled

I also tested this to compare the behavior in ROS Noetic, but a similar effect was not observed when enabling the align_depth parameter. My issue seems to be similar in nature to https://github.com/IntelRealSense/realsense-ros/issues/2595. I'd appreciate help to resolve this.

Thank you,
Chintan Desai

MartyG-RealSense:
Hi @hello-chintan  Yes, it does sound similar to the ROS2 case at https://github.com/IntelRealSense/realsense-ros/issues/2595 which had an internal report created by Intel about it so that it could be investigated.

Until that issue is resolved, I believe that the only course of action that you can take is to set `align_depth.enable` to false like you are already doing.

Arun-Prasad-V:
Hi @hello-chintan ,

A pull request https://github.com/IntelRealSense/realsense-ros/pull/2775 related to this issue has been merged recently.

Could you please check and confirm you have the latest RealSense ROS2 source code and observing this issue with that?

hello-chintan:
Hey @Arun-Prasad-V, I can confirm that I am using the latest realsense-ros drivers and still seeing the misaligned point cloud issue.

Arun-Prasad-V:
Hi @hello-chintan, 
Thanks for the confirmation. Could you please share the command and the input params you are using?

hello-chintan:
Sure, @Arun-Prasad-V.

[...]

PR #2868, opened 2023-09-06
Fix: Pointcloud topic's frame_id

Arun-Prasad-V:
Fixing: Issue [#2810](https://github.com/IntelRealSense/realsense-ros/issues/2810)

As per the previously merged PR [#2775](https://github.com/IntelRealSense/realsense-ros/pull/2775), Align_depth filter will be applied after applying pointcloud filter. Thus, the dependency of align_depth filter on pointcloud filter had been removed in that PR. But, missed to update the frame_id of pointcloud topic accordingly. 

So, fixing it here.

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -814,7 +814,7 @@ void BaseRealSenseNode::SetBaseStream()
 
 void BaseRealSenseNode::publishPointCloud(rs2::points pc, const rclcpp::Time& t, const rs2::frameset& frameset)
 {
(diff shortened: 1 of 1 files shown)


SamerKhshiboun:
hi @Arun-Prasad-V, and thanks for the fix.
did you get any chance to reproduce the issue and to test that this fix is really solves the problem ?

Arun-Prasad-V:
Hi @SamerKhshiboun ,
I was able to reproduce the misalignment issue by following the steps mentioned in https://github.com/IntelRealSense/realsense-ros/issues/2810#issuecomment-1706559042. 

With this fix, there is no misalignment.

SamerKhshiboun (approved):

PR #2874, opened 2023-09-13
add t265_odom_fps parameter

CatInTheRain:
Hi all ! The parameter `pose_fps` does not work so I created my own parameter `t265_odom_fps`, in a primitive way it works for some values because the modulo operator. Hope to be to inspiration for the developers


SamerKhshiboun:
Hi @CatInTheRain ,
The T265 product is discontinued since of February 2022. [Reference](https://www.intelrealsense.com/message-to-customers/?_ga=2.266607757.2059295677.1675047606-1846334074.1674552675)
Thanks for your contributing for the RealSense ROS Wrapper. Still, users can use this PR as reference if needed.
Closing this PR.

PR #2914, opened 2023-10-26
Don't enforce C++11 standard

rhaschke:
The default in Ubuntu 18.04 and 20.04 is C++14.
Ubuntu 22.04 uses C++17 as the gcc default, which is required by log4cxx there.

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -1,6 +1,5 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(realsense2_camera)
-add_compile_options(-std=c++11)
 
 option(BUILD_WITH_OPENMP "Use OpenMP" OFF)
 option(SET_USER_BREAK_AT_STARTUP "Set user wait point in startup (for debug)" OFF)
@@ -62,7 +61,7 @@ endif()
 
 if (WIN32)
 else()
-set(CMAKE_CXX_FLAGS "-fPIE -fPIC -std=c++11 -D_FORTIFY_SOURCE=2 -fstack-protector -Wformat -Wformat-security -Wall ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "-fPIE -fPIC -D_FORTIFY_SOURCE=2 -fstack-protector -Wformat -Wformat-security -Wall ${CMAKE_CXX_FLAGS}")
 endif()
 
 add_message_files(


realsense2_camera/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(realsense2_camera)
-add_compile_options(-std=c++11)
```
Nir-Az, line 3: Perhaps we need C++14 instead of C++17 as this is the minimal C++ version we require?

Nir-Az: We prefer the compilation to fail if C++17 or higher code is added

rhaschke: Then this won't compile on Ubuntu 22.04.

Nir-Az: But if c++11 compile on U22, why would C++14 won’t?

Nir-Az: Oh sorry just noticed that this is for the ros1 branch which we do not support/maintain anymore..

rhaschke: Yes, this is for the ROS1 branch. If you don't actively support that anymore, be happy about community contributions like this ;-)

Nir-Az: Thank you for that :)
Can you please explain what is fixed by this PR?
I can say that LibRS SDK require C++14 so I can understand why changing the ROS wrapper to C++14 is reasonable.
How removing any C++ specifier is better than require C++14?
We build LibRS on Ubuntu 22 and also it's tools which require C++11 without any issues.

rhaschke: This is not libRS, but `realsense-ros`. To build that for ROS1 on Ubuntu 22 (which you probably never tried), we need to enable C++17 support, which is required by log4cxx 0.12 shipped with Ubuntu 22. Explicitly requiring a specific standard, fixes C++ to this standard even if the system-default is newer. Specifically, fixing to C++11 or C++14, disables the Ubuntu22-system default of C++17, which is required for log4cxx, which in turn is pulled in by rosconsole.

Nir-Az: Since the CI failed as well, we will have to dive into this.
If you can take a look it will be great.
Thanks

rhaschke: The CI failures are not related to my change, I think. Looks like the installed `em` python-module (found in /usr/local/lib) is not compatible with ROS' message generation:
`AttributeError: module 'em' has no attribute 'RAW_OPT'`

On a native ROS system, `em` comes from /usr/lib/python3/dist-packages/em.py. 
The wrong version is installed by the `ros-tooling` action via pip:
`Invoking "sudo pip3 install --upgrade ... empy  ..."`

Nir-Az: I updated the GitHub actions and now it should work, can you please rebase?

rhaschke: > I updated the GitHub actions and now it should work, can you please rebase?

Done.

Nir-Az:
@rhaschke I appreciate your help to promote and maintain our ROS repo.
Even that we do not use to merge unsupported branched PRs I merged your PR as it shown a small risk and will help you and the community as well, and we always glad to have community contributions.
Wish you a happy new year and keep up the good work :)

PR #2915, opened 2023-10-26
Fix device ID of D405

rhaschke:
The camera is not found with the original ID. Please check!

### DIFF
--- a/realsense2_camera/include/realsense2_camera/constants.h
+++ b/realsense2_camera/include/realsense2_camera/constants.h
@@ -34,7 +34,7 @@ namespace realsense2_camera
     const uint16_t RS435i_RGB_PID   = 0x0B3A; // AWGC_MM
     const uint16_t RS465_PID        = 0x0b4d; // D465
     const uint16_t RS416_RGB_PID    = 0x0B52; // F416 RGB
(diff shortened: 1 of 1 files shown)


Nir-Az:
Hi @rhaschke ,
We currently do not support ROS1 and only focus on ROS2 development.

Looks like the actual D405 PID was never introduced to ROS1 code.
Since this camera is a special case (color stream from depth module), I am not sure this camera will work without code changes and we cannot support this currently.

If you still need it, I suggest forking this branch and verifying this change works for you.

Thanks

rhaschke:
Of course, I verified that the camera works with this fix - both RGB and depth streams.
Would be great if you consider this minor fix to also make the camera available to other users.
Thanks

Nir-Az:
> Of course, I verified that the camera works with this fix - both RGB and depth streams. Would be great if you consider this minor fix to also make the camera available to other users. Thanks

No problem since you verified it we will help and merge it.
FYI, no future release of ROS1 packages is planned.
Thanks for the PR

rhaschke:
Thanks for merging. 

> FYI, no future release of ROS1 packages is planned.

That's fine. I am co-maintaining the [ROS One](https://github.com/ros-o) release, thus we release ourselves.

PR #2934, opened 2023-11-21
Disabling hdr while updating exposure & gain values

Arun-Prasad-V:
Tracked on LRS-962

**Overview**:
- On FW versions 5.13.0.50 and below, the FW allowed to set the preset values (like exposure, gain, etc) when HDR is enabled.
- But in later FW versions 5.14.x.x and above, updating preset values are restricted when HDR is enabled.
- This FW change affected the ROS wrapper's flow for setting the exposure and gain of sequence IDs 1 & 2 (**during launch time**) through below params:
  - depth_module.exposure.1
  - depth_module.gain.1
  - depth_module.exposure.2
  - depth_module.gain.2

This PR makes sure that the HDR is disabled while setting those params during launch time.

During runtime, if the HDR is enabled, the presets cannot be updated. To update them, should disable HDR first and then try.

### DIFF
--- a/realsense2_camera/src/dynamic_params.cpp
+++ b/realsense2_camera/src/dynamic_params.cpp
@@ -25,6 +25,8 @@ namespace realsense2_camera
         _params_backend.add_on_set_parameters_callback(
             [this](const std::vector<rclcpp::Parameter> & parameters) 
                 { 
+                    rcl_interfaces::msg::SetParametersResult result;
+                    result.successful = true;
                     for (const auto & parameter : parameters) 
                     {
                         try
@@ -43,15 +45,15 @@ namespace realsense2_camera
                                 }
                             }
                         }
-                        catch(const std::out_of_range& e)
-                        {}
                         catch(const std::exception& e)
                         {
-                            std::cerr << e.what() << ":" << parameter.get_name() << '\n';
+                            result.successful = false;
+                            result.reason = e.what();
+                            ROS_WARN_STREAM("Set parameter {" << parameter.get_name()
+                                                            << "} failed: " << e.what());
                         }                            
                     }
-                    rcl_interfaces::msg::SetParametersResult result;
-                    result.successful = true;
+
                     return result;
                 });
         monitor_update_functions(); // Start parameters update thread
--- a/realsense2_camera/src/ros_sensor.cpp
+++ b/realsense2_camera/src/ros_sensor.cpp
@@ -102,6 +102,24 @@ void RosSensor::UpdateSequenceIdCallback()
     if (!supports(RS2_OPTION_SEQUENCE_ID))
         return;
 
+    bool is_hdr_enabled = static_cast<bool>(get_option(RS2_OPTION_HDR_ENABLED));
+
+    // Deleter to revert back the RS2_OPTION_HDR_ENABLED value at the end.
+    auto deleter_to_revert_hdr = std::unique_ptr<bool, std::function<void(bool*)>>(&is_hdr_enabled,
+                                                [&](bool* enable_back_hdr) {
+                                                    if (enable_back_hdr && *enable_back_hdr)
+                                                    {
+                                                        set_option(RS2_OPTION_HDR_ENABLED, true);
+                                                    }
+                                                });
+
+    // From FW version 5.14.x.x, if HDR is enabled, updating UVC controls like exposure, gain , etc are restricted.
+    // So, disable it before updating.
+    if (is_hdr_enabled)
+    {
+        set_option(RS2_OPTION_HDR_ENABLED, false);
+    }
+
     int original_seq_id = static_cast<int>(get_option(RS2_OPTION_SEQUENCE_ID));   // To Set back to default.
     std::string module_name = create_graph_resource_name(rs2_to_ros(get_info(RS2_CAMERA_INFO_NAME)));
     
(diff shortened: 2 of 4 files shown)


realsense2_camera/src/ros_sensor.cpp
```diff
@@ -102,6 +102,23 @@ void RosSensor::UpdateSequenceIdCallback()
     if (!supports(RS2_OPTION_SEQUENCE_ID))
         return;
 
+    bool is_hdr_enabled = static_cast<bool>(get_option(RS2_OPTION_HDR_ENABLED));
+
+    // Deleter to revert back the RS2_OPTION_HDR_ENABLED value at the end.
+    auto deleter_to_revert_hdr = std::unique_ptr<bool, std::function<void(void*)>>(nullptr,
```
Nir-Az: What if we pass the `is_hdr_enabled` to the deleter and add a condition inside?

Only turn it on if it was on.
This way we can avoid the reset.

What do you think?

Arun-Prasad-V: @Nir-Az ,
The unique_pointer is initialized with `nullptr`. So, by default, deleter will not be called.
If `is_hdr_enabled` is enabled alone, the unique_pointer is updated (at [line](https://github.com/IntelRealSense/realsense-ros/pull/2934/files#diff-440837bc6c77b72c935dc2c913ead13c628c014cbe0c150de364139a5c8736e3R119)).

Nir-Az: That will not cover all cases.
Let say LibRS throws at the line before
`set_option(RS2_OPTION_HDR_ENABLED, false);`
And it can throw after setting HDR_ENABLED = false
the deleter will run with nullptr.

IMO my suggestion cover that case too.
Thoughts?

Arun-Prasad-V: yeah, I get the point. Will update tomorrow.

realsense2_camera/src/ros_sensor.cpp
```diff
@@ -147,6 +164,11 @@ void RosSensor::UpdateSequenceIdCallback()
         return;
     }
 
+    if (is_hdr_enabled)
```
Arun-Prasad-V: I should remove this part.

realsense2_camera/src/ros_sensor.cpp
```diff
@@ -102,6 +102,24 @@ void RosSensor::UpdateSequenceIdCallback()
     if (!supports(RS2_OPTION_SEQUENCE_ID))
         return;
 
+    bool is_hdr_enabled = static_cast<bool>(get_option(RS2_OPTION_HDR_ENABLED));
+
+    // Deleter to revert back the RS2_OPTION_HDR_ENABLED value at the end.
+    auto deleter_to_revert_hdr = std::unique_ptr<bool, std::function<void(bool*)>>(&is_hdr_enabled,
+                                                [&](bool* enable_back_hdr) {
+                                                    if (*enable_back_hdr)
```
Nir-Az: Since we get a ptr please use
```
if (enable_back_hdr && *enable_back_hdr)
```

Nir-Az:
@Arun-Prasad-V I think we are OK with this fix,

about this
`If the user wants to update any presets dynamically, its users responsibility to disable the HDR first and then update the presets.`
Is it possible to add a ROS log warning if the user try to change it when HDR is enabled?
If it breaks the generic code we can try to count on LibRS warning.
We might want to rephrase it and say it is not supported.
thoughts?

realsense2_camera/src/sensor_params.cpp
```diff
@@ -70,14 +70,7 @@ std::map<std::string, int> get_enum_method(rs2::options sensor, rs2_option optio
...
 { 
-    try
-    {
-        sensor.set_option(option, parameter.get_value<T>());
-    }
-    catch(const std::exception& e)
-    {
-        std::cout << "Failed to set value: " << e.what() << std::endl;
-    }
+    sensor.set_option(option, parameter.get_value<T>());
```
Nir-Az, line 73: Why do we want to remove this try/catch?
Its protecting us from crashing no?

Arun-Prasad-V: There is another try/catch on top of the parameter callbacks in "dynamic_params.cpp". It is anyway getting caught there. Also, by catching there, FAILURE result can be sent to SetParameter request from user.

realsense2_camera/src/dynamic_params.cpp
```diff
@@ -43,15 +45,15 @@ namespace realsense2_camera
                                 }
                             }
                         }
-                        catch(const std::out_of_range& e)
-                        {}
                         catch(const std::exception& e)
                         {
-                            std::cerr << e.what() << ":" << parameter.get_name() << '\n';
+                            result.successful = false;
```
Arun-Prasad-V, line 50: sending FAILURE response for SetParameter service request.

Nir-Az:
So this PR is ready to be merged?
No more “stall” on librs set option exceptions?
I am not sure what fixed it.

realsense2_camera/src/sensor_params.cpp
Nir-Az: OK , got it

Arun-Prasad-V:
> So this PR is ready to be merged? No more “stall” on librs set option exceptions? I am not sure what fixed it.

The stall was happening during INIT, because HDR was not disabled while updating the exp & gain of seq_ids 1&2.
The logic to disable first -> update exp & gain -> enable back fixed the stall.

Nir-Az (approved):

PR #2936, opened 2023-11-22
Add possibility to set all qos profiles in rs launch file

fantalukas:
Hi, we are using this ROS2 package together with realsense D435i camera and first of all we would like to thank you for your work.

**REASON for this PR**:
We are facing the situation that we need to change qos profile of pointcloud (param pointcloud.pointcloud_qos) from DEFAULT to SENSOR_DATA. We are launching your rs_launch.py file from our project and currently is not possible to change qos profile of pointcloud via rs_launch.py file. 

The main reason for this change is that pointcloud topic (more specifically publisher of this topic) uses DEFAULT option of qos profile which uses reliability level: RELIABLE. On the other hand, all our subscribers in our project uses SENSOR_DATA with reliability level: BEST-EFFORT. According to official ros [doc](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-compatibilities) this should not be a problem a it is compatible. BUT we are facing a problem that sometimes for large amount of pointcloud data it could be a problem. Our subscribers sometimes do not see the data at all or stop receiving the data during runtime. By changing reliability level of your pointcloud publisher or our subscribers to the same value (RELIABLE or BEST-EFFORT ) solve the problem.  For our purpose is the best choice to use SENSOR_DATA qos profile with BEST-EFFORT reliability by default. 

**SOLUTION**:
The solution sufficient for us is to add option to change qos profile in rs_launch.py file. Commit in this PR implement our solution in a general way -> enable changing all available qos profiles not just  qos profile of pointcloud (param pointcloud.pointcloud_qos). All available qos params together with their default values were checked via `ros2 param list` command during launching of our realsense camera.

We are using Ubuntu 22.04.3 LTS and ros-rolling

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -81,6 +81,19 @@ configurable_parameters = [{'name': 'camera_name',                  'default': '
                            {'name': 'hdr_merge.enable',             'default': 'false', 'description': 'hdr_merge filter enablement flag'},
                            {'name': 'wait_for_device_timeout',      'default': '-1.', 'description': 'Timeout for waiting for device to connect (Seconds)'},
                            {'name': 'reconnect_timeout',            'default': '6.', 'description': 'Timeout(seconds) between consequtive reconnection attempts'},
+                           {'name': 'accel_info_qos',    'default': 'DEFAULT', 'description': 'set qos profile for accel info'},
+                           {'name': 'accel_qos',    'default': 'SENSOR_DATA', 'description': 'set qos profile for accel'},
+                           {'name': 'color_info_qos',    'default': 'DEFAULT', 'description': 'set qos profile for color info'},
+                           {'name': 'color_qos',    'default': 'SYSTEM_DEFAULT', 'description': 'set qos profile for color'},
+                           {'name': 'depth_info_qos',    'default': 'DEFAULT', 'description': 'set qos profile for depth info'},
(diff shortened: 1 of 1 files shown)


Nir-Az:
Hi @fantalukas .
Firstly, thanks for your PR.
The launch file we include in our source code is a reference launch file that allow basic operations of the RS ROS Wrapper.
The launch file is not a pard of the actual wrapper and it is also not released as part of the ROS release process, users can create their own launch file as there are many use cases on how to activate the ROS node.
Normally we try to keep the basic reference as minimal and basic as possible, so users won't be overwhelmed with data.

We will discuss about this PR internally and update.

Thanks !

Arun-Prasad-V:
Hi @fantalukas,
For this case, we would suggest you to load the params through YAML file. Because in YAML file, you can also add params which are not declared in `rs_launch.py`. Please refer [this](https://github.com/IntelRealSense/realsense-ros/tree/ros2-development/realsense2_camera/examples/launch_params_from_file) example.

fantalukas:
Hi, thank you very much for very quick responses.

@Nir-Az ok, now I understand the main purpose of `rs_launch.py`. We will try to create our own launch file in our project with our specific needs.

@Arun-Prasad-V very interesting approach I did not know that, thank you. We will try that approach as well.

SamerKhshiboun:
Hi @fantalukas,
Let us know if you need more help with the launch file.
Closing this PR.

PR #2940, opened 2023-11-29
Fixing the data_type of ROS Params exposure & gain

Arun-Prasad-V:
Tracked in RSDEV-1182

### DIFF
--- a/realsense2_camera/src/ros_sensor.cpp
+++ b/realsense2_camera/src/ros_sensor.cpp
@@ -154,8 +154,8 @@ void RosSensor::UpdateSequenceIdCallback()
             {
                 set_option(RS2_OPTION_SEQUENCE_ID, parameter.get_value<int>());
                 std::vector<std::function<void()> > funcs;
-                funcs.push_back([this](){set_sensor_parameter_to_ros(RS2_OPTION_GAIN);});
-                funcs.push_back([this](){set_sensor_parameter_to_ros(RS2_OPTION_EXPOSURE);});
+                funcs.push_back([this](){set_sensor_parameter_to_ros<int>(RS2_OPTION_GAIN);});
+                funcs.push_back([this](){set_sensor_parameter_to_ros<int>(RS2_OPTION_EXPOSURE);});
                 _params.getParameters()->pushUpdateFunctions(funcs);
             });
     }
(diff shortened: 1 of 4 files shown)


realsense2_camera/src/ros_sensor.cpp
```diff
@@ -153,7 +153,7 @@ void RosSensor::set_sensor_parameter_to_ros(rs2_option option)
 {
     std::string module_name = create_graph_resource_name(rs2_to_ros(get_info(RS2_CAMERA_INFO_NAME)));
     const std::string option_name(module_name + "." + create_graph_resource_name(rs2_option_to_string(option)));
-    float value = get_option(option);
+    int value = get_option(option);
```
Nir-Az: Can you explain how is this solve the issue?
I we do need it I suggest going with
‘auto value = static_cast<int>(get_option(option))’

Nir-Az: What about float options? All will be converted to int?
Exposure for example?

realsense2_camera/src/ros_sensor.cpp
```diff
@@ -153,7 +153,7 @@ void RosSensor::set_sensor_parameter_to_ros(rs2_option option)
...
     _params.getParameters()->setRosParamValue(option_name, &value);
```
Arun-Prasad-V, line 175: > Can you explain how is this solve the issue?

Here the address of the float variable "value" is getting passed to "setRosParamValue" function.
Since the ROS params (exposure & gain) are of int type, it expects an integer value in that address, and it tries to read an integer value from that address. Since, the representations of float and integer are different, wrong value is read.

> I we do need it I suggest going with
>‘auto value = static_cast(get_option(option))’

Will try this.

realsense2_camera/src/ros_sensor.cpp
```diff
@@ -136,8 +136,8 @@ void RosSensor::UpdateSequenceIdCallback()
             {
                 set_option(RS2_OPTION_SEQUENCE_ID, parameter.get_value<int>());
                 std::vector<std::function<void()> > funcs;
-                funcs.push_back([this](){set_sensor_parameter_to_ros(RS2_OPTION_GAIN);});
-                funcs.push_back([this](){set_sensor_parameter_to_ros(RS2_OPTION_EXPOSURE);});
+                funcs.push_back([this](){set_sensor_parameter_to_ros<int>(RS2_OPTION_GAIN);});
+                funcs.push_back([this](){set_sensor_parameter_to_ros<int>(RS2_OPTION_EXPOSURE);});
```
Nir-Az, line 158: Nice implementation .

Verified as working?
Anyway we can add a unit test for that?

Nir-Az (approved):
LGTM, let's add a test for that and merge it

realsense2_camera/src/ros_sensor.cpp
Arun-Prasad-V: Unfortunately this cannot be tested with rosbag testcase, since the depth controls (such as exposure, etc.,)  are not exposed for rosbag scenario. 
So, I have added a live camera test and verified locally with D455 that it is failing in 'ros2-development' branch and passing with this PR.

PrasRsRos (approved):

Nir-Az (approved):

PR #2946, opened 2023-12-14
fix ros random crash with error hw monitor command for asic temperature failed

gwen2018:
fix ros random crash with error hw monitor command for asic temperature failed

changes:
add error handling to avoid crash when asic temperature monitor occasionally fails and throws exception

Tracked on: LRS-982

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -1140,10 +1140,18 @@ void BaseRealSenseNode::startDiagnosticsUpdater()
             {
                 for (rs2_option option : _monitor_options)
                 {
-                    if (sensor->supports(option))
+                    try
                     {
-                        status.add(rs2_option_to_string(option), sensor->get_option(option));
-                        got_temperature = true;
+                        if (sensor->supports(option))
+                        {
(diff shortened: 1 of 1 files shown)


Nir-Az (approved):
LGTM

tonynajjar:
hey @Nir-Az , I'm finding myself needing this in 4.54.1, since this is a tag and not a branch, how do you handle backports in this repo? More generally how can I get this fix in 4.54.1 without upgrading anything else?

Nir-Az:
Hi, 
we released 4.55.1 today, should be merged soon.
If you need this fix on previous versions, you can cherry-pick it and build from source..

tonynajjar:
@Nir-Az Thanks for answering quickly

> you can cherry-pick it and build from source..

I was hoping I'd have to avoid this... but I guess there is no other way when the repo uses tags instead of branches. 
I'm just worried that 4.55.1 might be incompatible with the librealsense version I have (2.54.2). There is no "Supported RealSense SDK" for the latest release, can you say whether it's compatible?

Nir-Az:
We are trying to update our release flow in the future.
Could be the new flow can help on such cases.
Stay tuned :)

PR #2956, opened 2023-12-29
Extending LibRS's GL support to RS ROS2

Arun-Prasad-V:


realsense2_camera/CMakeLists.txt
```diff
@@ -117,6 +117,7 @@ find_package(tf2 REQUIRED)
 find_package(diagnostic_updater REQUIRED)
 
 find_package(realsense2 2.54.1)
+find_package(realsense2-gl 2.54.1)
```
Nir-Az: This means we will need to publish realsense2-gl as well right?
We need to think about if we need to make this optional somehow

realsense2_camera/CMakeLists.txt
```diff
@@ -215,6 +218,7 @@ set(dependencies
   nav_msgs
   tf2
   realsense2
+  realsense2-gl
```
Nir-Az: Same, OK for POC
For final solution we will need to rethink about it

realsense2_camera/src/parameters.cpp
```diff
@@ -82,6 +82,10 @@ void BaseRealSenseNode::getParameters()
     _base_frame_id = _parameters->setParam<std::string>(param_name, DEFAULT_BASE_FRAME_ID);
     _base_frame_id = (static_cast<std::ostringstream&&>(std::ostringstream() << _camera_name << "_" << _base_frame_id)).str();
     _parameters_names.push_back(param_name);
+
+    param_name = std::string("data_processing_mode");
```
Nir-Az: maybe `accelerate_with_gpu`  , on/off``?
Or as is with values `CPU/GPU` ?

Arun-Prasad-V: Changed the param name to `accelerate_with_gpu` and declared this param as `int` not `bool`. Because, if in case in future, we need to add new values (like for CUDA, etc.,), it can be added easily to the `int` param.
@Nir-Az, your thoughts?

realsense2_camera/CMakeLists.txt
Arun-Prasad-V: Have added a cmake variable `BUILD_ACCELERATE_WITH_GPU`.

Nir-Az: As discussed, can you check how we do it in the viewer?
Do we depend on realsense-gl and can disable/enable on runtime?

realsense2_camera/src/parameters.cpp
Nir-Az: Not sure CUDA is replacing GPU, let's make sure how it works in LibRS

realsense2_camera/include/gl_window.h
```diff
@@ -0,0 +1,84 @@
+// Copyright 2023 Intel Corporation. All Rights Reserved.
```
SamerKhshiboun, line 1: 2024

realsense2_camera/src/gl_gpu_processing.cpp
```diff
@@ -0,0 +1,45 @@
+// Copyright 2023 Intel Corporation. All Rights Reserved.
```
SamerKhshiboun, line 1: 2024

realsense2_camera/CMakeLists.txt
SamerKhshiboun: > As discussed, can you check how we do it in the viewer? Do we depend on realsense-gl and can disable/enable on runtime?

Yes, I think we should release/publish realsense2-gl as well.
The viewer is depending on realsense2-gl, and you can choose this from the settings of the viewer, whether to run or not the GLSL.
@Arun-Prasad-V, did you test if user changes the parameter from 0 to 1 during runtime, while the build was without OpenGL ?

realsense2_camera/CMakeLists.txt
SamerKhshiboun: @Nir-Az I don't think there is another way for doing that.. except if you want to create new branch for ROS, let's say, ros2-development-gl :)

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -22,6 +22,10 @@ using namespace rs2;
 
 void BaseRealSenseNode::setup()
 {
+#if defined (ACCELERATE_WITH_GPU)
+    bool use_gpu_processing = (_accelerate_with_gpu == accelerate_with_gpu::GL_GPU);
+    initOpenGLProcessing(use_gpu_processing);
```
SamerKhshiboun, line 27: can we save this call if `accelerate_with_gpu != GL_GPU` ?
or, we must call it and init things there ?

realsense2_camera/CMakeLists.txt
Nir-Az: We need to follow this guidelines same as for the SDK.
1. Users w/o GPU are not affected (meaning users w/o GPU can install and run the ROS node)
3. Users with GPU can decide if to leverage the GPU or not (launch file is sufficient)
4. We need to activate the internal formal conversion using GPU as well

realsense2_camera/CMakeLists.txt
Nir-Az: OK, So we will have to make sure this compile on platforms w/o a GPU, and that ROS build farm accept it

realsense2_camera/CMakeLists.txt
Arun-Prasad-V: > @Arun-Prasad-V, did you test if user changes the parameter from 0 to 1 during runtime, while the build was without OpenGL ?

Right now, dynamic change is not supported. User must set this param during launch.
And, if the build was without OpenGL, then this param won't be exposed to user.

realsense2_camera/src/rs_node_setup.cpp
Arun-Prasad-V: > or, we must call it and init things there ?

We must call `initOpenGLProcessing()` and init things even for `accelerate_with_gpu != GL_GPU` case, because the filter objects are created from `rs::gl::colorizer`, `rs2::gl::align` and `rs2::gl::pointcloud`. classes

realsense2_camera/include/gl_window.h
Arun-Prasad-V: CI is failing if we update it to 2024. We should update the script [pr_check.sh](https://github.com/IntelRealSense/realsense-ros/blob/ros2-development/scripts/pr_check.sh#L46) which checks the year first and update in all of the source files.

Arun-Prasad-V: Decided to do that in different PR.

SamerKhshiboun (approved):
LGTM, Excellent work !

Nir-Az (approved):
Great work!

PR #2957, opened 2024-01-01
to_urdf fun retrun a str, not a BufferedRandom

hellototoro:
My env:
Ubuntu22.04
ros2 humble

When I run `ros2 launch realsense2_description view_model.launch.py model:=test_d435i_camera.urdf.xacro`, I received a error message
`[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught exception when trying to load file of format [py]: expected str, bytes or os.PathLike object, not BufferedRandom`

I used Github's copilot to fix this error and it gives this modification.
It works now, but I don't know Whether this modification is appropriate


Arun-Prasad-V (approved):
LGTM

PR #2984, opened 2024-01-29
Fix All Profiles Test

deep0294:
* Use rs-enumerate-devices with verbose option to fetch necessary data
 * Fix the indices for parsing the Resolution and Frequency
 * Tested with D455 Camera

### DIFF
--- a/realsense2_camera/test/utils/pytest_live_camera_utils.py
+++ b/realsense2_camera/test/utils/pytest_live_camera_utils.py
@@ -76,16 +76,23 @@ def get_depth_profiles(long_data, start_index, end_index):
         if len(depth_profile) == 5:
             profile = depth_profile[0]
             value = depth_profile[1]+"x"+depth_profile[3]
-            format = depth_profile[4]
+            img_format = depth_profile[4]
         elif len(depth_profile) == 6:
             profile = depth_profile[0]+depth_profile[1]
             value = depth_profile[2]+"x"+depth_profile[4]
-            format = depth_profile[5]
+            img_format = depth_profile[5]
+        elif len(depth_profile) == 7:
+            profile = depth_profile[1]
+            value = depth_profile[2]+"x"+depth_profile[5]
+            img_format = depth_profile[3]
+        elif len(depth_profile) == 8:
+            profile = depth_profile[1]+depth_profile[2]
+            value = depth_profile[3]+"x"+depth_profile[6]
+            img_format = depth_profile[4]
         else:
             assert false, "Seems that the depth profile info format printed by rs-enumerate-devices changed"
-        value = value[:-2]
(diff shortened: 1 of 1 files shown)


realsense2_camera/test/utils/pytest_live_camera_utils.py
```diff
@@ -100,12 +107,15 @@ def get_color_profiles(long_data, start_index, end_index):
         if len(color_profile) == 5:
             profile = color_profile[0]
             value = color_profile[1]+"x"+color_profile[3]
-            format = color_profile[4]
+            img_format = color_profile[4]
+        elif len(color_profile) == 7:
```
Nir-Az: If just `rs-enumerate` output format was changed, why do we need this extra `if's`?

Nir-Az: Oh, I understand..
`color_profile` is not acctually color profile, it's a splitted line from rs-enumerate?
Maybe we can rename it and maybe add const vars for the indexes with the actual meaning?
example

```static const size_t color_res_width = 1;
static const size_t color_res_width = 3;```

Comment (deep0294) [reply realsense2_camera/test/utils/pytest_live_camera_utils.py]: Good point ! 
Thank you, We shall change.

File: realsense2_camera/test/utils/pytest_live_camera_utils.py [hunk 2]
```diff
@@ -72,27 +72,29 @@ def get_depth_profiles(long_data, start_index, end_index):
...
-        elif len(depth_profile) == 7:
-            profile = depth_profile[1]
-            value = depth_profile[2]+"x"+depth_profile[5]
-            img_format = depth_profile[3]
-        elif len(depth_profile) == 8:
-            profile = depth_profile[1]+depth_profile[2]
-            value = depth_profile[3]+"x"+depth_profile[6]
-            img_format = depth_profile[4]
+        enumerate_devices_line_splitted = long_data[line_no].split()
+        if len(enumerate_devices_line_splitted) == 7:
```
Nir-Az, line 76: 5 & 6 are not needed anymore?

deep0294: Since We moved the tool output from default to verbose, We don't need 5 and 6.
verbose outputs required data in either spllited length of 7/8.

Nir-Az (approved):

Nir-Az:
@SamerKhshiboun / @Arun-Prasad-V  any inputs from you before we merge this?

SamerKhshiboun:
> @SamerKhshiboun / @Arun-Prasad-V any inputs from you before we merge this?

Nothing special from my side.

Arun-Prasad-V (approved):
Looks good.

PR #2989, opened 2024-02-01
[GL-support] Dynamically switching b/w CPU & GPU processing

Arun-Prasad-V:
Tracked by LRS-1001

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,20 @@ User can set the camera name and camera namespace, to distinguish between camera
     - 1 -> **copy**: Every gyro message will be attached by the last accel message.
     - 2 -> **linear_interpolation**: Every gyro message will be attached by an accel message which is interpolated to gyro's timestamp.
   - Note: When the param *unite_imu_method* is dynamically updated, re-enable either gyro or accel stream for the change to take effect.
+- **accelerate_with_gpu**:
+  - GPU accelerated processing of PointCloud and Colorizer filters.
+  - integer:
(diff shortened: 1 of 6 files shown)


Arun-Prasad-V:
I still need to update the readme for this change.

SamerKhshiboun:
@Arun-Prasad-V, I understand the flow and the mechanism behind this PR. My question is, couldn't we use this function `BaseRealSenseNode::updateSensors()` somehow with some parameters, in order to reduce code duplications?
I see that `StartVideoSensors `and `StopVideoSensors `and also some other parts of the PR are common with `BaseRealSenseNode::updateSensors()`.
It could make the `updateSensors `method longer, and we can split it to multiple functions if needed, but again, my idea is to reuse the same code as much as possible, if that is possible :)

Arun-Prasad-V:
> @Arun-Prasad-V, I understand the flow and the mechanism behind this PR. My question is, couldn't we use this function `BaseRealSenseNode::updateSensors()` somehow with some parameters, in order to reduce code duplications? I see that `StartVideoSensors `and `StopVideoSensors `and also some other parts of the PR are common with `BaseRealSenseNode::updateSensors()`. It could make the `updateSensors `method longer, and we can split it to multiple functions if needed, but again, my idea is to reuse the same code as much as possible, if that is possible :)

Sure, @SamerKhshiboun. Makes sense. will update it.

SamerKhshiboun (approved):
LGTM

PR #2998, opened 2024-02-07
Use zeroed intrinsic for profiles that does not have defined intrinsic

SamerKhshiboun:
Tracked on [LRS-1022]

- removed also unused parameter `_stream_intrinsics`


Nir-Az (approved):
Sounds good to me

PR #3019, opened 2024-02-23
Support for selecting stream_type specific profile

Arun-Prasad-V:
Tracked by RSDEV-1745


realsense2_camera/src/profile_manager.cpp
```diff
@@ -139,6 +139,43 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
     return sip_default_profiles;
 }
 
+rs2::stream_profile VideoProfilesManager::validateAndGetSuitableProfile(rs2_stream stream_type, rs2::stream_profile given_profile)
+{
+    if (_all_profiles.empty()) 
+        throw std::runtime_error("Wrong commands sequence. No profiles set.");
```
SamerKhshiboun: can you elaborate on this situation? can you please give an example ?

realsense2_camera/src/profile_manager.cpp
```diff
@@ -139,6 +139,43 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
...
+            if (isSameProfileValues(temp_profile, video_profile.width(), video_profile.height(), video_profile.fps(), RS2_FORMAT_ANY))
+            {
+                is_given_profile_suitable = true;
+                break;
+            }
+        }
+    }
+
+    // If given profile is not suitable, choose the first available profile for the given stream.
+    if (!is_given_profile_suitable)
```
SamerKhshiboun, line 162: and also here, can you give me an example ?

SamerKhshiboun:
Looks good. Only answer my two questions please.
We need to decide if we take this change to development, and when, because it breaks old API.

realsense2_camera/src/profile_manager.cpp
Arun-Prasad-V: Depth Mapping Camera module doesn't have any default profiles (for both occupancy and labeled pointcloud streams). So, `getDefaultProfiles()` function will consider the first listed profile as default profile. Due to this reason, lpc stream's profile will be considered as default profile. But that profile is not supported by Occupancy stream. 

So here, for Occupancy stream, the given profile (lpc's profile) is not suitable. If not suitable, consider the first detected Occupancy stream profile.

Let's apply same scenario for Depth module. Depth stream has default profile, but IR streams doesn't. Fortunately, depth stream's default profile is suitable for IR streams. No issues for this scenario.

Arun-Prasad-V: This case can be handled either this way or inside `getDefaultProfiles()` function itself by providing one default profile per stream.

realsense2_camera/src/profile_manager.cpp
Arun-Prasad-V: This error check was already there in `getDefaultProfiles()` function. Will remove it as it is not required here.

Arun-Prasad-V:
This feature is now available in ros2-development branch. Reference PR https://github.com/IntelRealSense/realsense-ros/pull/3052. It will be taken to ros2-hkr branch while merging.

PR #3032, opened 2024-03-03
Support occupancy grid cells

SamerKhshiboun:
Tracked by [LRS-996]

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -568,6 +568,10 @@ void BaseRealSenseNode::frame_callback(rs2::frame frame)
             {
                 publishPointCloud(f.as<rs2::points>(), t, frameset);
             }
+            else if(stream_type == RS2_STREAM_OCCUPANCY)
+            {
+                publishOccupancyFrame(f, t);
+            }
             else
             {
                 if (stream_type == RS2_STREAM_DEPTH)
@@ -610,14 +614,21 @@ void BaseRealSenseNode::frame_callback(rs2::frame frame)
                     rs2_stream_to_string(stream_type), stream_index, frame.get_frame_number(), frame_time, t.nanoseconds());
             
         stream_index_pair sip{stream_type,stream_index};
-        if (frame.is<rs2::depth_frame>())
+        if(stream_type == RS2_STREAM_OCCUPANCY)
         {
-            if (_clipping_distance > 0)
-            {
-                clip_depth(frame, _clipping_distance);
-            }
+            publishOccupancyFrame(frame, t);
         }
-        publishFrame(frame, t, sip, _images, _info_publishers, _image_publishers);
+        else 
+        {
+            if (frame.is<rs2::depth_frame>())
+            {
+                if (_clipping_distance > 0)
+                {
+                    clip_depth(frame, _clipping_distance);
+                }
+            }
+            publishFrame(frame, t, sip, _images, _info_publishers, _image_publishers);
+        }
     }
     else if (frame.is<rs2::labeled_points>())
     {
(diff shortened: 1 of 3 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -828,7 +839,47 @@ bool BaseRealSenseNode::shouldPublishCameraInfo(const stream_index_pair& sip)
...
+    auto cell_size = static_cast<float>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_CELL_SIZE) / 100.0f); // convert to meters
+
+    // create GridCells msg and start filling it
+    nav_msgs::msg::GridCells msg;
+    msg.header.stamp = t;
+    msg.header.frame_id = FRAME_ID(OCCUPANCY);
+    msg.cell_width = cell_size;
+    msg.cell_height = cell_size;
+
+    for (auto i = 0; i < cols * rows; ++i)
```
Arun-Prasad-V, line 862: @SamerKhshiboun, just thinking out loud for future enhancement. Maybe it makes sense to parallelize this operation by using SSSE3 or GPU acceleration. your thoughts?

Nir-Az, line 862: Does this means rows and columns represent the cells and not the bytes?
meaning they are in `bit` units?
raws = 8 is 1 byte?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -828,7 +839,47 @@ bool BaseRealSenseNode::shouldPublishCameraInfo(const stream_index_pair& sip)
...
+    {
+        // check if current cell is occupied
+        // each byte include 8 bits <--> 8 cells
+        if ((frame_as_uint8_arr[i / 8U] & (1U << i % 8)) != 0) 
+        {
+            // find x,y,z positions of current index
+            geometry_msgs::msg::Point p3d;
+            uint32_t row = (i / cols);
+            uint32_t col = (i % cols);
+            p3d.x = (cell_size * static_cast<float>(rows)) - cell_size * (static_cast<float>(row) + 0.5f);
```
Nir-Az, line 876: I would expect x axis will use columns and not rows, can you explain this line?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -828,7 +839,47 @@ bool BaseRealSenseNode::shouldPublishCameraInfo(const stream_index_pair& sip)
...
+
+    for (auto i = 0; i < cols * rows; ++i)
+    {
+        // check if current cell is occupied
+        // each byte include 8 bits <--> 8 cells
+        if ((frame_as_uint8_arr[i / 8U] & (1U << i % 8)) != 0) 
+        {
+            // find x,y,z positions of current index
+            geometry_msgs::msg::Point p3d;
+            uint32_t row = (i / cols);
```
Nir-Az, line 874: raw is calculates from cols?
Why not from `RS2_FRAME_METADATA_OCCUPANCY_GRID_ROWS`?

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: in ROS CS, X is forward, Y is left, and Z is up.
in AICV algo ROWS represents the height is and COLS represents the width.
image

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: Yes, I think it worth checking if we can accelerate this.
If so, we can do this in a separated task.
Thanks for the advice !

SamerKhshiboun: rows = grid cell height, cols = grid cell width
added more documentation for this code

Nir-Az (approved):

PR #3042, opened 2024-03-13
Assert Fail if camera not found

kadiredd:
Live camera tests updated to have assertion when the camera device not connected to avoid false CI pass


kadiredd:
will reopen

Arun-Prasad-V (approved):
LGTM.

PR #3043, opened 2024-03-13
Update README.md

pedrombmachado:
sudo should not be used for ROS2

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,8 @@
   - Install dependencies
    ```bash
    sudo apt-get install python3-rosdep -y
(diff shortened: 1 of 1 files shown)


Nir-Az:
Thanks for the contribution.

According to ROS2 official docs we see it is required to use sudo normally 


image

SamerKhshiboun:
According to latest ROS documentation: We should `sudo rosdep init`
Reference: https://docs.ros.org/en/rolling/Tutorials/Intermediate/Rosdep.html

Issue #1952, opened 2021-06-25
error: ‘rmw_qos_profile_unknown’ was not declared in this scope


cardboardcode:
Hi Intel team and all, 

Thank you for maintaining this essential ROS driver for the Realsense cameras.

This issue has already been resolved on my side. However, I will submit it here for anyone who have encountered this exact error when compiling the driver package for **ROS2 Eloquent**.

## Issue 
I encountered the following compilation error when running a `colcon build` for **realsense_ros** in the branch [eloquent](https://github.com/IntelRealSense/realsense-ros/tree/eloquent). 
```bash
/home/user/ros_workspaces/ros2/official_realsense_ws/src/realsense-ros/realsense2_camera/src/base_realsense_node.cpp:769:16: error: ‘rmw_qos_profile_unknown’ was not declared in this scope
         return rmw_qos_profile_unknown;

```

# Environment
1. Ubuntu 18.04
2. ROS2 Eloquent 
3. ros-eloquent-rmw 0.8.1 and 0.8.2
```bash
ros-eloquent-rmw/bionic,now 0.8.1-1bionic.20201207.145507 amd64 [installed,automatic]
ros-eloquent-rmw-fastrtps-cpp/bionic,now 0.8.2-1bionic.20201207.150536 amd64 [installed,automatic]
ros-eloquent-rmw-fastrtps-shared-cpp/bionic,now 0.8.2-1bionic.20201207.145922 amd64 [installed,automatic]
ros-eloquent-rmw-implementation/bionic,now 0.8.2-2bionic.20201207.150907 amd64 [installed,automatic]
ros-eloquent-rmw-implementation-cmake/bionic,now 0.8.1-1bionic.20201204.214731 amd64 [installed,automatic]

```

# How To Reproduce Error
Run the commands below to reproduce the error above.
```bash
cd $HOME
mkdir -p official_realsense_ws/src && cd ~/official_realsense_ws/src
git clone --depth 1 --branch `git ls-remote --tags https://github.com/IntelRealSense/realsense-ros.git | grep -Po "(?<=tags/)3.\d+\.\d+" | sort -V | tail -1` https://github.com/IntelRealSense/realsense-ros.git
cd ~/official_realsense_ws/
source /opt/ros/eloquent/setup.bash
colcon build

```

Please see below for the solution to this issue.

cardboardcode:
## Solution

[...]

PR #3048, opened 2024-03-17
Update base_realsense_node.cpp

pedrombmachado:
Fix for  error: ‘rmw_qos_profile_unknown’ was not declared in this scope #1952


SamerKhshiboun:
Hi @pedrombmachado and thanks for your contribution.
This bug was fixed for new supported distros under our maintained branch **ros2-development**.
the **ros2-legacy** is not maintained any more. I will close your PR but keep it here as "closed" so it can benefits other users.

PR #3053, opened 2024-03-20
Fix Coverity issues + remove empty warning log

Nir-Az:
Tracked on [LRS-1049]


SamerKhshiboun (approved):
LGMT, did you run a Coverity scan and can see the new results ?

PR #3058, opened 2024-03-26
Merge ros2 development to ros2 hkr

Arun-Prasad-V:


SamerKhshiboun (approved):
LGTM

Nir-Az:
Please update after a short sanity if we should place this commit in the bundle
________________________________

PR #3063, opened 2024-03-27
Updated pr_check.sh script to not verify the year

Arun-Prasad-V:


Nir-Az:
Please notice that lately we changed our policy
New file gets current year.
Updated file gets the creation year - current year
i.e.
`// Copyright(c) 2015-2024 Intel Corporation. All Rights Reserved.`
Will this regex work on that cases?

Arun-Prasad-V:
@Nir-Az, No. This regex will fail in that case.

Nir-Az:
> @Nir-Az, No. This regex will fail in that case.

Then please adjust

Nir-Az:
> > @Nir-Az, No. This regex will fail in that case.
> 
> Then please adjust

@Arun-Prasad-V ?

Arun-Prasad-V:
@Nir-Az, for now, I will update such that both "2023" and "2015-2024" year formats will work.

In future, we are also thinking of updating the script such that it will perform the check only on the modified files of that particular PR. Also, to check whether the given year is current year.

Nir-Az:
@Arun-Prasad-V  thanks for the fix.
Do we know if it works?
I mean if your are missing a year, will this fail?
Hard to see from the UT log

Arun-Prasad-V:
> @Arun-Prasad-V thanks for the fix. Do we know if it works? I mean if your are missing a year, will this fail? Hard to see from the UT log

@Nir-Az, I have tested it locally.
Right now it passes if the file has any of the below two formats:
- `// Copyright(c) 2015-2024 Intel Corporation. All Rights Reserved.`
- `// Copyright(c) 2023 Intel Corporation. All Rights Reserved.`

It just checks whether there are 4 numeric digits in the YEAR location or not. It doesn't validate whether the year is current year or not. User must provide valid 4-digit year.

Nir-Az (approved):
LGTM

PR #3066, opened 2024-03-31
Revert installation from source for foxy distro on ubuntu 20

SamerKhshiboun:
Tracked on LRS-1060
- Updated main.yaml for github action to build ros wrapper for foxy on ubuntu 20.04
- added foxy to the cmake
- edited package.xml to include test dependencies only if not foxy
- fixed code compilation errors due to deprecated functions calls
- Updated readme
- Fixed rs_launch.py to support foxy (output param should be string)
- Removed compilation warnings


Nir-Az (approved):

Nir-Az (approved):
Nice work!

PR #3071, opened 2024-04-07
Temp fix for Rolling flow 

SamerKhshiboun:
Ref:
https://discourse.ros.org/t/psa-rolling-ci-or-docker-build-fix-from-rosdep-errors-in-24-04-transition/36902
and
https://github.com/ros2/ros2_documentation/pull/4244


SamerKhshiboun:
@Nir-Az, let me know what do you think about this temp fix for rolling.
The only thing that we need to do after that, is to track the rolling sources, and when it will be stable, we should remove this PR

Nir-Az:
> @Nir-Az, let me know what do you think about this temp fix for rolling. The only thing that we need to do after that, is to track the rolling sources, and when it will be stable, we should remove this PR

I think if it helps us see if we break rolling, so this WA should help. your call :)

SamerKhshiboun:
Will merge this after Checks pass, and will assign us a bi-weekly task to check if rolling sources are fixed.

PR #3114, opened 2024-05-30
Ubuntu 24.04 support for Rolling and Jazzy distros

Arun-Prasad-V:
Tracked on LRS-1104

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,16 +25,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [rolling, iron, humble, foxy]
+        ros_distro: [rolling, iron, humble, foxy, jazzy]
         include:
         - ros_distro: 'rolling'
-          os: ubuntu-22.04
+          os: ubuntu-24.04
         - ros_distro: 'iron'
           os: ubuntu-22.04
         - ros_distro: 'humble'
           os: ubuntu-22.04
         - ros_distro: 'foxy'
           os: ubuntu-20.04
+        - ros_distro: 'jazzy'
+          os: ubuntu-24.04
 
     steps:
 
@@ -53,14 +55,14 @@ jobs:
         ./pr_check.sh 
 
     # setup-ros@v0.6 is the last version supporting foxy (EOL)
-    # setup-ros@v0.7 is needed to support humble/iron/rolling
+    # setup-ros@v0.7 is needed to support humble/iron/rolling/jazzy
     # so, seperating steps with if conditions
     - name: build ROS2 for foxy
       if: ${{ matrix.ros_distro == 'foxy' }}
       uses: ros-tooling/setup-ros@v0.6
       with:
         required-ros-distributions: ${{ matrix.ros_distro }}
-    - name: build ROS2 for humble/iron/rolling
+    - name: build ROS2 for humble/iron/rolling/jazzy
       if: ${{ matrix.ros_distro != 'foxy' }}
       uses: ros-tooling/setup-ros@v0.7
       with:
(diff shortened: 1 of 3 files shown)


SamerKhshiboun:
- can you explain more about using the venv ?
- about the rolling.. shouldn't we keep the temp fix ?

Arun-Prasad-V:
- `can you explain more about using the venv ?`
For U24 alone, we are getting an error during installing python3-pip using command `sudo apt-get install python3-pip` similar to the issue mentioned below:
https://stackoverflow.com/questions/75602063/pip-install-r-requirements-txt-is-failing-this-environment-is-externally-mana/75696359#75696359 
The recommendation is to use python in virtual environment. And it works.

- `about the rolling.. shouldn't we keep the temp fix ?`
this temp fix is for U22. If we use U24, it should pass.

realsense2_camera/include/base_realsense_node.h
```diff
@@ -173,7 +173,7 @@ namespace realsense2_camera
         class CimuData
         {
             public:
-                CimuData() : m_time_ns(-1) {};
+                CimuData() : m_data({0,0,0}), m_time_ns(-1) {};
```
Arun-Prasad-V, line 176: if we don't initialize `m_data`, warnings are raised in Jazzy.

.github/workflows/main.yml
```diff
@@ -25,16 +25,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [rolling, iron, humble, foxy]
+        ros_distro: [rolling, iron, humble, foxy, jazzy]
         include:
         - ros_distro: 'rolling'
-          os: ubuntu-22.04
+          os: ubuntu-24.04
```
Nir-Az, line 31: Why iron is changed to U24?

.github/workflows/main.yml
```diff
@@ -126,7 +128,8 @@ jobs:
         # the next command might be needed for foxy distro, since this package is not installed
         # by default in ubuntu 20.04. For other distro, the apt install command will be ignored.
         sudo apt install -y ros-${{matrix.ros_distro}}-sensor-msgs-py
-        python3 src/realsense-ros/realsense2_camera/scripts/rs2_test.py non_existent_file
+        source ../.venv/bin/activate
```
Nir-Az, line 136: Why do we need this source command?

.github/workflows/main.yml
Arun-Prasad-V: I think you meant for 'rolling'. Rolling distro is actually targeted for U24. Till U24 became stable, we were using U22. https://docs.ros.org/en/rolling/Releases/Release-Rolling-Ridley.html#currently-supported-platforms

Nir-Az: Got it Tnx

.github/workflows/main.yml
```diff
@@ -90,15 +92,11 @@ jobs:
...
-        # temp fix for rolling sources.. TODO: track when we can remove the two commands below
-        # see https://discourse.ros.org/t/psa-rolling-ci-or-docker-build-fix-from-rosdep-errors-in-24-04-transition/36902
-        sudo sed -i "s|ros\/rosdistro\/master|ros\/rosdistro\/rolling\/2024-02-28|" /etc/ros/rosdep/sources.list.d/20-default.list
-        export ROSDISTRO_INDEX_URL=https://raw.githubusercontent.com/ros/rosdistro/rolling/2024-02-28/index-v4.yaml
         rosdep update --rosdistro ${{ matrix.ros_distro }} --include-eol-distros
         echo "================= ROSDEP INSTALL ===================="
         rosdep install -i --reinstall --from-path src --rosdistro ${{ matrix.ros_distro }} --skip-keys=librealsense2 -y
         echo "================== COLCON BUILD ======================"
-        colcon build  --cmake-args '-DBUILD_TOOLS=ON'
+        colcon build  --cmake-args '-DBUILD_TOOLS=ON' --no-warn-unused-cli
```
SamerKhshiboun, line 102: nice catch :)
Just an explanation to the record:
BUILD_TOOLS is relevant only for realsense2_camera package, so warnings on unused cmd line variable were printed for realsense2_camera_msgs and realsense2_description.
now with `--no-warn-unused-cli`, there will be no warnings.
ref: https://cmake.org/cmake/help/v3.0/manual/cmake.1.html
`--no-warn-unused-cli`
Don’t warn about command line options.
Don’t find variables that are declared on the command line, but not used.

.github/workflows/main.yml
SamerKhshiboun: I think to activate the virtual env (venv) created in line 116.
when creating venv, we should activate it to "run" inside it.
every command after it will be treated as inside the virtual env.
@Arun-Prasad-V  am I right ?

SamerKhshiboun (approved):
LGTM

.github/workflows/main.yml
Arun-Prasad-V: yes, to use python in virtual env.

.github/workflows/main.yml
```diff
@@ -126,7 +128,8 @@ jobs:
         # the next command might be needed for foxy distro, since this package is not installed
         # by default in ubuntu 20.04. For other distro, the apt install command will be ignored.
         sudo apt install -y ros-${{matrix.ros_distro}}-sensor-msgs-py
-        python3 src/realsense-ros/realsense2_camera/scripts/rs2_test.py non_existent_file
+        source ../.venv/bin/activate
+        ../.venv/bin/python3 src/realsense-ros/realsense2_camera/scripts/rs2_test.py non_existent_file
```
Arun-Prasad-V: I think we can use `python3` instead of `../.venv/bin/python3`, because the venv is already activated in previous line and no need to point the venv location here.

Nir-Az:
@Arun-Prasad-V  please add a Jira number (Tracked on xxx)

.github/workflows/main.yml
Nir-Az: Nice!
I suggest adding a short sentence as a comment above this line.
Same for python venv usage

.github/workflows/main.yml
```diff
@@ -114,9 +112,13 @@ jobs:
        
     - name: Install Packages For Tests
       run: |
+        sudo apt install python3-venv
+        python3 -m venv .venv
```
Nir-Az, line 120: I meant here 😀

Nir-Az (approved):

PR #3124, opened 2024-06-13
[ROS][Test Infra] Support testing ROS2 service call device_info

kadiredd:
Implementation of JIRA ticket req: [LRS-1122]

### DIFF
--- a/realsense2_camera/test/live_camera/test_camera_service_call.py
+++ b/realsense2_camera/test/live_camera/test_camera_service_call.py
@@ -1,0 +1,73 @@
+# Copyright 2024 Intel Corporation. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os, sys
+import pytest
+
+sys.path.append(os.path.abspath(os.path.dirname(__file__)+"/../utils"))
+import pytest_rs_utils
+import pytest_live_camera_utils
+from pytest_rs_utils import launch_descr_with_parameters
+from pytest_rs_utils import get_node_heirarchy
+
+test_params_test_srv_d455 = {
+    'camera_name': 'D455',
+    'device_type': 'D455',
+    }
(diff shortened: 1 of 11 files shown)


realsense2_camera/test/utils/pytest_rs_utils.py
```diff
@@ -833,7 +837,13 @@ def set_bool_param(self, param_name, param_value):
         new_param_value = ParameterValue(type=ParameterType.PARAMETER_BOOL, bool_value=param_value)
         req.parameters = [Parameter(name=param_name, value=new_param_value)]
         return self.send_param(req)
-
+    
+    def get_bool_param(self, param_name):
+        req = GetParameters.Request()
+        req.names = [param_name]
+        value = self.get_param(req)
```
Arun-Prasad-V, line 844: please add a check to verify whether the response is not None.

kadiredd: fixed

realsense2_camera/test/live_camera/test_camera_service_call.py
```diff
@@ -0,0 +1,72 @@
...
+'''
+The test is to check service call device_info with type DeviceInfo
+'''
+@pytest.mark.parametrize("launch_descr_with_parameters", [    
+    pytest.param(test_params_test_srv_d455, marks=pytest.mark.d455),
+    pytest.param(test_params_test_srv_d435i, marks=pytest.mark.d435i),
+    pytest.param(test_params_test_srv_d415, marks=pytest.mark.d415),    
+    ],indirect=True)
+@pytest.mark.launch(fixture=launch_descr_with_parameters)
+class TestCamera_ServiceCall(pytest_rs_utils.RsTestBaseClass):
```
Arun-Prasad-V: this test only uses DeviceInfo service call. So, shall we rename this function and file names such that it is specific to DeviceInfo service.

kadiredd: Yes, thought of it already - There are 8 service calls available. Plan is to have rest seven of them be added here in this file, so no need of changing the file name as of now. And, about changing the function name, yes/maybe -- to be addressed when implementing rest of the service calls. This is kind of placeholder test for validating service calls.

kadiredd: got it, each test will carry distinct function name and distinct class name as well, right? Modified both class and function names.

Arun-Prasad-V (approved):
LGTM.

SamerKhshiboun (approved):
LGTM, but don't we prefer to rename the file name to device info service ? or we want to add all services tests in the future in the same file ?

kadiredd:
> LGTM, but don't we prefer to rename the file name to device info service ? or we want to add all services tests in the future in the same file ?

We want to add all other services in the same file, Yes.

PR #3130, opened 2024-06-16
Install specific numpy version to avoid errors in github actions

SamerKhshiboun:
Numpy has released a new version (2.0.0) before a couple of hours.. and it is failing our GitHub workflows in the testing section... because we have `pip3 install numpy --upgrade` which will install latest version..
 
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,7 +113,8 @@ jobs:
     #    wget $bag_filename -P "records/"
     #    sudo apt install ros-${{ matrix.ros_distro}}-launch-pytest
        
-    - name: Install Packages For Tests
+    - name: Install Packages For Foxy Tests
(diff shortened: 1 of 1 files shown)


SamerKhshiboun:
pre-release builds are expected to fail, since they are working with latest release of librealsense2 (2.55.1), but the current ROS needs the new API (calib config, etc..) of librealsense2/development (which passes in our CIs). So, we can merge even if these two jobs fail.

Nir-Az:
Not related, should we add pre realse for Jazzy as well?

Nir-Az (approved):

SamerKhshiboun:
> Not related, should we add pre realse for Jazzy as well?

Yes, but they still doesn't support it.
https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#supported-ros-distributions

PR #3133, opened 2024-06-19
update librealsense2 version to 2.56.0

SamerKhshiboun:
ros2-development needs new API that is implemented in librealsense2/development (currently 2.56) which still not formally released.


Nir-Az (approved):

Nir-Az:
Why does the pre release fail?

PR #3138, opened 2024-06-22
Support Triggered Calibration as ROS2 Action

SamerKhshiboun:
Tracked on RSDEV-2204

Based on:
https://design.ros2.org/articles/actions.html
https://docs.ros.org/en/jazzy/Tutorials/Intermediate/Writing-an-Action-Server-Client/Cpp.html#writing-an-action-server

- Added new action TriggeredCalibration
- Updated README
- Created new actions.cpp file and implemented the action in it
- Added documentation.
- Testing will be done in RSDEV-2298

### DIFF
--- a/realsense2_camera/src/actions.cpp
+++ b/realsense2_camera/src/actions.cpp
@@ -1,0 +1,121 @@
+// Copyright 2024 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "../include/base_realsense_node.h"
+using namespace realsense2_camera;
+using namespace rs2;
+
+/***
+ * Implementation of ROS2 Actions based on:
+ * ROS2 Actions Design: https://design.ros2.org/articles/actions.html
+ * ROS2 Actions Tutorials/Examples: https://docs.ros.org/en/rolling/Tutorials/Intermediate/Writing-an-Action-Server-Client/Cpp.html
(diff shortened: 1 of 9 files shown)


realsense2_camera/src/actions.cpp
```diff
@@ -0,0 +1,121 @@
+// Copyright 2024 Intel Corporation. All Rights Reserved.
```
Nir-Az, line 1: actions.cpp is a generic name.
Do we want a calibration action dedicated name instead ?

SamerKhshiboun: I think we will add all actions in the future inside it (we can add FW-update action for example, etc..)
(In the same way, I will create a "services.cpp" file for all services - we can redesign this in a separate task)

realsense2_camera/src/actions.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * A callback that accepts a new goal (request) and starts processing it.
+ * Since the execution is a long-running operation, we spawn off a
+ * thread to do the actual work and return from handle_accepted quickly.
+*/
+void BaseRealSenseNode::TriggeredCalibrationHandleAccepted(const std::shared_ptr<GoalHandleTriggeredCalibration> goal_handle)
+{
+    using namespace std::placeholders;
+    ROS_INFO("TriggeredCalibrationAction: Request accepted");
+    // this needs to return quickly to avoid blocking the executor, so spin up a new thread
+    std::thread{std::bind(&BaseRealSenseNode::TriggeredCalibrationExecute, this, _1), goal_handle}.detach();
```
Nir-Az, line 71: Why not opening a new block for the {}?
Do we need to worry about the detach part?
Join on destruction or something?

Nir-Az:
Looks great overall :)

realsense2_camera/src/actions.cpp
SamerKhshiboun: As I saw in the formal examples in ROS2 documentation, this is enough (links attached in the PR discerption)
If you think there is a better way to maintain this, we can discuss this offline

realsense2_camera/src/actions.cpp
Nir-Az: On next action/service please consider creating a service/action folder and implement a file for each serve/action
With time it may grow..

Nir-Az (approved):

PR #3166, opened 2024-07-19
Update Calibration Config API

SamerKhshiboun:
Waiting for https://github.com/IntelRealSense/librealsense/pull/13168 to be merged.
When merged, will rerun failing tests again

part of RSDEV-2332

- Tested manually on SC after merging this change to the private repo
- Updated Readmes accordingly
- Added JSON example to the D500 tables folder
- Updated TC readme to disable some streams before calling TC


realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -560,8 +560,7 @@ void BaseRealSenseNode::CalibConfigReadService(const realsense2_camera_msgs::srv
     try
     {
         (void)req; // silence unused parameter warning
-        rs2_calibration_config calib_config = _dev.as<rs2::auto_calibrated_device>().get_calibration_config();
-        res->calib_config = _dev.as<rs2::auto_calibrated_device>().calibration_config_to_json_string(calib_config);
+        res->calib_config = _dev.as<rs2::auto_calibrated_device>().get_calibration_config();
```
Nir-Az, line 563: Will `as` throw if it's null?
If no, what will it return?

SamerKhshiboun:
```
auto_calibrated_device(device d)
            : calibrated_device(d)
        {
            rs2_error* e = nullptr;
            if (rs2_is_device_extendable_to(_dev.get(), RS2_EXTENSION_AUTO_CALIBRATED_DEVICE, &e) == 0 && !e)
            {
                _dev.reset();
            }
            error::handle(e);
        }
```
when running `as<T>` of device, it will call the constructor above, which may throw an exception if the device doesn't support auto calibration. So in this case, it throws an exception and the behavior is still well-defined, as we return success=false, and we also provide an error message.

Nir-Az (approved):

PR #3196, opened 2024-09-02
move unique ptr instead of copying it

SamerKhshiboun:


Nir-Az:
@SamerKhshiboun  still relevant or can be closed?

PR #3200, opened 2024-09-08
retry thrice finding devices with Ykush reset

kadiredd:
Since occasionally devices are not visible even after ykush reset, retrying thrice finding devices with ykush reset.
Validated the CI overnight and found to be very stable now, 100% of all triggered CI jobs succeeded.
https://rsjenkins.iil.intel.com/view/LRS/job/LRS_linux_compile_ros_ci/


Nir-Az (approved):

Nir-Az:
Any idea why the pre-release CI failed?

SamerKhshiboun:
Because it looks for librealsense 2.56.0 in the ROS apt.
While the last one we release there is 2.55.1

PR #3222, opened 2024-10-05
Support D555 and its motion profiles

SamerKhshiboun:
- RSDEV-991
- LRS-847
- Support D555 in USB and DDS Mode
- Add new Motion stream for D555 DDS Mode
- Fixed a bug in IMU_INFO (not related to DDS/D555 - reproducible also in other devices)

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -118,7 +118,7 @@ find_package(tf2_ros REQUIRED)
 find_package(tf2 REQUIRED)
 find_package(diagnostic_updater REQUIRED)
 find_package(OpenCV REQUIRED COMPONENTS core)
-find_package(fastrtps REQUIRED)  # a workaround, till librealsense2 cmake config will be fixed
+find_package(fastrtps REQUIRED)  # TODO: LRS-1203 - This is a workaround, till librealsense2 cmake config will be fixed
 
 find_package(realsense2 2.56.0)
 if (BUILD_ACCELERATE_GPU_WITH_GLSL)
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -118,6 +118,7 @@ find_package(tf2_ros REQUIRED)
 find_package(tf2 REQUIRED)
 find_package(diagnostic_updater REQUIRED)
 find_package(OpenCV REQUIRED COMPONENTS core)
+find_package(fastrtps REQUIRED)  # a workaround, till librealsense2 cmake config will be fixed
 
 find_package(realsense2 2.56.0)
 if (BUILD_ACCELERATE_GPU_WITH_GLSL)
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -480,7 +480,26 @@ void BaseRealSenseNode::imu_callback(rs2::frame frame)
                 frame.get_profile().stream_index(),
                 rs2_timestamp_domain_to_string(frame.get_frame_timestamp_domain()));
 
-    auto stream_index = (stream == GYRO.first)?GYRO:ACCEL;
+    stream_index_pair stream_index;
+    
+    if(stream == GYRO.first)
+    {
+        stream_index = GYRO;
+    }
+    else if(stream == ACCEL.first)
+    {
+        stream_index = ACCEL;
+    }
+    else if(stream == IMU.first)
+    {
+        stream_index = IMU;
+    }
+    else 
+    {
+        ROS_ERROR("Unknown IMU stream type.");
+        return;
+    }
+
     rclcpp::Time t(frameSystemTimeSec(frame));
 
     if(_imu_publishers.find(stream_index) == _imu_publishers.end())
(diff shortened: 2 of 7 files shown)


realsense2_camera/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ find_package(tf2_ros REQUIRED)
 find_package(tf2 REQUIRED)
 find_package(diagnostic_updater REQUIRED)
 find_package(OpenCV REQUIRED COMPONENTS core)
+find_package(fastrtps REQUIRED)  # a workaround, till librealsense2 cmake config will be fixed
```
Nir-Az: This can cause issues, we make ROS users be dependent on this package.
I am not sure which platforms support it and also which ros-distros.
We can add it for now but must find a solution for it.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -495,19 +514,35 @@ void BaseRealSenseNode::imu_callback(rs2::frame frame)
         ImuMessage_AddDefaultValues(imu_msg);
         imu_msg.header.frame_id = OPTICAL_FRAME_ID(stream_index);
 
-        auto crnt_reading = *(reinterpret_cast<const float3*>(frame.get_data()));
-        if (GYRO == stream_index)
+        const float3 *crnt_reading = reinterpret_cast<const float3 *>(frame.get_data());
+        if (IMU == stream_index)
+        {
+            // Expecting two float3 objects: first for accel, second for gyro
+            const float3 &accel_data = crnt_reading[0];
```
Nir-Az: Can we double check the size before we dereference it?
Otherwise we may crash on SEGFAULT
is it possible to cast to float[] instead of float *? (not sure)

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -352,8 +359,20 @@ void RealSenseNodeFactory::init()
 void RealSenseNodeFactory::startDevice()
 {
     if (_realSenseNode) _realSenseNode.reset();
+    std::string device_name = _device.get_info(RS2_CAMERA_INFO_NAME);
```
Nir-Az: `std::string pid_str(_device.get_info(RS2_CAMERA_INFO_PRODUCT_ID));`
can be placed at the `else` scope only right?

Nir-Az: Also we have 2 lines here.
both init a string both use different ways.
std::string x = y;
std::string x(y);
Any special reason?

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -295,8 +295,14 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
                 rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(qos), qos));
             // Publish Intrinsics:
             info_topic_name << "~/" << stream_name << "/imu_info";
-            _imu_info_publishers[sip] = _node.create_publisher<IMUInfo>(info_topic_name.str(),
-                                        rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(info_qos), info_qos));
+
```
Nir-Az, line 298: Where did we took the QOS before?
why do we need this change?

realsense2_camera/src/realsense_node_factory.cpp
SamerKhshiboun: fixed, unified the calls to be x(y) on both.
we cannot init it in the else, since it used down below in the switch statement.

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: before, we took it from "info_qos", but this was a buggy thing for imu_info.
This is the only topic in the whole ros-wrapper that defines the publisher and immediately send one message down below it (look for this `_imu_info_publishers[sip]->publish(info_msg);` )

so, I tested it on different cameras, and it's not working as expected.
it should publish the same static message for all subscribers, never mind when the subscriber is requesting this topic, but from my sanity checks, it echoes the relevant data only when we restart the node.
This is a bug not related to D555, but was reproducible also on other devices.

So, I changed the QoS to be manually setted.
For other INFO topics, the "info_qos" is still good, because they publish it wherever there are subscribers, not only once when defining the publisher.

realsense2_camera/src/base_realsense_node.cpp
SamerKhshiboun: added a check for that

realsense2_camera/CMakeLists.txt
SamerKhshiboun: I will open a JIRA and add its number beside this ticket with a "TODO"

SamerKhshiboun: I see you already opened LRS-1203

realsense2_camera/src/rs_node_setup.cpp
SamerKhshiboun: Fixed to used rmw_qos_profile_latched, like the extrinsics topics.

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -107,24 +107,31 @@ void RealSenseNodeFactory::getDevice(rs2::device_list list)
...
+                        msg << "Error extracting usb port from device with physical ID: " << pn << std::endl
+                            << "Please report on github issue at https://github.com/IntelRealSense/realsense-ros";
+                        if (_usb_port_id.empty())
+                        {
+                            ROS_WARN_STREAM(msg.str());
+                        }
+                        else
+                        {
+                            ROS_ERROR_STREAM(msg.str());
+                            ROS_ERROR_STREAM("Please use serial number instead of usb port.");
```
Nir-Az, line 126: The user asked for a USB port?
From the code it looks like the code is using it

Nir-Az (approved):
Small comment about a weird message

PR #3293, opened 2025-02-06
align_depth_to_infra2 enabled, pointcloud and align_depth filters to own files

remibettan:
Tracked by: RSDEV-321


realsense2_camera/examples/pointcloud/rs_pointcloud_infra_launch.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+from ament_index_python.packages import get_package_share_directory
+sys.path.append(os.path.join(get_package_share_directory('realsense2_camera'), 'launch'))
+import rs_launch
+
+local_parameters = [{'name': 'camera_name',                  'default': 'camera', 'description': 'camera unique name'},
+                    {'name': 'camera_namespace',             'default': 'camera', 'description': 'camera namespace'},
+                    {'name': 'enable_color',                 'default': 'false', 'description': 'enable color stream'},
+                    {'name': 'enable_depth',                 'default': 'true', 'description': 'enable depth stream'},
+                    {'name': 'enable_infra1',                'default': 'true', 'description': 'enable depth stream'},
```
Nir-Az: `infra1` is the left sensor?

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -635,6 +637,12 @@ void BaseRealSenseNode::frame_callback(rs2::frame frame)
                         publishFrame(f, t, COLOR, _depth_aligned_image, _depth_aligned_info_publisher, _depth_aligned_image_publishers, false);
                         continue;
                     }
+                    rs2::video_frame original_infra2_frame = frameset.get_infrared_frame(2);
```
Nir-Az: What will `get_infrared_frame(2)` return if we didn't turn on the ir?
Why not moving this into the if only when the user ask to align the depth?

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -258,7 +260,7 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
             _info_publishers[sip] = _node.create_publisher<sensor_msgs::msg::CameraInfo>(camera_info.str(),
                                     rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(info_qos), info_qos));
 
-            if (_align_depth_filter->is_enabled() && (sip != DEPTH) && sip.second < 2)
+            if (_align_depth_filter->is_enabled() && (sip != DEPTH))// && sip.second < 2)
```
Nir-Az: Do you prefer to keep this as a comment?
If yes, then I would add some explanation in the comment

realsense2_camera/examples/pointcloud/rs_pointcloud_infra_launch.py
remibettan: yes

realsense2_camera/src/base_realsense_node.cpp
remibettan: It will return a nullptr object.
I can move it in the if statement. I did it this way for code uniformity, as this is the way it is done in the color's if statement above.

realsense2_camera/src/rs_node_setup.cpp
remibettan: No, I will remove it.

Nir-Az:
@remibettan  can you please take a look on the failed CI flows?

Nir-Az (approved):

PR #3303, opened 2025-02-18
Enable rotation filter for color and depth sensors

noacoohen:
Tracked by LRS-1228


remibettan (approved):
LGTM

remibettan (commented):
In fact, I have a question: why only depth and color?
what about IR for example?

remibettan (approved):
LGTM

Issue #3312, opened 2025-03-04
Base Frame ID parameter missing in launch file


FarStryke21:
Hey! 

I am looking to link my robot with the RealSense camera (a D435i) which is mounted on the tool. My URDF defines the mounting of what would be the camera_link of the sensor. 

The documentation for the repository states the availability of a base_frame_id which can be set during launch. The `rs_launch.py` file seems to be missing this parameter. Apart from this, everything works well!

I have noticed the parameter defaults to `camera_link`. However, the ability to define it at runtime would be a great help as we would not need a redundant transform defining a fixed joint between the link created in the URDF and the camera_link.

Is there a way to expose this parameter to the launch file? I am looking for launching something like this:
`ros2 launch realsense2_camera rs_launch.py pointcloud.enable:=true base_frame_id:=tool_rs_mount_1`

MartyG-RealSense:
Hi @FarStryke21  If using `base_frame_id:=tool_rs_mount_1` in the launch command does not change the base_frame_id, you could try adding it to the end of the list of defined parameters in the rs_launch.py launch file after line 85.

https://github.com/IntelRealSense/realsense-ros/blob/ros2-master/realsense2_camera/launch/rs_launch.py#L85

Something like this:

`{'name': 'base_frame_id',             'default': 'tool_rs_mount_1', 'description': "''"},`

FarStryke21:
Hi @MartyG-RealSense !

That is exactly what I am looking for. Adding this exposes the `base_frame_id `parameter in the launch file. I spent some time tracing how the links are created and named. If I were to add this parameter to the launch configuration, wouldn't it be changed to `camera_tool_rs_mount_1` on execution? My understanding is that the RealSense package enforces a strict naming convention for each camera. 

Now, I could get around and change the link name in my URDF to `camera_tool_rs_mount_1`, solving the issue. But I was hoping to disable the `camera_` which would be prepended to the link names? Simply passing empty strings to the naming arguments in the launch file would still result in the `_` being prepended.

MartyG-RealSense:
My knowledge of URDF and its linkages is admittedly limited, unfortunately.

However, you can define a custom name and namespace for each camera.  Instructions for doing so can be found in the section of the ROS2 wrapper instructions linked to below.

https://github.com/IntelRealSense/realsense-ros?tab=readme-ov-file#camera-name-and-camera-namespace

If the name / namespace parameters are not used then the name and namespace are set to 'camera' by default.

[...]

PR #3313, opened 2025-03-06
Added base_frame_id param to the rs_launch.py file

FarStryke21:
Updates with reference to Issue #3312 .

The PR aims to expose the base_frame_id param which was missing in the `rs_launch.py` file. This allows the user to modify the default base_frame_id from `link` to something of the user's choice when using the launch file.

Following the specified guidelines, I have executed the `pr_check.sh` script, and no errors were reported!


sysrsbuild:

Gilaadb (approved):
My only concern was that the value "link" is defined also in [constants.h](https://github.com/IntelRealSense/realsense-ros/blob/1cbd81be81e807eefb46f098e76381888ffc7001/realsense2_camera/include/constants.h#L84) file. but since there is a low likelihood we will need to change this value in the future, and obviously it helps some users avoid the need to clone and own rs_launch.py - Approving the change.

Thanks for your contribution!

PR #3317, opened 2025-03-12
add shared memory

gkuppa24:
Added shared memory to realsense-ros driver.
Also, created a sh to launch cameras.


sysrsbuild:

PR #3325, opened 2025-04-01
use yakush to switch ports

ashrafk93:


realsense2_camera/test/live_camera/rosci.py
```diff
@@ -126,6 +126,12 @@ def find_devices_run_tests():
     global logdir, device_set, _device_by_sn
     max_retry = 3
 
+    # Mapping of device names to YKUSH hub ports (example mapping)
+    device_port_mapping = {
+        'D455': 1,  # Port 1 for D455
```
Nir-Az: Can we map it dynamically like we do in libci?
We create a map of camera name + sn to port.
This way if tomorrow we will switch ports it will still work

ashrafk93: sure i will try to figure out something with yakush cli.

realsense2_camera/test/live_camera/rosci.py
```diff
@@ -150,6 +188,19 @@ def find_devices_run_tests():
...
+                    
+                    # Enable the port for the target device
+                    port = device_port_mapping.get(device.upper())
+                    if port:
+                        log.i(f"Enabling port {port} for device {device.upper()}")
+                        subprocess.run(f'ykushcmd ykush3 -u {port}', shell=True)
+                        time.sleep(2.0)  # Allow time for the device to initialize
+
+                    # Disable all other ports
+                    for other_device, other_port in device_port_mapping.items():
```
Nir-Az: What do you say, maybe we should disable all and enable the one we want?
This is how we do it on libci

ashrafk93: isnt that what i am doing?
i first map dynamically ports to camera on yakush switch.
then for each device i only enable it and disable all others

Nir-Az: I meant the order.
This is from LibCI
image


First disable all than enable what you need.
We even wait until all removed (if you find it hard you can sleep a bit)
But create less side effect on the OS connection/disconnection effectss

Nir-Az (approved):

Nir-Az (approved):

Issue #3331, opened 2025-04-12
Feature Request - Support TF Prefixing


pondersome:
| Required Info	| |
| --- | --- |
| Camera Model	| D455 |
| Firmware Version | 5.16.0.1 |
| Operating System & Version	| Linux (Ubuntu 24.04) |
| Platform	| N100 x64 PC 16GB | 
| Librealsense SDK Version	| v2.55.1 |
| Language	| python |
| Segment	| Robot |
| ROS Distro	| humble |
| RealSense ROS Wrapper Version	| v4.55.1 |

This is a feature request to enhance support of TF prefix handling in the Realsense ROS2 Wrapper. The purpose is to enable a flexible organized hierarchy of camera related TF frame names. This should support single cameras as well as multiple cameras in the same robot network/domain.

When I launch with camera_namespace and camera_name arguments, the nodes and topics are properly namespaced and camera isolated. But the node-generated tf frames do not support the namespace. Instead they exist in the TF list named like:

```
camera1_link
camera1_color_frame
camera1_color_optical_frame
camera1_depth_frame	
camera1_depth_optical_frame

```

This at least allows us to isolate camera1 frames from camera2's, but it does not well-support more complex configurations. 

In my case, I have two cooperative robots, each with a 4 dof arm and D455s mounted at the end effector. I need to properly isolate tf frames from both cameras so I can attach them to the correct arm. A practice we follow is to mirror the naming of the tf frames off of the namespace hierarchy of the robot and its subsystems. For example we'll have:
```
Two chassis nodes in namespace /mybot1 and /mybot2
Two arm subsystems in namespace /mybot1/arm1 and /mybot2/arm1
```
Similarly we'd like the associated tf labelled with the same hierarchy including using a forward slash as the delimiter:
```
/mybot1/arm1/camera1_link
/mybot2/arm1/camera1_link
```
Even though we can walk the tree in a properly jointed hierarchy, it's also good to support just listing all of the frames by name and they will naturally be ordered and grouped in a similar fashion as the connected tree. This helps with organizing and navigating a complicated TF frame space.

While I use slashes to demark the organization of my frame names, users should have full freedom over the construction of the prefix, so my actual request is to honor a tf_prefix parameter like was apparently supported in the original ROS1 wrapper. This parameter should allow any characters that are legal in a TF frame name. To maintain compatibility with existing usage, the camera_name parameter can still be appended to an optionally provided prefix.

[...]

PR #3332, opened 2025-04-12
Support TF Prefixing

pondersome:
- Added a tf_prefix parameter that can be used to prefix all TF frames produced by the node.
- Added documentation for this parameter and the base_frame_id parameter into the readme. Fixed some typos there too.
- Added functionality to the launch file to warn about unsupported parameters
- Added a test for the new tf_prefix parameter, but only supporting the D455 which I can test with
- To run just the new test: python3 -m pytest test_camera_tf_tests.py -v -k "TestCamera_TestTF_Prefix"


sysrsbuild:

pondersome:
this is meant to address:
https://github.com/IntelRealSense/realsense-ros/issues/3331

realsense2_camera/launch/rs_launch.py
```diff
@@ -106,6 +107,20 @@ def launch_setup(context, params, param_name_suffix=''):
     _config_file = LaunchConfiguration('config_file' + param_name_suffix).perform(context)
     params_from_file = {} if _config_file == "''" else yaml_to_dict(_config_file)
 
+    # Get list of supported parameters
+    supported_params = set(param['name'] for param in configurable_parameters)
+    
+    # Check for unsupported parameters in command line arguments
```
ashrafk93, line 114: i see what you tried to do here, but its not alerting about arguments passed with command line.
i.e.
```
ros2 launch realsense2_camera rs_launch.py blabla:=true
```
doesn't throw any warning

for config file the check works though

image

pondersome: should be good to go. also added another missing parameter. the launch file still doesn't expose all available params, but i'm only adding those i've needed and tested.

ashrafk93 (approved):

PR #3340, opened 2025-04-28
Support rgbd type and fix bug for FPS lower than 1

Gilaadb:


Nir-Az (approved):
Great addition! Thanks

PR #3347, opened 2025-05-04
few logs and exceptions catches added

remibettan:
Tracked by: LRS-1152

### DIFF
--- a/realsense2_camera/src/ros_sensor.cpp
+++ b/realsense2_camera/src/ros_sensor.cpp
@@ -27,7 +27,7 @@ using namespace rs2;
         }                                                            \
         catch(std::exception& e)                                     \
         {                                                            \
-            ROS_WARN_STREAM("Failed to set option: " << e.what());   \
+            ROS_WARN_STREAM("Failed to set option: " << rs2_option_to_string(opt) << " to val: " << val << " error: " << e.what());   \
         }                                                            \
     }                                                                \
     
--- a/realsense2_camera/src/ros_sensor.cpp
+++ b/realsense2_camera/src/ros_sensor.cpp
@@ -18,6 +18,22 @@
 using namespace realsense2_camera;
 using namespace rs2;
 
+#define SAFE_SET_OPTION(opt, val)                                    \
+    if (this->supports(opt))                                          \
+    {                                                                \
+        try                                                          \
+        {                                                            \
+            this->set_option(opt, val);                               \
+        }                                                            \
+        catch(std::exception& e)                                     \
+        {                                                            \
+            ROS_WARN_STREAM("Failed to set option: " << e.what());   \
+        }                                                            \
+    }                                                                \
+    
+    
+        
+
 void RosSensor::setupErrorCallback()
 {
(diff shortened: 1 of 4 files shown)


realsense2_camera/src/ros_sensor.cpp
```diff
@@ -94,7 +94,14 @@ void RosSensor::setParameters(bool is_rosbag_file)
     // So, during init of the node, forcefully disabling the HDR upfront and update it with new values.
     if((!is_rosbag_file) && supports(RS2_OPTION_HDR_ENABLED))
     {
-        set_option(RS2_OPTION_HDR_ENABLED, false);
+        try
+        {
+            set_option(RS2_OPTION_HDR_ENABLED, false);
```
Nir-Az: Why should we fail here?

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -152,14 +161,16 @@ void BaseRealSenseNode::setAvailableSensors()
...
 
-    _dev_sensors = _dev.query_sensors();
-
     for(auto&& sensor : _dev_sensors)
     {
         const std::string module_name(rs2_to_ros(sensor.get_info(RS2_CAMERA_INFO_NAME)));
         std::unique_ptr<RosSensor> rosSensor;
-        if (sensor.is<rs2::depth_sensor>() ||
-            sensor.is<rs2::color_sensor>())
+        if (sensor.is<rs2::depth_sensor>())
```
Nir-Az: If and else code match, what was the purpose here?

remibettan: right

realsense2_camera/src/ros_sensor.cpp
remibettan: This is a call to fw, can eventually fail.
Should not, be better to protect it, as we do for every call to FW.

Nir-Az: On this case I suggest a macro / function that do this is 1 place.
```
if supported
   try
      ...
   catch
      ...
```
as you use it in several places

remibettan: macro added

realsense2_camera/src/ros_sensor.cpp
```diff
@@ -18,6 +18,22 @@
...
+#define SAFE_SET_OPTION(opt, val)                                    \
+    if (this->supports(opt))                                          \
+    {                                                                \
+        try                                                          \
+        {                                                            \
+            this->set_option(opt, val);                               \
+        }                                                            \
+        catch(std::exception& e)                                     \
+        {                                                            \
+            ROS_WARN_STREAM("Failed to set option: " << e.what());   \
```
Nir-Az: Can you please add more info?
Something like
```
ROS_WARN_STREAM("Failed to set option: " << opt << " to val: " << val << " error: " e.what());
```

(just make sure if some casting is needed)

OhadMeir (commented):
I have no additional comments. Please handle previous comments and CI.

Nir-Az (approved):

Issue #3206, opened 2024-09-18
Empty pointcloud data when accelerate_gpu_with_glsl = True


edgarcamilocamacho:
Hello

| Required Info | |
|-|-|
| Camera Model | D435i | 
| Firmware Version | 5.16.0.1 | 
| Operating System | Ubuntu 20.04.6 LTS | 
| Platform | NVIDIA Jetson Orin AGX |
| Jetpack | 5.1.1 [L4T 35.3.1] |
| Kernel Version (Linux Only) | 5.10.104-tegra | 
| Librealsense SDK Version | 2.56.1 | 
| ROS Distro | Humble | 
| RealSense ROS Wrapper Version | 4.56.1 | 

### Issue Description

When I set `accelerate_gpu_with_glsl = True` (at the begining of the execution, not dinamically), I get empty data pointcloud like:

```
header:
  stamp:
    sec: 1726699016
    nanosec: 654941162
  frame_id: camera_depth_optical_frame
height: 1
width: 0
fields:
- name: x
  offset: 0
  datatype: 7
  count: 1
- name: y
  offset: 4
  datatype: 7
  count: 1
- name: z
  offset: 8
  datatype: 7
  count: 1
is_bigendian: false
point_step: 16
row_step: 0
data: []
is_dense: true
```

But with `accelerate_gpu_with_glsl = False` it works perfectly.

Maybe I'm setting something wrong, I attach the list of ros params:

[rs-params.txt](https://github.com/user-attachments/files/17051183/rs-params.txt)

MartyG-RealSense:
Hi @edgarcamilocamacho  In rare cases on some computers, a pointcloud will disappear when GLSL is enabled.  There is no way to predict in advance which computers will be unable to use GLSL, unfortunately.

If you are using a Jetson then the usual way to accelerate a pointcloud will be to enable the librealsense SDK's CUDA support.  This can be done by installing the Jetson version of the librealsense packages on the **installation_jetson.md** instruction page or by building from source code with the flag **-DBUILD_WITH_CUDA=ON** included in the CMake build instruction.

edgarcamilocamacho:
So if I install `librealsense` and `realsense-ros` in the Jetson, both from apt packages, like:

* `librealsense`: https://github.com/IntelRealSense/librealsense/blob/master/doc/installation_jetson.md
![image](https://github.com/user-attachments/assets/1232c617-697f-42d1-8441-245a6a1f4c18)


[...]

PR #3356, opened 2025-05-09
Ashraf/glsl pointcloud

ashrafk93:
Following this PR
https://github.com/IntelRealSense/librealsense/pull/13462#event-17470833592
and issue
https://github.com/IntelRealSense/realsense-ros/issues/3206

While debugging pointcloud behavior with GLSL acceleration enabled, I noticed that in the RealSense Viewer, pointclouds were being generated and rendered correctly. However, in the ROS 2 wrapper (realsense2_camera), pointclouds would not appear under certain conditions — specifically when the parameter pointcloud.stream_profile was left at its default value: any (RS2_STREAM_ANY).

Digging deeper into the ROS pipeline, I observed that when stream_profile is set to any, the system is unable to find a valid texture frame to associate with the pointcloud.
https://github.com/IntelRealSense/realsense-ros/blob/711d1f726ef2b424b4047fe1bcf5ec48128eb811/realsense2_camera/src/pointcloud_filter.cpp#L99

As a result GLSL-based pointcloud computation silently fails due to the lack of a texture stream.

Unlike CPU-based fallback implementations, GLSL requires a valid texture stream (e.g., color or IR) to drive the pointcloud generation pipeline.

This explains why realsense-viewer works while ros2 wrapper doesnt -- it explicitly assigns a texture stream.

To resolve this, I verified that explicitly setting the parameter pointcloud.stream_profile to a valid stream 
successfully enables the pointcloud pipeline, even when GLSL acceleration is used (e.g.,):
```
1 → RS2_STREAM_DEPTH
2 → RS2_STREAM_COLOR
3 → RS2_STREAM_INFRARED
```

RS2_STREAM_DEPTH pointcloud
image

RS2_STREAM_COLOR pointcloud
image

RS2_STREAM_INFRARED pointcloud
image

### DIFF
--- a/realsense2_camera/src/pointcloud_filter.cpp
+++ b/realsense2_camera/src/pointcloud_filter.cpp
@@ -15,6 +15,7 @@
 #include <pointcloud_filter.h>
 #include <fstream>
 #include <sensor_msgs/point_cloud2_iterator.hpp>
+#include "dynamic_params.h"
 
 
 using namespace realsense2_camera;
@@ -95,11 +96,13 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
         if ((!_pointcloud_publisher) || (!(_pointcloud_publisher->get_subscription_count())))
             return;
     }
+    
     rs2_stream texture_source_id = static_cast<rs2_stream>(_filter->get_option(rs2_option::RS2_OPTION_STREAM_FILTER));
     bool use_texture = texture_source_id != RS2_STREAM_ANY;
     static int warn_count(0);
     static const int DISPLAY_WARN_NUMBER(5);
     rs2::frameset::iterator texture_frame_itr = frameset.end();
+    
     if (use_texture)
     {
         std::set<rs2_format> available_formats{ RS2_FORMAT_RGB8, RS2_FORMAT_Y8, RS2_FORMAT_Z16 };
@@ -115,12 +118,15 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
             return;
         }
         warn_count = 0;
-    } else {
+    } 
+    else {
         warn_count++;
         std::string texture_source_name = _filter->get_option_value_description(rs2_option::RS2_OPTION_STREAM_FILTER, static_cast<float>(texture_source_id));
-        ROS_WARN_STREAM_COND(warn_count == DISPLAY_WARN_NUMBER, "No stream match for pointcloud chosen texture " \
-            << texture_source_name << " - " << "set pointcloud.stream_profile parameter to color/depth/infra stream");
-        return;
+        ROS_WARN_STREAM_COND(
+            warn_count == DISPLAY_WARN_NUMBER,
+            "No stream match for pointcloud chosen texture: " << texture_source_name
+            << ". The 'pointcloud.stream_profile' parameter is set to 'any'"
+        );        
     }
 
     int texture_width(0), texture_height(0);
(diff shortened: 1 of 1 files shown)


ashrafk93:
also the PR here:
https://github.com/IntelRealSense/librealsense/pull/13462#event-17470833592

is a valid fallback to calculating point cloud when no texture is found
it Uses only depth intrinsics + the raw depth frame  bypassing the need for a texture
and can be merged as well

Nir-Az:
@remibettan can you please review and maybe try this code?

realsense2_camera/src/pointcloud_filter.cpp
```diff
@@ -115,6 +116,18 @@ void PointcloudFilter::Publish(rs2::points pc, const rclcpp::Time& t, const rs2:
...
+        warn_count++;
+        std::string texture_source_name = _filter->get_option_value_description(
+            rs2_option::RS2_OPTION_STREAM_FILTER,
+            static_cast<float>(texture_source_id)
+        );
+        ROS_WARN_STREAM_COND(
+            warn_count == DISPLAY_WARN_NUMBER,
+            "No matching stream for texture '" << texture_source_name
+            << "'. Set 'pointcloud.stream_profile' to 'depth(1)', 'color(2)', or 'infrared(3)' to enable textured pointclouds."
+        );
```
remibettan, line 130: It seems that you tell the user what is the error, and why the pointcloud cannot be displayed until the texture param is changed - great.
Let's change also the default value, so that the pointcloud will be displayed by default (I think COLOR looks great)

remibettan (commented):
one comment - great catch!

realsense2_camera/src/pointcloud_filter.cpp
ashrafk93: the question is if we want to this merge this PR
https://github.com/IntelRealSense/realsense-ros/pull/3356#issuecomment-2865545051

 as default (same behavior as cpu - non glsl)
or just put a warning.
the first approach is more robust 
@Nir-Az @remibettan

ashrafk93: @remibettan 
providing more details on the issue
when glsl is not enabled everything is fine.
and whether pointcloud.stream_profile param provided or not the pointcloud data is never empty and there are fallbacks for all cases.

when using glsl there is an issue with stream_profile 0 and its confusing to users.

**reproducing the issue:**
 
build realsense ros branch development with glsl on cmake flag
```
 colcon build --cmake-args '-DBUILD_ACCELERATE_GPU_WITH_GLSL=ON' -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
 ```
 
 run realsense node without specifiying a stream profile (i.e default pointcloud.stream_profile when not provided is 0 -> STREAM_ANY)
 ```
 source install/setup.bash
 ros2 run realsense2_camera realsense2_camera_node --ros-args -p initial_reset:=True -p accelerate_gpu_with_glsl:=True -p pointcloud.enable:=True -p enable_color:=True -p enable_depth:=True -p enable_infra1:=True -p enable_infra2:=True -p clip_distance:=2.5 -p pointcloud.stream_filter:=0 --log-level info
```

you would see that point cloud data is empty
```
ros2 topic echo /camera/camera/depth/color/points
```
image

this PR (realsense ros repo) prints a warning to the user when this happens explaning what is the issue
image

and that he should specify the stream_profile param to something else than 0 (STREAM_ANY)

now regarding realsense fix PR that can complete this issue to make realsense ros utilize gpu and calculate pointcloud even with no stream_profile can be provided
this PR https://github.com/IntelRealSense/librealsense/pull/13462/files contributed by the public
handles the case where a stream_profile is not provided or 0.
and implements a new shader that can calculate the pointcloud without any texture (depth, color,infra)

we can just merge he warning msg for now and do some deeper tests on the realsense PR with the fallback suggestions

Nir-Az (approved):

PR #3383, opened 2025-06-18
Fix RGBD camera_info and frame_id

Gilaadb:
The correct frame_id for aligned_depth and the depth camera_info should be the same as the color frame_id and camer_info.


Nir-Az (approved):

Gilaadb:
Will be merged to ros2-development branch first.

PR #3404, opened 2025-07-30
Use package:// to discover meshes

adamwhats:
This is a very small change to the URDFs to allow better portability of the robot_description between computers

I found when the robot description was published by my robot I couldn't view them in RViz on my main PC as the path the the `install/share/realsense2_description/` directory was different between the two workspaces

By replacing the `file://$(find realsense2_description)` with `package://realsense2_description` then the ros resource_retriever can run on the receiving PC and meshes will be retrieved in a more robust manner

### DIFF
--- a/realsense2_description/urdf/_d405.urdf.xacro
+++ b/realsense2_description/urdf/_d405.urdf.xacro
@@ -75,7 +75,7 @@ aluminum peripherial evaluation case.
         <!-- the mesh origin is at front plate in between the two infrared camera axes -->
         <origin xyz="${d405_zero_depth_to_glass + d405_glass_to_front} ${-d405_cam_depth_py} 0" rpy="${M_PI/2} 0 ${M_PI/2}"/>
         <geometry>
-	        <mesh filename="file://$(find realsense2_description)/meshes/d405.stl" scale="0.001 0.001 0.001" />
+	        <mesh filename="package://realsense2_description/meshes/d405.stl" scale="0.001 0.001 0.001" />
         </geometry>
         <material name="aluminum"/>
       </visual>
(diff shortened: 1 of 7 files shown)


sysrsbuild:

Gilaadb (commented):
Hi @adamwhats. Thank you for your contribution!
The changes look good to me, but did you deliberately skip some of the "find" statements?
for example, I saw that some of the files still begin with:
  <xacro:include filename="$(find realsense2_description)/urdf/_materials.urdf.xacro" />
  <xacro:include filename="$(find realsense2_description)/urdf/_usb_plug.urdf.xacro" />
Is it intentionally left that way?

adamwhats:
Hi @Gilaadb, thanks! I actually tested that. What I found is that the xacro preprocesser doesn't seem to resolve the resolve the `package://` URI, and instead it was concatentating with the current directory path, leading to a malformed path:
```
XacroException: No such file or directory: /publisher_ws/install/realsense2_description/share/realsense2_description/urdf/package://realsense2_description/urdf/_d455.urdf.xacro [Errno 2] No such file or directory: '/publisher_ws/install/realsense2_description/share/realsense2_description/urdf/package://realsense2_description/urdf/_d455.urdf.xacro'
```
My understanding is that `package://` URIs aren't resolved during the xacro preprocessing step, and instead are passed through to the final URDF as-is. Then any node which receives the robot description resolves them locally using the [resource_retriever](https://wiki.ros.org/resource_retriever). In contrast, `$(find realsense2_description)` is resolved during xacro processing to an absolute file path, which is what was leading to the portability issue this PR aims to solve.

Here's a Docker setup you can use to validate it. It creates two containers with different workspace paths. If you run it initially, you'll see it error when the rviz-viewer container tries to load the robot_description. Then if you change the `BRANCH_NAME` variable in `.env` to `mesh-package-discovery` and rebuild, you'll (hopefully) see it work

[realsense_mesh_fix_test.zip](https://github.com/user-attachments/files/21565445/realsense_mesh_fix_test.zip)

Gilaadb (approved):
LGTM. Approved.
Thanks!

Nir-Az:
@adamwhats thanks for your PR,
1 small request is to base and target your PR to ros2-development branch so we can merge it.
We do not merge PR's directly to ros2-master, thanks

adamwhats:
Hi @Nir-Az, that should be all sorted now

Nir-Az:
@adamwhats  thanks!

PR #3424, opened 2025-08-19
Use point cloud transport to allow compressed point clouds 

kbrameld:
Replaces the publisher in ``pointcloud_filter.cpp`` with a point_cloud_transport publisher. point_cloud_transport is similar to image_transport, and makes topics available for each installed point_cloud_transport plugin. 

RViz uses point_cloud_transport too and with these changes, it is possible to compress / decompress point clouds efficiently without the need for separate compression / decompression nodes. In RViz, you need to manually enter the "/zstd" topic suffix (if using zstd plugin), as the compressed transports don't show up in the drop down automatically.

Screenshot from 2025-08-19 13-26-51

### DIFF
--- a/realsense2_camera/CMakeLists.txt
+++ b/realsense2_camera/CMakeLists.txt
@@ -120,6 +120,7 @@ find_package(tf2_ros REQUIRED)
 find_package(tf2 REQUIRED)
 find_package(diagnostic_updater REQUIRED)
 find_package(OpenCV REQUIRED COMPONENTS core)
+find_package(point_cloud_transport REQUIRED)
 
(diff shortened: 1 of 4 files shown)


sysrsbuild:

Nir-Az:
Hi, thanks for the PR.
Some questions?

Does it relates to image_transport package?
Does it support intra_process configuration or it needs to be disabled on such like in image_transport?

Will it be triggered to work if the package is installed and if not will work as the normal topic? like in image_transfer?

We need to be back compatable or to add this as an extra feature the user can turn on

PR #3430, opened 2025-08-27
Create a singleton wrapper to rs2::context

Gilaadb:
Maintain a single context when running in a composable node. This should help with having just a single realsense SDK and single source that access shared resources.

Extra minor change: modified some header file includes to just reference the file instead of having a relative path.

### DIFF
--- a/realsense2_camera/include/context_singleton_wrapper.h
+++ b/realsense2_camera/include/context_singleton_wrapper.h
@@ -1,0 +1,32 @@
+// Copyright 2025 RealSense, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <librealsense2/rs.hpp>
+#include <memory>
+
+class RSContextSingletonWrapper {
+public:
+    static rs2::context& getInstance() {
+        static rs2::context ctx;
+        return ctx;
+    }
+
+private:
+    RSContextSingletonWrapper() = delete;
+    ~RSContextSingletonWrapper() = delete;
+    RSContextSingletonWrapper(const RSContextSingletonWrapper&) = delete;
+    RSContextSingletonWrapper& operator=(const RSContextSingletonWrapper&) = delete;
+};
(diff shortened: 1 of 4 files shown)


realsense2_camera/include/context_singleton_wrapper.h
```diff
@@ -0,0 +1,33 @@
+// Copyright 2023 RealSense, Inc. All Rights Reserved.
```
Nir-Az: New files should get current year `2025`

realsense2_camera/include/context_singleton_wrapper.h
```diff
@@ -0,0 +1,33 @@
...
+public:
+    static rs2::context& getInstance() {
+        static rs2::context ctx;
+        return ctx;
+    }
+
+private:
+    RSContextSingletonWrapper() = delete;
+    ~RSContextSingletonWrapper() = delete;
+    RSContextSingletonWrapper(const RSContextSingletonWrapper&) = delete;
```
Nir-Az, line 30: Maybe we should add 
```
    RSContextSingletonWrapper(RSContextSingletonWrapper&&) = delete;
    RSContextSingletonWrapper& operator=(RSContextSingletonWrapper&&) = delete;
```

ChatGPT:
```
Why deleting move is necessary

std::move would otherwise allow you to duplicate the singleton object.

Even if the moved-from object remains valid (depends on how the move is implemented), you’ve violated the invariant of “exactly one instance”.

It also makes ownership semantics unclear — should someone be allowed to “take” the singleton instance away? The answer is no.

That’s why both copy and move operations must be deleted.
```

Nir-Az: He also advise not to delete the Ctor/Dtor but to use default but not sure he is right there.
```
RSContextSingletonWrapper() = default;   // must exist for `instance()` to work
    ~RSContextSingletonWrapper() = default;  // private dtor prevents manual delete
};
```

Nir-Az:
Overall looks great!

realsense2_camera/include/context_singleton_wrapper.h
Gilaadb: It is not correct because the way to access getInstance here is by calling the class method and not an object one. And I've validated it to be working.
I also think that if you are not able to create an object, then you probably are not able to std::move it... Right?

realsense2_camera/include/context_singleton_wrapper.h
```diff
@@ -0,0 +1,33 @@
...
+#include <librealsense2/rs.hpp>
+#include <memory>
+#include <iostream>
+
+class RSContextSingletonWrapper {
+public:
+    static rs2::context& getInstance() {
+        static rs2::context ctx;
+        return ctx;
+    }
```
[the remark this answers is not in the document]
Gilaadb: It is compiled with C++11 or newer.

Nir-Az (approved):

Nir-Az (approved):

PR #3461, opened 2025-12-15
Update cv_bridge header suffix for new ROS2 versions - .h was deprecated

Gilaadb:


remibettan (approved):
Thanks!

remibettan:
Sorry, I need to walk back my earlier approval — looking at this more carefully I see a few issues:

**1. The distro macros (`HUMBLE`, `IRON`, `GALACTIC`, `FOXY`) are not defined anywhere in `realsense2_rgbd_plugin`.** They are only set in `realsense2_camera/CMakeLists.txt`, not for this package. So on every distro the `#else` branch is taken and `cv_bridge.hpp` is always included — which means the build will now break on Foxy/Galactic (and older Humble), the exact distros the conditional was meant to protect.

**2. `realsense2_rgbd_plugin/src/ros_image_texture.cpp` still has the bare `#include <cv_bridge/cv_bridge.h>`.** Same package, same problem — so even on the newer distros this PR is meant to fix, the package still won't build.

**3. It diverges from the existing pattern in the repo.** `realsense2_camera/include/base_realsense_node.h` already solves this with a version-based guard:

```cpp
#if defined(CV_BRDIGE_HAS_HPP)
#include <cv_bridge/cv_bridge.hpp>
#else
#include <cv_bridge/cv_bridge.h>
#endif
```

driven by `realsense2_camera/CMakeLists.txt` which checks `cv_bridge_VERSION >= 3.3.0`. That's the correct approach — it works regardless of distro name and matches the cv_bridge upstream guidance.

**Suggested changes:**
- Apply the same fix to `ros_image_texture.cpp`.
- Reuse the `CV_BRDIGE_HAS_HPP` pattern instead of hardcoded distro names — add the equivalent `cv_bridge_VERSION` check to `realsense2_rgbd_plugin/CMakeLists.txt` and use `#if defined(CV_BRDIGE_HAS_HPP)` in both files.

Nir-Az (approved):
LGTM

PR #3499, opened 2026-03-20
Enable D421 for ROS1

hsuys:
Enable D421 sensor for ROS1


Nir-Az:
@hsuys since we do not compile and test D421 I can merge based on your advise.
Should I merge this change?

hsuys:
@Nir-Az, thank you for the review. 
We've tested D421 working with ros1 wrapper on Ubuntu 20.04 for customer who wants to use this SKU but hasn't migrated to ROS2. 
Please help to merge. Thank you.

Nir-Az (approved):
LGTM

PR #3504, opened 2026-03-31
Switch CI to ros-core containers for strict dep checking

Nir-Az:
## Summary
- Use `ros:<distro>-ros-core` Docker containers instead of `ros-tooling/setup-ros` to start from a minimal ROS base
- This catches missing dependency declarations in `package.xml` (e.g., `action_msgs` on humble) that the full ROS install masks — same class of issues the pre-release workflow was meant to catch
- Builds librealsense2 from the `development` branch directly inside the container (`--depth 1` clone, `$(nproc)` parallel build)
- Removes `pre-release.yml` since `main.yml` now provides the same strict dependency resolution plus all existing checks (copyright, integration tests, all 5 distros)

## Why
The old `pre-release.yml` used industrial_ci `PRERELEASE` mode, but its triple-nested Docker architecture (`host → industrial_ci Docker → ros_buildfarm Docker`) makes it impossible to inject custom-built librealsense2 into the innermost build container. Meanwhile, `main.yml` used `ros-tooling/setup-ros` which installs a full ROS distribution, masking missing deps.

By switching `main.yml` to `ros-core` containers, we get strict dep checking **and** librealsense2 from development in one workflow.

## Test plan
- [ ] All 5 matrix jobs (rolling, iron, humble, jazzy, kilted) pass
- [ ] Verify in logs: container is `ros:<distro>-ros-core`, librealsense2 cloned from `development`, rosdep skips librealsense2 apt package
- [ ] Regression test: temporarily remove a declared dependency (e.g., `action_msgs`) from `package.xml` and verify CI fails


realsense2_camera/package.xml
```diff
@@ -25,6 +25,7 @@
   <depend>rclcpp_lifecycle</depend>
   <depend>lifecycle_msgs</depend>
 
+  <depend>rclcpp_action</depend>
```
Nir-Az, line 28: was missing, the strict GHA flow caught it

realsense2_rgbd_plugin/package.xml
```diff
@@ -12,6 +12,7 @@
 
     <buildtool_depend>ament_cmake</buildtool_depend>
 
+    <depend>cv_bridge</depend>
```
Nir-Az, line 15: was missing, the strict GHA flow caught it

Nir-Az:
@remibettan 
Still thinking what we do with the pre-release that fails.
once the librealsense2 be synced it will pass until the next API change.
So it has added value but also can fail constantly after an API change, thoughts?

remibettan (approved):
Great job!

PR #3506, opened 2026-04-01
Fix TF frame id in camera_info of second infrared camera

exo-core:
I just noticed that the `camera_info` topic published for the `infra2` camera holds the `frame_id` of `infra1`:

```
$ ros2 topic echo /realsense_forward/infra2/camera_info
header:
  stamp:
    sec: 1775037085
    nanosec: 737094482
  frame_id: realsense_forward_infra1_optical_frame
height: 480
```

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -895,7 +895,7 @@ void BaseRealSenseNode::updateExtrinsicsCalibData(const rs2::video_stream_profil
     float fx = _camera_info[right].k.at(0);
     float fy = _camera_info[right].k.at(4);
     const auto& ex = right_video_profile.get_extrinsics_to(left_video_profile);
-    _camera_info[right].header.frame_id = OPTICAL_FRAME_ID(left);
+    _camera_info[right].header.frame_id = OPTICAL_FRAME_ID(right);
     _camera_info[right].p.at(3) = -fx * ex.translation[0] + 0.0; // Tx - avoid -0.0 values.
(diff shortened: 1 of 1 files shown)


MartyG-RealSense:
Hi @exo-core  This question about the infra1 and infra2 frame_id was previously raised in 2020 in regards to the ROS1 wrapper.  My understanding is that the RealSense team member responsible for developing the ROS wrapper at the time opted to remove the infra2 frame_id and have both infras use the same frame_id - please see https://github.com/realsenseai/realsense-ros/pull/1242

I am not part of the current ROS2 wrapper development team though, so do not personally know if this same ID for both infras approach is still used.  A member of the ROS dev team will be able to offer advice on this PR.

exo-core:
Thanks for pointing this out, @MartyG-RealSense. I wasn't aware that `stereo_image_proc` expects `frame_id` to be the _target frame_ after applying stereo rectification. From the documentation of `camera_info` I expected it to be the _source frame_:

```
# Header frame_id should be optical frame of camera
```

In my case, I was only using the intrinsic calibration and was kinda surprised that TF provided me an identity matrix when querying the transformation between the cameras.
But yeah, it also makes sense to tell where that projection takes your data, when you already have the source frame in your image header...

Nir-Az:
@exo-core thanks for the PR, can you please make sure you are based on ros2-developemnt branch?
@Gilaadb can you help in reviewing this? I think we touched it lately, right?

exo-core:
> @exo-core thanks for the PR, can you please make sure you are based on ros2-developemnt branch?

Done. After reading through the threads @MartyG-RealSense linked, I am unsure whether this is still a desired change though...

PR #3517, opened 2026-05-17
Imu handling in component

Gilaadb:
When multiple camera nodes are loaded into the same composable node container they share a process, which exposes several bugs caused by static local variables behaving as process-wide singletons instead of per-instance state.

Commit 1: IMU message pollution
FillImuData_LinearInterpolation and FillImuData_Copy each used a static local variable (_imu_history, _accel_data) to carry state between frames. These were shared across all instances, causing IMU data from one camera to corrupt the output of another.
Replaced both with a single deque member variable.

Commit 2: Shared IMU callback mutex
imu_callback_sync used a static std::mutex to serialize access to IMU state. In a composable container all instances shared the same mutex, meaning cameras serialized against each other rather than independently. Promoted to a member mutex and switched to std::lock_guard for RAII safety.

Commit 3: Race on time base initialization
The hardware-clock time base (_ros_time_base, _camera_time_base, _previous_frame_time) was initialized via atomic_bool::compare_exchange_strong from three different callbacks (frame_callback, imu_callback, imu_callback_sync). This left the non-atomic members unprotected against concurrent writes. Initialization is now consolidated into frameSystemTimeSec under a dedicated _time_base_mutex, and _is_initialized_time_base is a plain bool (protected by the same mutex).

Commit 4: _imu_history corruption on unite_imu_method change
_imu_history is shared between FillImuData_Copy and FillImuData_LinearInterpolation. When unite_imu_method is changed at runtime, stale entries from the previous mode remain in the deque. In the worst case a GYRO sample is used as accel data in COPY mode. The parameter callback now clears _imu_history under _imu_callback_mutex on every mode change.

Testing
Validated with two RealSense D436 cameras in the same composable node container with unite_imu_method set to COPY

### DIFF
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -353,7 +353,6 @@ template <typename T> T lerp(const T &a, const T &b, const double t) {
 
 void BaseRealSenseNode::FillImuData_LinearInterpolation(const CimuData imu_data, std::deque<sensor_msgs::msg::Imu>& imu_msgs)
 {
-    static std::deque<CimuData> _imu_history;
     _imu_history.push_back(imu_data);
     stream_index_pair type(imu_data.m_type);
     imu_msgs.clear();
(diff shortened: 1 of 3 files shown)


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -400,16 +399,16 @@ void BaseRealSenseNode::FillImuData_Copy(const CimuData imu_data, std::deque<sen
 {
     stream_index_pair type(imu_data.m_type);
 
-    static CimuData _accel_data(ACCEL, {0,0,0}, -1.0);
     if (ACCEL == type)
     {
-        _accel_data = imu_data;
+        _imu_history.clear();
```
Nir-Az, line 404: Why do we clear on each frame? so why do we need a deque?

Nir-Az (approved):
Great enhancement!

PR #3525, opened 2026-06-18
cmake: add ROS 2 Lyrical to supported distros

remibettan:
Tracked by: RSDSO-21542

**Overview:** Unblock the realsense2_camera build on ROS 2 Lyrical (Ubuntu 26.04 Resolute). Rolling-on-Resolute workarounds live in a separate workflow so they don't touch the main CI matrix.

## Why
The official ROS build farm started building realsense2_camera on Lyrical (Ubuntu 26.04 Resolute) and failed with `Unsupported ROS Distribution: lyrical` from `realsense2_camera/CMakeLists.txt`. The CMake distro switch hard-rejects anything outside its allow-list (foxy/humble/iron/rolling/jazzy/kilted), so Lyrical needs an explicit branch.

Failing Lyrical build: https://build.ros2.org/job/Lbin_uR64__realsense2_camera__ubuntu_resolute_amd64__binary/27/

ROS Rolling has migrated off Noble too, and several upstream gaps need transitional workarounds: `ros:rolling-ros-core` is still pinned to a Noble Dockerfile in `docker-library/official-images`, `ros-rolling-*` debs for `resolute` are only in `packages.ros.org/ros2-testing` until the next stable sync, and `image_transport` made a non-backwards-compatible API change on Rolling/Lyrical. To keep the main workflow clean, the rolling matrix entry is split into its own file with all the workarounds isolated.

## Summary
- **`realsense2_camera/CMakeLists.txt`** â€” add `elseif("$ENV{ROS_DISTRO}" STREQUAL "lyrical")` branch defining `-DLYRICAL` and pulling in `src/ros_param_backend.cpp`, mirroring the other distro entries. This is the change RSDSO-21542 originally asked for.
- **`realsense2_camera/CMakeLists.txt`** â€” gate the libfastdds load-order workaround on `fastcdr_FOUND` so it only fires when librealsense was built with bundled DDS (`BUILD_WITH_DDS=ON`). Plain (non-DDS) librealsense leaves `fastcdr_FOUND` unset, so users on Cyclone DDS / RMW Zenoh don't end up with a FastDDS NEEDED entry on the wrapper executable.
- **`realsense2_camera/src/image_publisher.cpp`** â€” Rolling/Lyrical's `image_transport::create_publisher` deduces `NodeT` from a Node reference (it calls `node.get_node_base_interface()`) and takes `rclcpp::QoS` instead of `rmw_qos_profile_t`. Guarded by `#if defined(LYRICAL) || defined(ROLLING)`; legacy distros keep the old `&node + rmw_qos_profile_t` form.
- **`.github/workflows/main.yml`** â€” reverted to the pre-PR shape. Matrix is `[humble, jazzy, kilted]` (also picks up Iron-drop from #3528). No `container_image` / `ros_distro_runtime` / AMENT overlay / ros2-testing apt-source step / rolling-skip guard.
- **`.github/workflows/main-rolling.yml`** (new) â€” standalone rolling job with all the chimera: `ros:lyrical-ros-core-resolute` container, `ROS_DISTRO=lyrical`, ros2-testing parallel apt source step, `/opt/ros/lyrical/` + `/opt/ros/rolling/` overlay in build/test steps. Integration tests intentionally not run here (they time out under the chimera â€” see header). Header lists every workaround and the upstream condition that lets each be reverted; fold the whole file back into `main.yml` once the gates close.

Follow-up ticket tracking each workaround's revert condition: RSDSO-21626.

## Test plan
- [ ] GHA `CI` matrix (humble, jazzy, kilted) passes â€” same set as today's `ros2-development`.
- [ ] GHA `CI - Rolling` job (rolling, ubuntu-26.04, lyrical-ros-core-resolute) passes build + unit tests.
- [ ] Next `build.ros2.org` `Lbin_uR64__realsense2_camera__ubuntu_resolute_amd64__binary` run succeeds.


Nir-Az:
@remibettan  please update readme tag and supported platforms on install step

.github/workflows/main.yml
```diff
@@ -25,16 +25,35 @@ jobs:
     name: Build on ROS2 ${{ matrix.ros_distro }} and ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
```
Nir-Az, line 26: Can we do the same and split rolling to a different file?

realsense2_camera/package.xml
```diff
@@ -36,6 +36,12 @@
   <depend>tf2</depend>
   <depend>tf2_ros</depend>
   <depend>diagnostic_updater</depend>
+
+  <!-- Pulls in libfastdds.so. Required at build time so CMake links it into
+       the executables (fastcdr load-order workaround, see CMakeLists.txt),
+       and at runtime so the NEEDED entry resolves. -->
+  <depend>rmw_fastrtps_cpp</depend>
```
Nir-Az: About this and the upper change in the CMakeList file.
How can we not enforce the user to link to fastdds/fastrtps if he use a different RMW in his distro?

remibettan: Gated the link on bundled-DDS detection in 5334010c. The CMake block now runs `find_library(FASTDDS_LIBRARY)` + the per-target `-Wl,--no-as-needed` only when `fastcdr_FOUND` is set.

Detection: when librealsense is built with `BUILD_WITH_DDS=ON`, its installed `realsense2Config.cmake` transitively calls `find_dependency(fastcdr CONFIG REQUIRED)` (templated in via `@REALSENSE2_DDS_DEPENDENCIES@` — see `librealsense/CMake/external_fastdds.cmake`). That makes `fastcdr_FOUND=TRUE` immediately after our `find_package(realsense2)` at line 155. A plain (non-DDS) librealsense leaves the template substitution empty, `fastcdr_FOUND` stays unset, and we skip the link entirely. So a user on Cyclone DDS / RMW Zenoh with a non-bundled-DDS librealsense doesn't get a FastDDS NEEDED entry on the wrapper executable.

The `<depend>rmw_fastrtps_cpp</depend>` here in `package.xml` stays unconditionally because it's still needed for the bundled-DDS binary case (Tier 1 deb builds from `build.ros2.org`, where librealsense ships with bundled DDS and the ABI clash is real). Non-bundled-DDS users have FastDDS installed on disk by rosdep but never loaded into the wrapper process — harmless. Making the depend itself conditional would require a build-time env var that rosdep can read, which doesn't exist for "librealsense was built with `BUILD_WITH_DDS=ON`".

.github/workflows/main.yml
remibettan: Done in ee44e62b + 7996a48f. Rolling moved to its own workflow `.github/workflows/main-rolling.yml` (single rolling-only job with the lyrical-container override, ros2-testing apt source, AMENT overlay, and integration tests intentionally not run — same set as before, just isolated). This `main.yml` is back to the pre-PR shape: matrix is `[humble, jazzy, kilted]` (also synced with #3528 dropping Iron), no `container_image` expression, no `ros_distro_runtime`, no overlay blocks, and the integration-tests step has no rolling `if:` guard. Header in `main-rolling.yml` lists every workaround it carries and the fold-back condition.

realsense2_camera/package.xml
Nir-Az: This `<depend>` is unconditional, so it pulls `rmw_fastrtps_cpp` (and FastDDS) into **every** install — including users running Cyclone DDS or RMW Zenoh, and users on plain non-DDS librealsense. That works against the `fastcdr_FOUND` gating in `CMakeLists.txt`, which deliberately avoids forcing FastDDS on exactly those users.

Can we scope this so non-FastDDS / non-DDS users aren't forced to install FastDDS — e.g. gate it the same way the CMake link is gated, or drop to a narrower dependency? Also, the comment says *"Required at build time so CMake links it into the executables"*, but the link only fires when `fastcdr_FOUND` (DDS-built librealsense); for the default non-DDS build CMake skips the link, so the comment is misleading for that case.

realsense2_camera/CMakeLists.txt
```diff
@@ -347,17 +351,42 @@ target_link_libraries(${PROJECT_NAME}
...
+# pull fastcdr in, fastcdr_FOUND stays unset, and we skip the link so users
+# on Cyclone DDS / RMW Zenoh don't end up with a FastDDS NEEDED entry on
+# the wrapper executable.
+if(fastcdr_FOUND AND UNIX AND NOT APPLE)
+  find_library(FASTDDS_LIBRARY NAMES fastdds)
+endif()
+function(_rs_link_fastdds_workaround target)
+  if(FASTDDS_LIBRARY)
+    target_link_libraries(${target}
+      "-Wl,--no-as-needed" ${FASTDDS_LIBRARY} "-Wl,--as-needed")
```
Nir-Az, line 391: `--no-as-needed` forces a hard `libfastdds` `NEEDED` entry on the wrapper executable. When librealsense is DDS-built (`fastcdr_FOUND`), every `realsense2_camera_node` then loads FastDDS at startup **regardless of the user's selected RMW** — so Cyclone DDS / Zenoh users get FastDDS force-loaded into their process. Can we confirm a force-loaded-but-unused `libfastdds` stays dormant (no participant created, no threads spun) for those users? If it's purely a load-order fix, a one-line note here saying so would settle the "other-MW users unharmed" question.

realsense2_camera/CMakeLists.txt
```diff
@@ -347,17 +351,42 @@ target_link_libraries(${PROJECT_NAME}
...
+#
+# Only applies when librealsense was built with bundled DDS (BUILD_WITH_DDS=ON
+# upstream). In that case its installed realsense2Config.cmake transitively
+# calls `find_dependency(fastcdr CONFIG REQUIRED)`, which sets fastcdr_FOUND
+# after find_package(realsense2) above. Plain (non-DDS) librealsense doesn't
+# pull fastcdr in, fastcdr_FOUND stays unset, and we skip the link so users
+# on Cyclone DDS / RMW Zenoh don't end up with a FastDDS NEEDED entry on
+# the wrapper executable.
+if(fastcdr_FOUND AND UNIX AND NOT APPLE)
+  find_library(FASTDDS_LIBRARY NAMES fastdds)
```
Nir-Az: `NAMES fastdds` only matches the library on Kilted/Lyrical/Rolling. On Humble and Jazzy the library is still `libfastrtps.so` (the package/lib was renamed `fastrtps`->`fastdds` in Kilted). Since this block is gated on `fastcdr_FOUND` (not on distro), a DDS-built librealsense on Humble/Jazzy hits `find_library` -> `NOTFOUND` and silently skips the workaround, leaving the fastcdr clash unfixed. Suggest `NAMES fastdds fastrtps`, or use the `fastdds`/`fastrtps` imported CONFIG target.

Nir-Az (commented):
Focused review on keeping non-DDS and non-FastDDS (Cyclone DDS / RMW Zenoh) users unharmed by the FastDDS load-order workaround.

realsense2_camera/CMakeLists.txt
remibettan: Good catch — fixed in 2b00ddd5. `find_library` now does `NAMES fastdds fastrtps`, so Kilted/Lyrical/Rolling match `libfastdds.so` (first hit) and Humble/Jazzy fall through to `libfastrtps.so`. Comment in the block names the rename.

realsense2_camera/CMakeLists.txt
remibettan: Added a comment in 2b00ddd5 inside `_rs_link_fastdds_workaround` saying so. Short version: the loaded-but-unused libfastdds/libfastrtps is inert. The shared object just exposes symbols — no constructors create a DomainParticipant or spin threads. FastDDS only does work once user code (here: the RMW layer) actually instantiates a participant. Cyclone/Zenoh users have ROS load `rmw_cyclonedds_cpp` / `rmw_zenoh_cpp` instead, so the FastDDS code paths are never entered. We just pay the cost of the `.so` being mapped into the address space.

realsense2_camera/package.xml
remibettan: Dropped the unconditional `<depend>rmw_fastrtps_cpp</depend>` in 2b00ddd5. Cyclone/Zenoh users and plain (non-DDS) librealsense users no longer get FastDDS pulled in by rosdep.

For bundled-DDS librealsense users: librealsense's own package is expected to make libfastdds/libfastrtps available transitively (it bundles FastDDS at build time). If for any reason the lib isn't on the system at configure time, our `find_library` returns `NOTFOUND` and the workaround silently no-ops — the bundled-DDS segfault would resurface, but only in the narrow case of bundled-DDS librealsense installed without its transitive FastDDS being reachable.

Replaced the (misleading) package.xml comment with a longer note in `CMakeLists.txt` explaining the trade-off — the package.xml gate would have been ideal, but rosdep can't conditionalise on a CMake-time variable, so the cleanest answer is to not declare it at all and rely on librealsense's own deps.

Nir-Az (approved):

nvcyc:
@Nir-Az @remibettan I noticed that the official Lyrical package distribution [list](https://github.com/ros/rosdistro/blob/bb43d8d55a79e54e929cffc0d19e6fb65559b857/lyrical/distribution.yaml#L7318) still has `realsense2_camera` listed with an old version `4.57.7-3` and hence the error "Unsupported ROS Distribution: lyrical" originally mentioned in this PR still exists in recent Lyrical [build](https://build.ros2.org/view/Lbin_uR64/job/Lbin_uR64__realsense2_camera__ubuntu_resolute_amd64__binary/105/console). 

May I know if there is any plan to update the distro list to a version that has the fix here (which currently only lives in the `ros2-development` branch I suppose) for Lyrical?

Nir-Az:
> @Nir-Az @remibettan I noticed that the official Lyrical package distribution [list](https://github.com/ros/rosdistro/blob/bb43d8d55a79e54e929cffc0d19e6fb65559b857/lyrical/distribution.yaml#L7318) still has `realsense2_camera` listed with an old version `4.57.7-3` and hence the error "Unsupported ROS Distribution: lyrical" originally mentioned in this PR still exists in recent Lyrical [build](https://build.ros2.org/view/Lbin_uR64/job/Lbin_uR64__realsense2_camera__ubuntu_resolute_amd64__binary/105/console).
> 
> May I know if there is any plan to update the distro list to a version that has the fix here (which currently only lives in the `ros2-development` branch I suppose) for Lyrical?

Hi @nvcyc , firstly great to hear users care and waits for it :)
It surely in the plans, we didn't publish Lyrical yet so I don't know how it got there :), probably clones from another distro when Lyrical was created
We plan to do it in the coming 58.4 release that should be pushed in the few coming weeks.
Is that sounds like a reasonable timeline for you?

nvcyc:
@Nir-Az Thanks! Really appreciate your update!
It is great to hear that a release for Lyrical (and to the buildfarm) is coming.
Lyrical users and developers certainly can't wait to try out things that depend on Realsense in Lyrical as early as possible!
I think an early to mid August release time frame would make sense considering that 2026 ROSCon Global is taking place in September and people will be eager to try out Lyrical ahead.

Issue #1370, opened 2020-09-08
L515 |  time synchronization issue between IMU and image streams 


lvj5077:
Hi,
We are experiencing time synchronization issue between IMU and image streams.
The timestamps among images are synchronized while there are always some seconds delay between IMU and image's timestamp. Also, the temporal offset is not a constant and changes every time.

In addition, when I turn off the "global timestamp" function, the IMU timestamp is switched to the internal clock while the image streams are still using the timestamp the same as the ROS.

bigbellmercy:
This is another similar report for IMU time sync problem of L515.

### Problem
The timestamps of gyro and accel increase of itself continuously. The rate is about 2 ~ 4 ms per second.

### Environment
Ubuntu 18.04 kernel **5.4.0-47**-generic
libRealSense v2.37.0
realsense-ros wrapper v2.2.16
L515 fw v1.5.0

### Reproducing

1. Install `librealsense2` debian package using `apt-get` as [Installing the packages](https://github.com/IntelRealSense/librealsense/blob/master/doc/distribution_linux.md#installing-the-packages)
2. Install `realsense-ros` by `apt-get`
3. Edit `rs_camera.launch` in the `launch` folder of `realsense-ros`:
- Change `enable_gyro` to `true`
- Change `enable_accel` to `true`
- Change `accel_fps` to one of 100/200/400 Hz (because default 250 Hz is not supported for L515)
4. Connect L515 and Run `roslaunch realsense2_camera rs_camera.launch`
5. Check delay time:
`rostopic delay /camera/gyro/sample`
`rostopic delay /camera/accel/sample`

### Comment
1. According to [Linux Distribution](https://github.com/IntelRealSense/librealsense/blob/master/doc/distribution_linux.md#using-pre-build-packages), my Ubuntu kernel v5.4.0 is not listed as following:
> The Realsense DKMS kernel drivers package (librealsense2-dkms) supports Ubuntu LTS kernels 4.4, 4.8, 4.10, 4.13, 4.15, 4.18* 5.0* and 5.3*. 
2. To try to bypass the problem, if the dynamic parameter `global_time_enabled` is set to `true` for IMU(gyro and accel) and detph image, the timestamp value is doubled (twice bigger than current system time). (The issue [1370](https://github.com/IntelRealSense/realsense-ros/issues/1273) is similar.)

lvj5077:
Hi, @RealSenseCustomerSupport
[...]

PR #3526, opened 2026-06-24
Make live-camera device-name parsing vendor-prefix-agnostic

Nir-Az:
**Overview:** The ROS CI nightly (`LRS_linux_compile_ros_ci` #1370) went UNSTABLE with all 12 live-camera tests failing at their `check_if_camera_connected()` precheck (`assert False`). Root cause is librealsense PR #15178 ("Drop 'Intel' prefix from camera visible names"), which changed `rs-enumerate-devices` output from `Intel RealSense D455` to `RealSense D455`. Two parsers in `pytest_live_camera_utils.py` hard-coded the old `Intel RealSense <model>` layout and no longer recognize any camera, so every live-camera test bails in its connectivity precheck.

**Changes:**
- `check_if_camera_connected` (`-s` output): match the model token anywhere in the line instead of requiring `name_line[0] == "Intel"` and the model at a fixed column; the serial is taken as the token right after the model.
- `parse_device_info` (`-v` output): match `device_type` among the `Name :` value tokens instead of the fixed index 4 (which now reads the wrong token / out of range). Removed the now-unused `NAME_LINE_NAME_OFFSET`.
- Removed all `Intel` references from the test files: reworded the explanatory comments, and updated the `camera_n_mqtt_nodes.py` docstring sample output to the rebranded `RealSense D585S` format.

Both parsers now accept device names with or without the vendor prefix, so they work against pre- and post-#15178 librealsense.

🤖 Generated by AI

### DIFF
--- a/realsense2_camera/test/utils/pytest_live_camera_utils.py
+++ b/realsense2_camera/test/utils/pytest_live_camera_utils.py
@@ -122,6 +122,7 @@ def get_color_profiles(long_data, start_index, end_index):
     return cap
 
 NAME_LINE_INDEX = 1
+NAME_LINE_VALUE_OFFSET = 2  # tokens after the "Name :" prefix
 SERIAL_NO_LINE_INDEX = 2
 SERIAL_NO_VALUE_OFFSET = 3
 def parse_device_info(long_data, start_index, end_index, device_type, serial_no):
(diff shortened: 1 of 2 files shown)


realsense2_camera/test/utils/pytest_live_camera_utils.py
```diff
@@ -132,10 +131,14 @@ def parse_device_info(long_data, start_index, end_index, device_type, serial_no)
     name_line = long_data[start_index+NAME_LINE_INDEX].split()
     if name_line[0] != "Name":
         assert False, "rs-enumerate-devices output format changed"
-    if name_line[4] != device_type:
-        debug_print("device not matching:" + name_line[NAME_LINE_NAME_OFFSET])
+    # Name value tokens follow the "Name :" prefix. The model token (e.g.
+    # "D455") is present whether or not a vendor prefix is shown, so match
+    # on it directly.
+    name_value_tokens = name_line[2:]
```
remibettan: please replace this 2: by some constant

Nir-Az: Done — added `NAME_LINE_VALUE_OFFSET = 2` and used it here instead of the literal.

remibettan (approved):

PR #3530, opened 2026-06-25
Add power line parameter to avoid warning in launch file

cychitivav:
image

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -37,6 +37,7 @@ configurable_parameters = [{'name': 'camera_name',                  'default': '
                            {'name': 'rgb_camera.color_profile',     'default': '0,0,0', 'description': 'color stream profile'},
                            {'name': 'rgb_camera.color_format',      'default': 'RGB8', 'description': 'color stream format'},
                            {'name': 'rgb_camera.enable_auto_exposure', 'default': 'true', 'description': 'enable/disable auto exposure for color image'},
+                           {'name': 'rgb_camera.power_line_frequency', 'default': '2', 'description': 'power line frequency for color image. 0=Disabled, 1=50Hz, 2=60Hz'},
                            {'name': 'enable_depth',                 'default': 'true', 'description': 'enable depth stream'},
                            {'name': 'enable_infra',                 'default': 'false', 'description': 'enable infra0 stream'},
                            {'name': 'enable_infra1',                'default': 'false', 'description': 'enable infra1 stream'},


sysrsbuild-gh:

Nir-Az:
@cychitivav  thanks for the PR.
Any idea why the default value to be set is 3?
And why the new is 2?

regardless, please base your PE on ros2-development branch and target it, master is the stable release branch
Thanks!

MartyG-RealSense:
Hi @Nir-Az I just wanted to comment that the power line parameter warning in the launch log has been present for years and occurs in many users' logs but has no negative effects.  Users do occasionally ask about it though out of concern that it might be a problem.

 Setting number 3 is Auto.  I'm not certain how it decides what the appropriate value to use is though.  The other settings are Disabled (0), 50 Hz (1) and 60 Hz (2).  50 Hz is common for European regions and 60 Hz is typically used in North American regions.

The most common reason for configuring this parameter manually is to reduce noise interference on the depth image when using the camera in a location with overhead fluorescent ceiling strip lights, which generate image noise due to flickering at rates difficult to see with the human eye.  Matching the power line frequency setting of the camera as closely as possible to the electrical power Hz frequency used for lights in the country the camera is located in can reduce fluorescent light noise.

Support does not to get any reports of users with fluorescent lights these days though as LED ceiling lights have become much more common and they to not produce flicker noise.

I would recommend leaving the setting on 3 (Auto) and just removing the power line frequency warning.

Nir-Az:
@MartyG-RealSense thanks for the info, we will evaluate this soon

remibettan:
Hi @cychitivav, thanks for the PR — a few things to share before we merge anything.

Where the warning actually comes from. The camera's factory default for POWER_LINE_FREQUENCY is 3 (Auto). On startup, the ROS node reads that value and tries to declare a ROS parameter for it, using the option range reported by the UVC driver as the descriptor's integer range. On a stock Linux kernel, the UVC subsystem's power_line_frequency_controls menu only lists {0: Disabled, 1: 50Hz, 2: 60Hz} — no Auto — so the range comes back as [0, 2]. rclcpp then rejects the initial value 3 because it's out of range, which is exactly the "Could not set param: rgb_camera.power_line_frequency with 3 Range: [0, 2]" warning you're seeing.

Why we don't want to fix this by defaulting to 2. Setting the launch default to 2 does silence the warning, but not by fixing the underlying mismatch — it silences it by forcing every user to 60Hz on startup. That's a regression for users in 50Hz-mains regions (most of Europe/Asia/etc.), and it silently overrides the camera's own Auto default. We'd rather not change observable behavior for every user of the launch file just to hide a log line.

The recommended fix on the user side. Install librealsense2-dkms from the RealSense apt repo. The DKMS package includes a small kernel patch (realsense-powerlinefrequency-control-fix.patch) that adds {3, "Auto"} to the UVC menu. With that in place, the driver reports the option range as [0, 3], the FW default of Auto validates cleanly, and the warning disappears — without any launch-file change and without altering the effective power-line-frequency setting.

# 1. Add the RealSense apt repo (skip if already configured)
sudo mkdir -p /etc/apt/keyrings
curl -sSf https://librealsense.realsenseai.com/Debian/librealsenseai.asc \
  | gpg --dearmor | sudo tee /etc/apt/keyrings/librealsenseai.gpg > /dev/null
sudo apt-get install -y apt-transport-https
echo "deb [signed-by=/etc/apt/keyrings/librealsenseai.gpg] https://librealsense.realsenseai.com/Debian/apt-repo $(lsb_release -cs) main" \
  | sudo tee /etc/apt/sources.list.d/librealsense.list
sudo apt-get update

# 2. Install the DKMS package (rebuilds the patched uvcvideo module for your kernel)
sudo apt-get install -y librealsense2-dkms

# 3. Reboot (or reload uvcvideo) and verify the patched module is loaded
sudo modprobe -r uvcvideo && sudo modprobe uvcvideo
modinfo uvcvideo | grep -i version   # should contain "realsense"

Supported HWE kernels for the DKMS package are listed in the RealSense Linux distribution doc; for other kernels the same patch can be applied manually via scripts/patch-realsense-ubuntu-*.sh.

Given the above, could we close this PR? Happy to keep it open if you want to propose any other alternative.

---------------------------------------

Also, thanks @MartyG-RealSense for the historical context — very helpful.
On your suggestion to just remove the warning: we'd actually rather leave it in place. It's a useful (if noisy) signal that the user's kernel is missing our UVC patch — the whole reason [0, 2] shows up instead of [0, 3] is that the stock uvcvideo menu is incomplete. Silencing it (either via a launch-file default or a node-side suppression) would hide that real mismatch between what the FW supports and what the driver exposes, and users would lose the one hint pointing them at librealsense2-dkms. So we prefer to keep the warning  visible and treat it as a "you're missing the kernel patch" indicator rather than something to paper over.

Nir-Az:
@cychitivav  thanks for the PR, since no response we will close it for now.
If anything is still needed feel free to reopen and addressing the review by @remibettan

PR #3534, opened 2026-07-01
RSDSO-21561: Add D421 URDF model and mesh to realsense2_description

Nir-Az:
**Overview:** Adds a URDF model and visual mesh for the RealSense Depth Module **D421** to `realsense2_description`, so the D421 can be visualized/simulated in ROS 2 like the other D4xx devices.

Jira: [RSDSO-21561](https://rsjira.realsenseai.com/browse/RSDSO-21561)

### Changes
- `urdf/_d421.urdf.xacro` — stand-alone `sensor_d421` macro. The D421 is a **bare single-board module** (no enclosure, no cover glass, no tripod/bottom screw), so it does **not** reuse the cased-camera terms. Dual IR imagers + IR projector, **no** RGB/IMU; 50 mm baseline; dimensions 95 × 32 × 10.18 mm; mass 0.0375 kg.
- `urdf/test_d421_camera.urdf.xacro` — standalone test wrapper (auto-discovered by `view_model.launch.py`).
- `meshes/d421.stl` — visual mesh of the flat board (two imagers + projector).

No `CMakeLists.txt`/launch changes needed — `urdf/` and `meshes/` are glob-installed and the view launcher auto-lists any `test_*.urdf.xacro`.

### Frames
- **`camera_mount_frame`** — model origin, the **centre of the back panel** (back surface). A module is fastened flat against the user's rig, so this is the natural, physically-locatable attachment point.
- **`camera_link`** — `camera_link`/depth origin at the centre of the **left (infra1) imager** on the front face, derived from the mount frame (forward by the board thickness, left by half the baseline). The ROS driver anchors its TF tree here, so it must stay on the left imager.
- `infra2` sits one 50 mm baseline to the right.

**Data source:** D421 Datasheet v1.1 + Leopard LI-DS5-OV9281 mechanical drawing.

### Validation (ROS 2 Humble)
`xacro` processes the model cleanly (exit 0), and it loads and renders in **rviz2** with `camera_mount_frame` as the fixed frame. Frame tree confirmed: `camera_mount_frame → camera_link [0.01018, 0.025, 0] → {depth, infra1 @ 0, infra2 @ -0.050}`.

> **Note:** the mesh is a simplified board representation, to be replaced with a detailed CAD/STEP export when available. Board thickness is **10.18 mm** from the mechanical drawing (the datasheet's two tables list 10.2 and 10.5 mm).

### Render (rviz2)
D421 model in rviz2 — grey board with the two IR imagers and USB-C plug, TF axes with the origin at the back-panel centre:

D421 in rviz2

🤖 Generated by AI


realsense2_description/urdf/_d421.urdf.xacro
```diff
@@ -0,0 +1,154 @@
...
+    <xacro:property name="d421_cam_depth_to_infra2_offset" value="-0.050"/>
+
+    <!-- The following values model the D421 single-board module. The module has
+        no tripod mount, so the module datum (bottom_screw_frame) is taken at the
+        board's geometric centre, which is horizontally centred between the two
+        imagers. Per the datasheet the depth ground-zero is the left imager
+        centre, and the left centre mounting hole sits 18.5 mm from it. -->
+    <xacro:property name="d421_cam_width" value="0.095"/>
+    <xacro:property name="d421_cam_height" value="0.032"/>
+    <xacro:property name="d421_cam_depth" value="0.0102"/>
```
Gilaadb: The docs mention different values for this: 10.12 mm, 10.2mm, 10.5mm need to check what is the correct value.

realsense2_description/urdf/_d421.urdf.xacro
```diff
@@ -0,0 +1,154 @@
...
+    <!-- module datum at board centre, so the front face is half the board depth forward -->
+    <xacro:property name="d421_cam_mount_from_center_offset" value="0.0051"/>
```
Gilaadb: Better use d421_cam_depth/2 since this is what it is by definition.
But actually, I don't know what this value is supposed to be used for, since in D435, D455 it would represent the distance from the center of the bottom mounting screw to the front edge of the camera. Here we don't have either.

realsense2_description/urdf/_d421.urdf.xacro
```diff
@@ -0,0 +1,154 @@
...
+    <!-- glass cover is 0.1 mm inwards from front plate -->
+    <xacro:property name="d421_glass_to_front" value="0.1e-3"/>
```
Gilaadb: What glass? Where this info was it taken from?

realsense2_description/urdf/_d421.urdf.xacro
```diff
@@ -0,0 +1,154 @@
...
+    <!-- approximate distance from depth ground-zero to front glass (update from mechanical drawing) -->
+    <xacro:property name="d421_zero_depth_to_glass" value="4.2e-3"/>
```
Gilaadb: I didn't see this value in the spec.

Gilaadb (changes requested):
The URDF followed D455/D435 family of cameras too stricktly. It describes sizes that don't physically exist in this module. For example, there is no bottom screw. it makes no sense to expect a user to use a non existing bottom screw as the attachment point of the camera's URDF to their rig/robot design. My suggestion is to rewrite this file as a stand-alone and not copy unrelated terms from other cased cameras.
To me, it makes most sense to define the origin as the center of the back plate, on he back surface of it. A user can easily derive it from where they attach the camera to their rig/robot.

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: The mechanical drawing (Leopard LI-DS5) shows 10.18 mm; datasheet v1.1 lists 10.2 mm (features table) and 10.5 mm (properties table). I used 10.2 mm here as it matches the mechanical drawing (10.18 rounded). Can you confirm which is the authoritative value?

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: Removed. A module has no bottom screw, so this offset had no physical meaning. The model origin is now the centre of the back panel; `cam_depth/2` is used only where it genuinely is half the board thickness (the collision box).

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: Removed - the D421 is a bare module with no cover glass. This term was copied from the cased-camera files by mistake.

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: Removed for the same reason - it isn't in the spec and doesn't apply to a glassless module.

Nir-Az:
Rewrote `_d421.urdf.xacro` stand-alone, dropping the cased-camera terms that don't exist on this module (cover glass, bottom-screw / mount-from-centre offsets, `bottom_screw_frame`).

As suggested, the model origin is now the **centre of the back panel** (`camera_mount_frame`) - the flat surface a user actually mounts against. `camera_link` (the left-imager depth origin, where the ROS driver anchors the TF tree) is derived from it: forward by the board thickness and left by half the 50 mm baseline. Only physically-real values remain (95 x 32 x 10.2 mm, 50 mm baseline).

Validated on a Linux box with ROS 2 Humble: `xacro` processes cleanly and the model renders in rviz2 (see updated PR description). Please see the per-comment replies above; one open question on the exact board thickness.

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: Confirmed with Nir - going with 10.18 mm from the mechanical drawing. Updated `d421_cam_depth` to 0.01018 and regenerated the mesh to match.

realsense2_description/urdf/_d421.urdf.xacro
```diff
@@ -0,0 +1,153 @@
...
+        <parent link="${name}_infra2_frame" />
+        <child link="${name}_infra2_optical_frame" />
+      </joint>
+      <link name="${name}_infra2_optical_frame"/>
+    </xacro:if>
+
+    <xacro:if value="$(arg add_plug)">
+      <!-- USB-C connector on the bottom edge, board-centred -->
+      <xacro:usb_plug parent="${name}_link" name="${name}_usb_plug">
+        <origin xyz="${-d421_cam_depth/2} ${-d421_baseline/2} ${-d421_cam_height/2}" rpy="${M_PI/2} 0 0"/>
```
Gilaadb: Don't know if anyone cares, but this is not the location of the usb connector. it is not in the middle of the camera.

realsense2_description/urdf/_d421.urdf.xacro
```diff
@@ -0,0 +1,153 @@
...
+      <xacro:insert_block name="origin" />
+      <parent link="${parent}"/>
+      <child link="${name}_mount_frame" />
+    </joint>
+    <link name="${name}_mount_frame"/>
+
+    <!-- camera_link at the depth origin (left imager, front face): forward from
+         the back-panel centre by the board thickness, and left by half the baseline -->
+    <joint name="${name}_link_joint" type="fixed">
+      <origin xyz="${d421_cam_depth} ${d421_baseline/2} 0" rpy="0 0 0"/>
```
Gilaadb, line 67: For the x value, we need to accommodate the offset of the Z=0 origin. 
We need to substract z' value specific for this module. I'm working on finding it out.
Image

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: Good catch. The plug was board-centred as a placeholder and only rendered when `add_plug:=true` (off by default). Rather than ship a wrong location, I removed the `usb_plug` block entirely (also dropped the now-unused `add_plug` arg and `_usb_plug` include).

✅ Fixed in 66f4938.

🤖 Generated by AI

realsense2_description/urdf/_d421.urdf.xacro
Nir-Az: Agreed — `camera_link` should sit at the imager plane, not the front face, so `x = d421_cam_depth - z'`. Right now it uses the full board thickness (front face). Once you have `z'` for this module I will update the `${name}_link_joint` origin (and the mesh offset if needed) to match. Ping me with the value.

⏸ Deferred — waiting on the module-specific `z'` offset.

🤖 Generated by AI

Gilaadb: What i've been told is that since D421 module doesn't have a stiffener then z' = 0.
It means the current calculation is accurate.

Gilaadb (approved):
lgtm

PR #3536, opened 2026-07-08
Ngoren/occupancy gridcells fix

nivgo:
# Occupancy Grid — How It Works

## 1. Overall Pipeline

```mermaid
flowchart TD
    FW["🔲 D585S Firmware\npacked bitmask — 1 bit per cell\n(1=occupied, 0=not occupied)"]

    subgraph fn["publishOccupancyFrame()"]
        INIT["Init every cell → -1\n(unknown)"]
        FOV{"FOV cone check\n|y| ≥ x · tan(hfov/2) ?"}
        RANGE{"Max range check\nx > occupancy_max_range ?"}
        BIN["Assign to angular bin\nθ = atan2(y, x)"]
        RAY["Scan ray nearest → farthest\nwith per-bin occlusion state"]
        OCC{"Occupied\nbit set?"}
        SHAD{"Ray already\noccluded?"}
        SPREAD["Spread shadow to\n± n_spread neighbour bins\nn_spread = ceil(cell_size·N_bins / 2·x·bin_range)"]
    end

    FREE["0\nconfirmed free"]
    OB["100\noccupied"]
    UNK["-1\nunknown"]
    OUT["📦 nav_msgs/OccupancyGrid"]

    FW --> INIT --> FOV
    FOV -->|outside camera FOV| UNK
    FOV -->|inside FOV| RANGE
    RANGE -->|beyond 2.5 m default| UNK
    RANGE -->|within range| BIN --> RAY --> OCC
    OCC -->|yes| OB --> SPREAD
    OCC -->|no| SHAD
    SHAD -->|yes| UNK
    SHAD -->|no| FREE
    SPREAD -->|"neighbours → -1"| UNK
    FREE --> OUT
    OB  --> OUT
    UNK --> OUT
```

---

## 2. Cell Value Semantics

```mermaid
flowchart LR
    subgraph legend["Cell value in OccupancyGrid.data[]"]
        direction TB
        A["0\n✅ confirmed free\nRay passed through — no obstacle seen"]
        B["100\n🔴 occupied\nFirmware bit = 1"]
        C["-1\n❓ unknown\nOutside FOV, beyond max range,\nor shadowed behind an obstacle"]
    end
```

---

## 3. Live Output

Real grid published by the driver, shown in rviz with the live point cloud overlaid.
White = free (`0`), gray = unknown (`-1`), occupied cells (`100`) sit under the detected boxes —
note the FOV wedge, the shadows fanning out behind each obstacle, and the unknown
region beyond `occupancy_max_range`:

Occupancy grid live output
occupancy_new_type

---

## 4. Column-Based vs True Angular Raycasting

```
OLD — Column-based (wrong)          NEW — Angular raycasting (correct)
─────────────────────────────       ─────────────────────────────────────
Camera at bottom (x=0)             Camera at bottom (x=0)

      Y=0.5m fixed                        angle θ fixed
         │                                    ╱
x=0.5m  ●                          x=0.5m  ●  (y=0.5m)
         │                                    ╲
x=1.0m  ●                          x=1.0m    ●  (y=1.0m)
         │                                      ╲
x=2.0m  ●                          x=2.0m        ●  (y=2.0m)
         │
x=3.0m  ●                          Shadow FANS outward
                                    y grows with depth: y = x · tan(θ)
Shadow is a vertical stripe         ✔ physically correct
✘ wrong at wide angles
```

---

## 5. Shadow Propagation Along a Ray

```
Ray (single angular bin), scanning nearest → farthest:

  Camera
    │
    ▼
[ free  ]  ← no obstacle yet, clear line of sight  → cell = 0
[ free  ]
[ free  ]
[BLOCKED]  ← firmware bit = 1                       → cell = 100  ← occlusion starts here
[shadow ]  ← ray is occluded                        → cell = -1
[shadow ]
[shadow ]  ← stays unknown all the way to max range → cell = -1
```

---

## 6. Gap-Ray Fix (Shadow Spreading)

Without the fix, a ray can slip through the gap between two adjacent occupied cells
that fall in different angular bins:

```
BEFORE gap-ray fix                    AFTER gap-ray fix
                                      (n_spread bins on each side)

Angular bins:  …23  24  25  26…       Angular bins:  …23  24  25  26…
                    │   │                                  │   │
depth=near          ■   ■  ← 2 occupied cells             ■   ■
depth=mid       0   0   0   0    ← gap ray free!      -1  -1  -1  -1  ← all shadowed
depth=far       0   0   0   0                         -1  -1  -1  -1

■ = occupied   0 = free   -1 = unknown/shadow

n_spread = ceil( cell_size × N_bins / (2 × x_obstacle × bin_range) )
         = large at close range (wide angular footprint)
         = 1    at far range   (narrow angular footprint)
```

---

## 7. FOV Masking

```
Top-down view of the grid (camera at bottom centre):

        ╔═══════════════════════╗   ← far edge (x = max_range)
        ║  -1  -1 │  0   0 │-1 -1║  ← -1 outside FOV cone
        ║  -1  -1 │  0   0 │-1 -1║
        ║  -1   0 │  0   0 │ 0 -1║
        ║   0   0 │100 100 │ 0  0║  ← obstacles
        ║   0   0 │ -1  -1 │ 0  0║  ← shadow behind obstacles
        ║   0   0 │ -1  -1 │ 0  0║
        ╚═══════════════════════╝   ← near edge (x ≈ 0)
                   camera

  FOV cone boundary: |y| = x · tan(hfov/2)
  Anything outside the cone → -1 (unknown)
  Anything beyond max_range → -1 (unknown)
```

---

## 8. Coordinate Mapping: Firmware → OccupancyGrid

```
Firmware layout                    OccupancyGrid layout (ROS convention)
(fw_row=0 = farthest)              (col_idx=0 = nearest, X forward)

fw_row=0   [col 0..cols-1]    →   og_col_idx = width-1-fw_row  (maps far→high X)
fw_row=1   [col 0..cols-1]         og_row_idx = cols-1-fw_col  (maps left→high Y)
...
fw_row=rows-1 [col 0..cols-1] →   data[og_row_idx * width + og_col_idx]

Firmware "cols" = lateral (Y axis) = msg.info.height
Firmware "rows" = depth   (X axis) = msg.info.width
```

### DIFF
--- a/realsense2_camera/launch/rs_launch.py
+++ b/realsense2_camera/launch/rs_launch.py
@@ -68,6 +68,7 @@ configurable_parameters = [{'name': 'camera_name',                  'default': '
                            {'name': 'motion_fps',                   'default': '0', 'description': "'motion stream samples per second'"},
                            {'name': 'unite_imu_method',             'default': "0", 'description': '[0-None, 1-copy, 2-linear_interpolation]'},
                            {'name': 'clip_distance',                'default': '-2.', 'description': "''"},
+                           {'name': 'occupancy_max_range',          'default': '2.5', 'description': "'max reliable range for occupancy raycasting in meters; cells beyond this are unknown (-1)'"},
                            {'name': 'angular_velocity_cov',         'default': '0.01', 'description': "''"},
                            {'name': 'linear_accel_cov',             'default': '0.01', 'description': "''"},
                            {'name': 'diagnostics_period',           'default': '0.0', 'description': 'Rate of publishing diagnostics. 0=Disabled'},
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -922,41 +922,52 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time& 
     if(!_occupancy_publisher || 0 == _occupancy_publisher->get_subscription_count())
         return;
 
-    ROS_DEBUG("Publishing Occupancy GridCells Frame");
+    ROS_DEBUG("Publishing Occupancy Grid Frame");
 
-    // get frame bytes and frame metadata relevant info
     auto frame_as_uint8_arr = (uint8_t*)f.get_data();
-    auto cols = static_cast<int>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_GRID_COLUMNS)); // grid cells width
-    auto rows = static_cast<int>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_GRID_ROWS)); // grid cells height
-    auto cell_size = static_cast<float>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_CELL_SIZE) / 100.0f); // convert to meters
+    auto cols = static_cast<int>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_GRID_COLUMNS));
+    auto rows = static_cast<int>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_GRID_ROWS));
+    auto cell_size = static_cast<float>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_CELL_SIZE) / 100.0f); // cm -> m
 
-    // create GridCells msg and start filling it
-    nav_msgs::msg::GridCells msg;
+    nav_msgs::msg::OccupancyGrid msg;
     msg.header.stamp = t;
     msg.header.frame_id = FRAME_ID(OCCUPANCY);
-    msg.cell_width = cell_size;
-    msg.cell_height = cell_size;
+
+    // Grid metadata — self-describing so subscribers need no out-of-band knowledge.
+    // Axes follow ROS convention (X: Forward, Y: Left):
+    //   width  = cells along X (forward) = rows in firmware
+    //   height = cells along Y (left)    = cols in firmware
+    // Origin is the lower-left corner of cell (width-1, 0) in map frame:
+    //   origin.x = 0           (nearest boundary, X direction)
+    //   origin.y = -resolution * height / 2  (rightmost boundary, Y direction)
+    msg.info.map_load_time = t;
+    msg.info.resolution = cell_size;
+    msg.info.width  = static_cast<uint32_t>(rows);
+    msg.info.height = static_cast<uint32_t>(cols);
+    msg.info.origin.position.x = 0.0;
+    msg.info.origin.position.y = -cell_size * static_cast<float>(cols) / 2.0f;
+    msg.info.origin.position.z = 0.0;
+    msg.info.origin.orientation.w = 1.0;
+
+    // data[row_idx * width + col_idx], 0 = free, 100 = occupied.
+    // col_idx increases in +X (forward), row_idx increases in +Y (left).
+    // Firmware row 0 = farthest  → col_idx = width-1-fw_row
+    // Firmware col 0 = leftmost  → row_idx = height-1-fw_col
+    msg.data.assign(rows * cols, 0);
 
     for (auto i = 0; i < cols * rows; ++i)
     {
-        // AICV algo is packing each 8 cells into one byte. Each byte include 8 bits <--> 8 cells
-        // The rightest bit (LSB) inside the packed byte from AICV algo represnts the closest cell we want to work with in the grid.
-        // e.g. Original Occupancy Cells: 0 0 1 1 0 0 1 0 ---> AICV packing algo ---> 01001100 (not the opposite order)
-        // In this if we check if current cell is occupied.
-        // Note that we start working from the most left bit, aka, the farest point of the grid.
+        // AICV algo packs 8 cells per byte; bit i%8 of byte i/8 = cell i.
         if ((frame_as_uint8_arr[i / 8U] & (1U << i % 8)) != 0)
         {
-            // Find x,y,z positions of current index
-            // Remember, in ROS CS: (X: Forward, Y: Left, Z: Up)
-            geometry_msgs::msg::Point p3d;
-            uint32_t row = (i / cols);
-            uint32_t col = (i % cols);
-            p3d.x = (cell_size * static_cast<float>(rows)) - cell_size * (static_cast<float>(row) + 0.5f);
-            p3d.y = (cell_size * static_cast<float>(cols)) / 2 - cell_size * (static_cast<float>(col) + 0.5f);
-            p3d.z = 0;
-            msg.cells.push_back(p3d);
+            uint32_t fw_row = static_cast<uint32_t>(i / cols);
+            uint32_t fw_col = static_cast<uint32_t>(i % cols);
+            uint32_t og_col_idx = static_cast<uint32_t>(rows) - 1u - fw_row;
+            uint32_t og_row_idx = static_cast<uint32_t>(cols) - 1u - fw_col;
+            msg.data[og_row_idx * static_cast<uint32_t>(rows) + og_col_idx] = 100;
         }
     }
+
     _occupancy_publisher->publish(msg);
 }
 
--- a/realsense2_camera/src/base_realsense_node.cpp
+++ b/realsense2_camera/src/base_realsense_node.cpp
@@ -929,7 +929,8 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time& 
     // The FOV mask and ray binning are meaningless without it - drop the frame
     // rather than publish a grid built on incomplete information.
     const auto depth_info_it = _camera_info.find(DEPTH);
-    if (depth_info_it == _camera_info.end() || depth_info_it->second.k.at(0) <= 0.0)
+    if (depth_info_it == _camera_info.end() || depth_info_it->second.k.at(0) <= 0.0
+        || depth_info_it->second.width == 0)
     {
         ROS_WARN("Occupancy grid not published: depth stream intrinsics are not available");
         return;
@@ -950,14 +951,15 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time& 
     //   width  = cells along X = firmware rows
     //   height = cells along Y = firmware cols
     // Origin is the corner of cell (0,0): nearest boundary in X, rightmost in Y.
-    // (cols - cols/2) instead of cols/2 keeps origin.y aligned with the integer
-    // division used in the cell coordinates below when cols is odd.
+    // The grid is symmetric about the camera axis, so the right boundary is at
+    // y = -cols/2 cells regardless of parity; the per-cell y below uses the
+    // same convention.
     msg.info.map_load_time = t;
     msg.info.resolution = cell_size;
     msg.info.width  = static_cast<uint32_t>(rows);
     msg.info.height = static_cast<uint32_t>(cols);
     msg.info.origin.position.x = 0.0;
-    msg.info.origin.position.y = -cell_size * static_cast<float>(cols - cols / 2);
+    msg.info.origin.position.y = -cell_size * static_cast<float>(cols) * 0.5f;
     msg.info.origin.position.z = 0.0;
     msg.info.origin.orientation.w = 1.0;
 
@@ -972,7 +974,7 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time& 
     const auto width = msg.info.width;
 
     const float half_fov_rad = std::atan(tan_half_hfov);
-    const float bin_range = 2.0f * half_fov_rad;
+    const float fov_span = 2.0f * half_fov_rad; // total angular window covered by the bins
 
     // Full grid extent, unless limited by occupancy_max_range.
     const float x_far = (static_cast<float>(rows) - 0.5f) * cell_size;
@@ -980,7 +982,7 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time& 
 
     // Enough angular bins to resolve one cell width at the farthest depth.
     const int N_bins = std::max(cols,
-        static_cast<int>(std::ceil(x_far * bin_range / cell_size)) + 1);
+        static_cast<int>(std::ceil(x_far * fov_span / cell_size)) + 1);
 
     // Rows are scanned nearest-first while bin_occluded tracks which rays are
     // already blocked. Occlusion is applied only after a row completes, so cells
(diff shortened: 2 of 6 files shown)


sysrsbuild-gh:

Gilaadb (approved):

Nir-Az:
@remibettan please help reviewing this PR, I also suggest running it

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
         return;
 
-    ROS_DEBUG("Publishing Occupancy GridCells Frame");
+    ROS_DEBUG("Publishing Occupancy Grid Frame");
+
+    // Horizontal FOV from depth intrinsics: tan(half_hfov) = (width/2) / fx.
+    // The FOV mask and ray binning are meaningless without it - drop the frame
+    // rather than publish a grid built on incomplete information.
+    const auto depth_info_it = _camera_info.find(DEPTH);
+    if (depth_info_it == _camera_info.end() || depth_info_it->second.k.at(0) <= 0.0)
```
remibettan: Also guard on `depth_info_it->second.width > 0` here. If `fx > 0` but `width == 0`, `tan_half_hfov` becomes 0, the FOV test `|y| >= 0` trips for every cell, and the grid is silently published entirely as `-1` with no warning. Adding the width check makes the existing `ROS_WARN` fire instead.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
-            // Remember, in ROS CS: (X: Forward, Y: Left, Z: Up)
-            geometry_msgs::msg::Point p3d;
-            uint32_t row = (i / cols);
-            uint32_t col = (i % cols);
-            p3d.x = (cell_size * static_cast<float>(rows)) - cell_size * (static_cast<float>(row) + 0.5f);
-            p3d.y = (cell_size * static_cast<float>(cols)) / 2 - cell_size * (static_cast<float>(col) + 0.5f);
-            p3d.z = 0;
-            msg.cells.push_back(p3d);
+            // Integer cols/2, consistent with origin.y above.
+            const float y = (static_cast<float>(cols / 2) -
```
remibettan: For odd `cols`, this formula shifts `y = 0` by half a cell relative to the previous code. Old: `(cols/2) * cs - (col + 0.5) * cs` — for `cols = 7`, `col = 3` gave `y = 0` at the geometric center of the middle cell. New: for `cols = 7`, `fw_col = 3` gives `y = -0.5 * cs`. The origin and cell coords are self-consistent internally (I checked), but any downstream consumer relying on the old convention would see a half-cell lateral offset. Worth confirming firmware always emits even `cols`; if odd is possible, the shift should probably be called out or removed.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
+                             static_cast<float>(fw_col) - 0.5f) * cell_size;
+            if (std::fabs(y) >= x * tan_half_hfov) continue; // outside FOV
+
+            const float theta = std::atan2(y, x);
+            const int bin = std::min(
+                static_cast<int>((theta + half_fov_rad) / bin_range * static_cast<float>(N_bins)),
+                N_bins - 1);
+            if (bin < 0) continue;
```
remibettan: Dead. The FOV mask at line 1005 guarantees `|theta| < half_fov_rad`, so `(theta + half_fov_rad) / bin_range * N_bins` is strictly in `(0, N_bins)` and the `std::min(..., N_bins - 1)` handles the upper bound. `bin` can't be negative here — drop the check.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
+    // indices are flipped when writing into the message.
+    // Cells are traced along rays grouped by angle (theta = atan2(y, x)),
+    // nearest to farthest: free until the first obstacle, occupied at it,
+    // unknown behind it.
+    msg.data.assign(rows * cols, -1);
+
+    const auto width = msg.info.width;
+
+    const float half_fov_rad = std::atan(tan_half_hfov);
+    const float bin_range = 2.0f * half_fov_rad;
```
remibettan: `bin_range` reads like the angular width of a single bin, but it's actually the total FOV span (`2 * half_fov_rad`). It makes the derived formulas harder to sanity-check — `x_far * bin_range` is arc length, and `cell_size * N_bins / (2 * x_obs * bin_range)` is bins per half-cell angular width. Renaming to `fov_span` (or similar) makes the physics of each formula immediately readable.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
+    //   height = cells along Y = firmware cols
+    // Origin is the corner of cell (0,0): nearest boundary in X, rightmost in Y.
+    // (cols - cols/2) instead of cols/2 keeps origin.y aligned with the integer
+    // division used in the cell coordinates below when cols is odd.
+    msg.info.map_load_time = t;
+    msg.info.resolution = cell_size;
+    msg.info.width  = static_cast<uint32_t>(rows);
+    msg.info.height = static_cast<uint32_t>(cols);
+    msg.info.origin.position.x = 0.0;
+    msg.info.origin.position.y = -cell_size * static_cast<float>(cols - cols / 2);
```
remibettan: Nit: `(cols + 1) / 2` is the standard ceil-of-half idiom and evaluates to the same value for all `cols`.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
+
+    // Rows are scanned nearest-first while bin_occluded tracks which rays are
+    // already blocked. Occlusion is applied only after a row completes, so cells
+    // at the same depth never shadow each other.
+    // uint8_t rather than bool to avoid vector<bool>'s bit-proxy overhead.
+    std::vector<uint8_t> bin_occluded(N_bins, 0);
+    std::vector<std::pair<int,float>> pending;  // (bin, x) of this row's obstacles
+    pending.reserve(cols);
+
+    for (int fw_row = rows - 1; fw_row >= 0; --fw_row)
```
remibettan, line 995: No unit tests are added for the new algorithm. It has several distinct edge cases worth pinning: odd/even `cols`, FOV-boundary exclusion, `occupancy_max_range` clamp, single-bin shadow propagation, gap-ray shadow spread across adjacent bins, and the `x_obs <= 2 * cell_size` mark-but-no-shadow rule. A fixture that hand-packs a byte array and asserts `data[]` values would catch future regressions cheaply.

realsense2_camera/launch/rs_launch.py
```diff
@@ -68,6 +68,7 @@
                            {'name': 'motion_fps',                   'default': '0', 'description': "'motion stream samples per second'"},
                            {'name': 'unite_imu_method',             'default': "0", 'description': '[0-None, 1-copy, 2-linear_interpolation]'},
                            {'name': 'clip_distance',                'default': '-2.', 'description': "''"},
+                           {'name': 'occupancy_max_range',          'default': '2.5', 'description': "'max reliable range for occupancy raycasting in meters; cells beyond this are unknown (-1)'"},
```
remibettan, line 71: Please also document `occupancy_max_range` in the README / parameters page wherever the other user-facing tunables (`clip_distance`, etc.) are listed. The one-line launch description alone isn't discoverable.

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -251,9 +251,9 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
 
             if (profile.stream_type() == RS2_STREAM_OCCUPANCY)
             {
-                // special handling for occupancy stream, since it is a topic of nav_msgs/msg/GridCells messages
+                // special handling for occupancy stream, since it is a topic of nav_msgs/msg/OccupancyGrid messages
                 // and not a normal image publisher
-                 _occupancy_publisher = _node.create_publisher<nav_msgs::msg::GridCells>("~/occupancy",
+                 _occupancy_publisher = _node.create_publisher<nav_msgs::msg::OccupancyGrid>("~/occupancy",
```
remibettan, line 256: Breaking change for any external subscriber of `~/occupancy` — the message type moves from `nav_msgs/GridCells` to `nav_msgs/OccupancyGrid`. Please call this out in the release notes / commit message so integrators know they need to update their subscribers.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
+
+    // Enough angular bins to resolve one cell width at the farthest depth.
+    const int N_bins = std::max(cols,
+        static_cast<int>(std::ceil(x_far * bin_range / cell_size)) + 1);
+
+    // Rows are scanned nearest-first while bin_occluded tracks which rays are
+    // already blocked. Occlusion is applied only after a row completes, so cells
+    // at the same depth never shadow each other.
+    // uint8_t rather than bool to avoid vector<bool>'s bit-proxy overhead.
+    std::vector<uint8_t> bin_occluded(N_bins, 0);
```
remibettan, line 991: If this topic runs at high rate, `bin_occluded` and `pending` allocate/free every frame. Hoisting both to members and using `bin_occluded.assign(N_bins, 0)` / `pending.clear()` would eliminate that. Not a correctness issue.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
+    for (int fw_row = rows - 1; fw_row >= 0; --fw_row)
     {
-        // AICV algo is packing each 8 cells into one byte. Each byte include 8 bits <--> 8 cells
-        // The rightest bit (LSB) inside the packed byte from AICV algo represnts the closest cell we want to work with in the grid.
-        // e.g. Original Occupancy Cells: 0 0 1 1 0 0 1 0 ---> AICV packing algo ---> 01001100 (not the opposite order)
-        // In this if we check if current cell is occupied.
-        // Note that we start working from the most left bit, aka, the farest point of the grid.
-        if ((frame_as_uint8_arr[i / 8U] & (1U << i % 8)) != 0)
+        const float x = (static_cast<float>(rows - fw_row) - 0.5f) * cell_size;
+        if (x > max_range) continue; // beyond max range: leave unknown
```
remibettan: `x` is monotonically increasing across the nearest-first loop, so once it exceeds `max_range` every remaining row does too. `break` is equivalent and slightly cheaper. Micro-optim only.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -922,41 +923,129 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
-            p3d.x = (cell_size * static_cast<float>(rows)) - cell_size * (static_cast<float>(row) + 0.5f);
-            p3d.y = (cell_size * static_cast<float>(cols)) / 2 - cell_size * (static_cast<float>(col) + 0.5f);
-            p3d.z = 0;
-            msg.cells.push_back(p3d);
+            // Integer cols/2, consistent with origin.y above.
+            const float y = (static_cast<float>(cols / 2) -
+                             static_cast<float>(fw_col) - 0.5f) * cell_size;
+            if (std::fabs(y) >= x * tan_half_hfov) continue; // outside FOV
+
+            const float theta = std::atan2(y, x);
```
remibettan, line 1011: One `atan2` per in-FOV cell — O(rows * cols) trig ops per frame. Fine at typical sizes; if the firmware ever ships larger grids, a precomputed `(fw_row, fw_col) -> bin` lookup table (or a `y / x` fast path since `atan2` is monotonic in `y/x` for fixed `x > 0`) would remove all trig from the hot path.

nivgo: Agreed this is the right optimization if grids ever grow, but the grid size is fixed by firmware and one atan2 per in fov cell at stream rate is negligible today. Keeping the straightforward version for readability and noting the look up option  as the go to if larger grids ship.

realsense2_camera/src/base_realsense_node.cpp
nivgo: Agreed, but publishOccupancyFrame is a private method taking an rs2::frame, so meaningful tests require first extracting the grid construction into a standalone testable function. I would prefer to do that refactor plus a fixture (hand packed byte array -> expected data[], covering odd and even cols, FOV boundaries, max-range clamp, shadow propagation and gap ray spread) as a follow-up PR than grow this one.

realsense2_camera/src/base_realsense_node.cpp
nivgo: I prefer keeping these local  it's ~50 bytes per frame, complete noise next to allocating the message itself. If it ever shows up in profiling i will happily move them.

realsense2_camera/src/base_realsense_node.cpp
remibettan: Fixed in 1d0fb20 — added `depth_info_it->second.width == 0` to the guard.

realsense2_camera/src/base_realsense_node.cpp
remibettan: Fixed in 1d0fb20 — both `origin.position.y` and the per-cell `y` now use float `cols * 0.5f`, so the grid is symmetric about the camera axis for both odd and even `cols` (matches the pre-refactor semantics for odd cols).

realsense2_camera/src/base_realsense_node.cpp
remibettan: Fixed in 1d0fb20 — check removed, with a comment at the site explaining why `bin` is non-negative.

realsense2_camera/src/base_realsense_node.cpp
remibettan: Fixed in 1d0fb20 — renamed to `fov_span` with a clarifying trailing comment; all downstream uses updated.

realsense2_camera/src/base_realsense_node.cpp
remibettan: Moot after 1d0fb20 — the formula switched to float centering (`cols * 0.5f`), so the integer ceil-of-half idiom no longer applies.

realsense2_camera/launch/rs_launch.py
remibettan: Fixed in 1d0fb20 — added an `occupancy_max_range` entry to the README parameters list (default, disable-via-0, example, and dep on depth stream).

realsense2_camera/src/base_realsense_node.cpp
remibettan: Fixed in 1d0fb20 — now `break`, with a comment noting nearest-first monotonicity.

remibettan (approved):

PR #3538, opened 2026-07-19
RSDSO-21716: .db3 rosbag playback tests + de-flake test_align_depth

AviaAv:
## What

- **De-flake `test_align_depth`.** Wait for the camera node before subscribing, so the subscriptions aren't created before the node's publishers exist. Creating them too early races the best-effort sensor QoS match and intermittently yielded zero frames for the whole 50s window (`"Timedout: Data expected, but not received"`, passed on rerun). Every `live_camera` test already gates on `wait_for_node`; `test_align_depth` was the odd one out.
- **Migrate rosbag playback coverage to ROS2 `.db3`.** New `test/rosbag/test_rosbag_db3_playback_tests.py` drives the node's `rosbag_filename` device-playback path (node -> `load_device(.db3)` -> pipeline) with an `rs-convert`'ed `.db3`, for color / depth / aligned_depth_to_color / accel. This is distinct from `ros2 bag play`: the alignment is produced by the node pipeline, not stored as a topic. Ground truth is still read from the source `.bag` (`importRosbag` can't read `.db3`).
- **Helpers.** `get_db3_file_path` (cached `bag -> db3` convert) and `db3_playback_supported` added to `pytest_rs_utils`.

## Notes

- The `.db3` tests **skip** where the installed librealsense predates `.db3` playback (no `BUILD_ROSBAG2` / `rs-convert --output-db3`)  -  same pattern as the existing `.db3` tests. They run where librealsense is built from source with rosbag2.
- Follow-up (not in this PR): host the pre-converted `.db3` alongside the `.bag` and download it, to drop the runtime `rs-convert` dependency.

Tracked on: [RSDSO-21716]

### DIFF
--- a/realsense2_camera/test/post_processing_filters/test_align_depth.py
+++ b/realsense2_camera/test/post_processing_filters/test_align_depth.py
@@ -78,6 +78,11 @@ class TestBasicAlignDepthEnable(pytest_rs_utils.RsTestBaseClass):
             initialize, run and check the data 
             '''
             self.init_test('RsTest'+params['camera_name'])
+            # Wait for the node before subscribing: creating the subscriptions
+            # before the node's publishers exist races the best-effort QoS match
+            # and can yield zero frames for the whole window (the intermittent
+            # flake). Every live_camera test already gates on this.
+            self.wait_for_node(params['camera_name'])
             ret = self.run_test(themes)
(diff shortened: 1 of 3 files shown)


AviaAv:
Closing - opened via the wrong process. Will recreate the correct way.

PR #3539, opened 2026-07-19
Migrate rosbag tests to db3 and deflake align_depth

AviaAv:
Tracked on: [RSDSO-21716]

### DIFF
--- a/realsense2_camera/test/rosbag/test_rosbag_dec_point_tests.py
+++ b/realsense2_camera/test/rosbag/test_rosbag_dec_point_tests.py
@@ -34,8 +34,11 @@ from pytest_rs_utils import delayed_launch_descr_with_parameters
 from pytest_rs_utils import get_rosbag_file_path
 from pytest_rs_utils import get_node_heirarchy
 
+DB3 = "outdoors_1color.db3"
 
-test_params_depth_avg_decimation_1 = {"rosbag_filename":get_rosbag_file_path("outdoors_1color.bag"),
+
+test_params_depth_avg_decimation_1 = {"rosbag_filename":get_rosbag_file_path(DB3),
+    'rosbag_loop': 'true',
     'camera_name': 'Align_Depth_Color_1',
     'color_width': '0',
     'color_height': '0',
@@ -55,7 +58,7 @@ the command used to run is "python3 realsense2_camera/scripts/rs2_test.py depth_
 class TestDepthAvgDecimation1(pytest_rs_utils.RsTestBaseClass):
     def test_depth_avg_decimation_1(self,launch_descr_with_parameters):
         params = launch_descr_with_parameters[1]
-        data = pytest_rs_utils.ImageDepthGetData_decimation(params["rosbag_filename"])
+        data = pytest_rs_utils.ImageDepthGetData_decimation(get_rosbag_file_path(DB3))
         themes = [
         {'topic':get_node_heirarchy(params)+'/depth/image_rect_raw',
          'msg_type':msg_Image,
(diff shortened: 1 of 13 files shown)


Nir-Az (dismissed):

Nir-Az:
Notice all GHA failed

realsense2_camera/test/utils/pytest_rs_utils.py
```diff
@@ -123,18 +124,36 @@ def get_rosbag_file_path(filename):
...
+    ''' Returns the deserialized messages of the topic, read via rosbag2's own reader. '''
+    reader = rosbag2_py.SequentialReader()
+    reader.open(rosbag2_py.StorageOptions(uri=db3_filename, storage_id='sqlite3'),
+                rosbag2_py.ConverterOptions('', ''))
+    reader.set_filter(rosbag2_py.StorageFilter(topics=[topic]))
+    msgs = []
+    while reader.has_next():
+        _, data, _ = reader.read_next()
+        msgs.append(deserialize_message(bytes(data), msg_type))
+    return msgs
```
sysrsbuild-gh-agentic, line 137: ⚠️ **[rs-agentic bot]** Resource leak: `SequentialReader` is never closed before returning. SQLite holds a file lock; if the same `.db3` is opened again in the same process (e.g. a second test case) before Python GC collects this reader, the open will fail or return stale data. Wrap in try/finally:
```python
try:
    while reader.has_next():
        _, data, _ = reader.read_next()
        msgs.append(deserialize_message(bytes(data), msg_type))
finally:
    del reader  # releases the SQLite connection
```

realsense2_camera/test/rosbag/test_rosbag_dec_point_tests.py
```diff
@@ -138,7 +143,7 @@ def process_data(self, themes):
 class TestDepthAvgDecimation1(pytest_rs_utils.RsTestBaseClass):
```
sysrsbuild-gh-agentic, line 143: ⚠️ **[rs-agentic bot]** `TestDepthAvgDecimation1` is defined a second time here (first definition is around line 58). In Python the second definition silently overwrites the first — pytest only discovers and runs this second class, so the first `test_params_depth_avg_decimation_1` fixture (decimation test variant 1) is never executed. Same issue exists for `test_params_depth_avg_decimation_1` which is also re-declared. Consider renaming the second pair to distinct names.

sysrsbuild-gh-agentic (commented):
Auto-generated review by rs-agentic bot.

Migration from `.bag` to `.db3` looks correct overall. Two bugs found below.

realsense2_camera/test/utils/pytest_rs_utils.py
sysrsbuild-gh-agentic: ⚠️ **[rs-agentic bot]** Still not fixed — `SequentialReader` resource leak remains in this push. Please add `try/finally: del reader` as suggested above.

realsense2_camera/test/rosbag/test_rosbag_dec_point_tests.py
sysrsbuild-gh-agentic: ⚠️ **[rs-agentic bot]** Still not fixed — `TestDepthAvgDecimation1` and `test_params_depth_avg_decimation_1` are still defined twice in this push. The first definition is silently overwritten by the second; please rename one pair.

realsense2_camera/test/utils/pytest_rs_utils.py
AviaAv: Not applying. CPython refcounting destroys the reader at function return (no reference cycles here), which closes the SQLite connection — the suggested `del reader` in a `finally` block is equivalent to plain scope exit. The suite already opens the same .db3 several times per process (the play test reads 3 topics, plus every ground-truth helper) and runs green repeatedly. SQLite read locks are shared, so successive readers don't conflict.

---
🤖 AI-assisted reply, generated with [Claude Code](https://claude.com/claude-code)

realsense2_camera/test/rosbag/test_rosbag_dec_point_tests.py
AviaAv: Pre-existing on ros2-development — the identical duplicate class/params pair exists upstream (lines 55/138); this PR only touched the neighboring param-dict lines. Keeping the migration scoped; the duplicate deserves its own cleanup PR.

---
🤖 AI-assisted reply, generated with [Claude Code](https://claude.com/claude-code)

Nir-Az (approved):

PR #3551, opened 2026-08-18
Ashraf/zerocopy nitors v2

ashrafk93:
Add optional NITROS (GPU zero-copy) color publishing
Opt-in path to publish the color stream to an Isaac ROS / NITROS graph on the GPU, skipping the sensor_msgs/Image → DDS → cudaMemcpy round-trip. Additive and gated — default builds and ~/color/image_raw are unchanged. (RSDEV-6254)

What's new
nitros_image_publisher.{h,cpp} — wraps ManagedNitrosPublisher<NitrosImage>; publishes on ~/color/nitros_image.
publishNitrosFrame() hooked into publishFrame(); enable_color_nitros param (default off).
CMake BUILD_WITH_NITROS option (default OFF) → finds/links CUDA + isaac_ros_nitros packages.
Enable
colcon build --packages-select realsense2_camera_msgs realsense2_camera \
  --cmake-args -DBUILD_WITH_NITROS=ON
ros2 run realsense2_camera realsense2_camera_node --ros-args -p enable_color_nitros:=true
Needs an Isaac ROS workspace + librealsense built with BUILD_WITH_CUDA_ZEROCOPY.

### DIFF
--- a/realsense2_camera/src/nitros_image_publisher.cpp
+++ b/realsense2_camera/src/nitros_image_publisher.cpp
@@ -1,0 +1,308 @@
+// Copyright 2026 RealSense, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifdef BUILD_WITH_NITROS
+
+#include "nitros_image_publisher.h"
+
+#include <isaac_ros_managed_nitros/managed_nitros_publisher.hpp>
+#include <isaac_ros_nitros_image_type/nitros_image.hpp>
+#include <isaac_ros_nitros_image_type/nitros_image_builder.hpp>
+
+#include <cuda_runtime.h>
+#include <algorithm>
+#include <chrono>
+#include <cstdint>
+#include <cstdlib>
+#include <exception>
+#include <optional>
+#include <utility>
+
+// cudaMemPool_t / cudaMallocFromPoolAsync (the stream-ordered allocator) need CUDA 11.2.
+#if defined(CUDART_VERSION) && CUDART_VERSION >= 11020
+#define RS_NITROS_HAS_MEMPOOL 1
+#else
+#define RS_NITROS_HAS_MEMPOOL 0
+#endif
+
+namespace realsense2_camera
+{
+using nvidia::isaac_ros::nitros::ManagedNitrosPublisher;
+using nvidia::isaac_ros::nitros::NitrosImage;
+using nvidia::isaac_ros::nitros::NitrosImageBuilder;
+
+namespace
+{
+rclcpp::Logger logger()
+{
+    return rclcpp::get_logger("NitrosImagePublisher");
+}
+
+// Set RS_NITROS_LEGACY_ALLOC=1 to get a plain cudaMalloc + synchronous cudaMemcpy per frame instead
+// of the pooled + async path. Only useful for A/B benchmarking the two allocation strategies.
+bool legacyAllocRequested()
+{
+    const char * env = std::getenv("RS_NITROS_LEGACY_ALLOC");
+    return env != nullptr && env[0] == '1';
+}
+
+// How often the accumulated per-frame publish cost is reported (at DEBUG).
+constexpr uint64_t STATS_LOG_INTERVAL = 100;
+}  // namespace
+
+struct NitrosImagePublisher::Impl
+{
+    std::shared_ptr<ManagedNitrosPublisher<NitrosImage>> pub;
+
+    // Stream-ordered pool the GXF-owned buffers are carved from; null when pooling is unavailable
+    // or RS_NITROS_LEGACY_ALLOC=1, in which case publish() falls back to cudaMalloc.
+#if RS_NITROS_HAS_MEMPOOL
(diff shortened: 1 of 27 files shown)


realsense2_camera/src/nitros_image_publisher.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+#if RS_NITROS_HAS_MEMPOOL
+    if (legacyAllocRequested()) {
+        RCLCPP_INFO(logger(), "RS_NITROS_LEGACY_ALLOC=1: using cudaMalloc/cudaFree per frame");
+    } else {
+        cudaMemPoolProps props{};
+        props.allocType = cudaMemAllocationTypePinned;
+        props.handleTypes = cudaMemHandleTypeNone;
+        props.location.type = cudaMemLocationTypeDevice;
+        props.location.id = 0;
```
ashrafk93, line 143: **Bug: CUDA device hardcoded to index 0.**
`props.location.id = 0` always targets the first CUDA device. On a system with multiple CUDA devices (or in a multi-process deployment), the memory pool will be created on the wrong device if the sensor data lives on a different one. Use `cudaGetDevice(&props.location.id)` before this line to query the currently active device.

*Auto-generated by rs-agentic bot.*

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -300,6 +300,30 @@ void BaseRealSenseNode::startPublishers(const std::vector<stream_profile>& profi
...
+                // Additive NITROS-native GPU zero-copy publisher for the color stream.
+                if (_enable_color_nitros && profile.stream_type() == RS2_STREAM_COLOR)
+                {
+                    std::string nitros_format, encoding;
+                    unsigned int bpp = 0;
+                    if (getNitrosImageFormat(profile.format(), nitros_format, encoding, bpp))
+                    {
+                        std::stringstream nitros_topic;
+                        nitros_topic << "~/" << stream_name << "/nitros_image";
+                        _nitros_image_publishers[sip] =
```
ashrafk93, line 313: **Counterpart `stopPublishers()` cleanup is missing.**
This entry is added here but `stopPublishers()` does not erase it. When the color stream stops (profile change, stream disable), the `NitrosImagePublisher` — with its CUDA stream, event, and memory pool — will persist until this map entry is overwritten by the next `startPublishers()` call, or until node shutdown if the stream is not restarted. Add `_nitros_image_publishers.erase(sip);` in `stopPublishers()` alongside the existing `_image_publishers.erase(sip)`.

*Auto-generated by rs-agentic bot.*

ashrafk93 (commented):
**Missing `stopPublishers()` cleanup — CUDA resource leak on stream restart**

`_nitros_image_publishers[sip]` is populated inside `startPublishers()` but `stopPublishers()` has no matching `_nitros_image_publishers.erase(sip)`. On any sensor/profile change (resolution switch, stream enable/disable toggle) the node calls `stopPublishers()` + `startPublishers()`; without the erase the old `NitrosImagePublisher` — holding a CUDA stream, event, and memory pool — lives until `startPublishers()` overwrites the map entry (or node shutdown if the stream is not restarted). Please add `_nitros_image_publishers.erase(sip);` inside the `#ifdef BUILD_WITH_NITROS` block in `stopPublishers()`, mirroring how `_image_publishers.erase(sip)` is handled.

*Auto-generated by rs-agentic bot.*

realsense2_camera/src/nitros_image_publisher.cpp
```diff
@@ -0,0 +1,308 @@
...
+        err = cudaMemPoolCreate(&_impl->pool, &props);
+        if (err != cudaSuccess) {
```
ashrafk93, line 145: **Device ID hardcoded to 0.** `props.location.id = 0` always targets CUDA device 0. On any system where the active device is not 0 (possible in multi-GPU or future Jetson configurations) the pool will be created on the wrong device, causing `cudaMallocFromPoolAsync` to fail or silently allocate from the wrong memory domain. Replace with:
```cpp
int current_device = 0;
cudaGetDevice(&current_device);
props.location.id = current_device;
```

*Auto-generated by the rs-agentic review bot.*

ashrafk93 (commented):
**BUG — `_nitros_image_publishers` not cleaned up in `stopPublishers()`**
`stopPublishers()` (`rs_node_setup.cpp` ~line 207) erases from `_image_publishers` (and `_depth_aligned_image_publishers`, `_info_publishers`, …) but never touches `_nitros_image_publishers`. On any stream stop/reconfigure cycle (e.g. reconnect, `changeResolution`) the stale `NitrosImagePublisher` entry keeps its CUDA stream, event, and memory pool alive until node exit. `stopPublishers()` should add `_nitros_image_publishers.erase(sip)` for color streams, guarded by `#ifdef BUILD_WITH_NITROS`, mirroring the existing `_image_publishers.erase(sip)` logic.

**`f.get_gpu_data_or_upload()` is not part of the standard public librealsense 2 API**
Enabling `-DBUILD_WITH_NITROS=ON` against an unpatched librealsense (one not built with `BUILD_WITH_CUDA_ZEROCOPY`) will silently compile and then fail with a cryptic linker error. The CMake `BUILD_WITH_NITROS` block finds/links the Isaac ROS packages but never verifies the librealsense was built with CUDA zerocopy. Consider a `check_symbol_exists` / `try_compile` step or a `REALSENSE2_HAS_GPU_DATA` compile-definition from the SDK headers so the error surfaces clearly at configure time.

**Minor — `publishNitrosFrame()` invoked for every frame on every stream when `_enable_color_nitros=false`**
The `#ifdef BUILD_WITH_NITROS` block in `publishFrame()` calls `publishNitrosFrame()` unconditionally. For depth, infrared, and all non-color streams the map lookup immediately hits `end()` and returns, so results are correct, but a top-level `if (!_enable_color_nitros) return;` guard inside `publishNitrosFrame()` would make the early-exit intent explicit and save the map-lookup overhead on every non-color frame.

---
*Auto-generated by the rs-agentic review bot.*

PR #3552, opened 2026-08-19
nitros publish color, depth and aligned depth as GPU images on Jetson

ashrafk93:
### What

Optional GPU-image publishing for Jetson / [Isaac ROS](https://nvidia-isaac-ros.github.io/): the wrapper hands a frame's CUDA device pointer to NITROS so a GPU consumer reads it in place instead of copying it out of host memory.

Everything is gated twice — `-DBUILD_WITH_NITROS=ON` (default **OFF**) and a per-stream parameter (default **false**). With the CMake flag off the wrapper is unchanged; with it on but the parameters off, nothing extra is created. The existing `image_raw` topics are never touched.

### Streams

| Parameter | Topic | Format |
| --- | --- | --- |
| `enable_color_nitros` | `~/color/nitros_image` | `nitros_image_rgb8` (also `bgr8`/`rgba8`/`bgra8`) |
| `enable_depth_nitros` | `~/depth/nitros_image` | `nitros_image_mono16` (Z16) |
| `enable_aligned_depth_nitros` | `~/aligned_depth_to_<stream>/nitros_image` | `nitros_image_mono16` |

Formats with no NITROS equivalent (YUYV color, RAW8) are skipped with a warning. Point cloud is not implemented — it needs a different message type and an XYZ repack.

### Composition is required, and the node says so

A GPU pointer is only valid in the process that created it, so the consumer must be a component in the same container. Launched separately, rclcpp runs the `NitrosImage` type adapter and copies every frame device→host→device — measurably *worse* than the plain topic. Rather than leave that silent, the publisher measures its own per-frame cost and warns once when it sees the cross-process pattern. It also warns once per stream when librealsense had to upload a frame, i.e. the stream wasn't GPU-resident and NITROS is costing an extra copy.

Note `use_intra_process_comms` must **not** be set at node level: NITROS enables intra-process comms on its own publishers, and setting it node-wide also applies it to the transient-local negotiation topics, which makes the node fail to start.

### Implementation notes

- Isaac ROS / GXF / CUDA headers are confined to `nitros_image_publisher.cpp`; the rest of the wrapper sees only a pimpl handle and never pulls in NITROS.
- GXF's release callback on release-3.2 is a plain `cudaFree`, so it cannot be handed the SDK frame-pool pointer. Buffers come from a CUDA stream-ordered memory pool instead — `cudaFree` is legal on that memory and returns it to the pool. Per frame at 1080p this took allocation from ~517 µs to ~18 µs.
- The device-to-device copy runs on a private high-priority non-blocking stream and is awaited on an event, since release-3.2's `NitrosImage` carries no CUDA event for a downstream process to synchronise on.
- Publishing is skipped entirely while nothing is subscribed (the graph is queried every 10 frames), matching how the `image_raw` path gates on `get_subscription_count()`.
- The rs2→ROS encoding mapping is reused from `initializeFormatsMaps()`; only the NITROS type name is new.
- Incompatible with `-DUSE_LIFECYCLE_NODE=ON` (NITROS needs an `rclcpp::Node`) — enforced with a CMake error.

### Measurements

AGX Orin, D435i, color 1920x1080 @ 30, one camera, whole pipeline as percent of one CPU core:

| Setup | CPU | FPS |
| --- | --- | --- |
| NITROS, composed | **17.5%** | 27.97 |
| `image_raw`, composed + `use_intra_process_comms` | 20.6% | 26.27 |
| `image_raw`, composed | 25.5% | 27.56 |
| `image_raw`, two processes | 33.1% | 27.12 |

### Testing

Jetson AGX Orin (JetPack 6.2, CUDA 12.6), Isaac ROS 3.2, RealSense D435i:

- Builds clean with `-DBUILD_WITH_NITROS=ON` and with the default (off) — 0 warnings both ways.
- `cppcheck` CI gate: 0 errors; with `BUILD_WITH_NITROS` defined and `--enable=style`: 0 findings in the new files.
- Composed consumer on each stream: color 312/312, depth 362/362, aligned depth 363/363 frames zero-copy, 0 uploads.
- Verified the subscriber gating does no work with nothing subscribed, and that the cross-process warning fires exactly once.

### Dependency

Requires an Isaac ROS workspace (`isaac_ros_nitros`, `isaac_ros_managed_nitros`, `isaac_ros_nitros_image_type`) and a librealsense built with `BUILD_WITH_CUDA_ZEROCOPY`. Depth being GPU-resident on RSUSB builds additionally needs realsenseai/librealsense#15580; without it the depth stream still works but the node warns that it was uploaded.


realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -1346,6 +1355,101 @@ void BaseRealSenseNode::publishFrame(
...
+        return;
+    }
+
+    std_msgs::msg::Header header;
+    header.stamp = t;
+    header.frame_id = OPTICAL_FRAME_ID(stream);
+
+    // A frame the SDK had to upload was never GPU-resident, so publishing it through NITROS costs
+    // an extra copy over the plain image topic. Say so once per stream rather than every frame.
+    if (copied && _nitros_upload_warned.insert(stream).second)
```
ashrafk93: **BUG – data race on `_nitros_upload_warned`.**
`frame_callback()` is invoked concurrently from per-sensor SDK threads (one per active sensor), so `_nitros_upload_warned.insert()` can be called from multiple threads simultaneously. `std::set` is not thread-safe for concurrent writes; this is undefined behaviour. Protect the insert (and the `_nitros_upload_warned` field generally) with the existing node-level mutex, or use an `std::atomic` flag per stream instead.

*Auto-generated by rs-agentic bot.*

realsense2_camera/src/nitros_image_publisher.cpp
```diff
@@ -0,0 +1,327 @@
...
+                    cudaGetErrorString(err));
+        _impl->copy_done = nullptr;
+    }
+
+#if RS_NITROS_HAS_MEMPOOL
+    cudaMemPoolProps props{};
+    props.allocType = cudaMemAllocationTypePinned;
+    props.handleTypes = cudaMemHandleTypeNone;
+    props.location.type = cudaMemLocationTypeDevice;
+    props.location.id = 0;
```
ashrafk93: **BUG – CUDA device hardcoded to 0.**
`props.location.id = 0` assumes the active CUDA device is always device 0. If the node is ever run on a multi-GPU host (or if the CUDA context is created on a different device), the pool will be on the wrong device and `cudaMallocFromPoolAsync` will fail or silently misallocate. Use `cudaGetDevice(&props.location.id)` instead to query the current device at construction time.

*Auto-generated by rs-agentic bot.*

ashrafk93 (commented):
Missing `CHANGELOG.rst` entry — the contribution guidelines require new features to be recorded in `realsense2_camera/CHANGELOG.rst`.

*Auto-generated by rs-agentic bot.*

ashrafk93:
`bench/` is a local benchmarking workspace (NITROS consumer/viewer components used to produce the numbers in the description) that is deliberately not part of the repo. Happy to drop the ignore line if you'd rather not carry it.

realsense2_camera/src/base_realsense_node.cpp
ashrafk93: change approved

realsense2_camera/src/nitros_image_publisher.cpp
ashrafk93: change approved

PR #3553, opened 2026-08-24
Fix realsense driver deadlock when quickly toggling stream parameters

nnarain:
**Context**

We want to be able to toggle the camera stream off and on to save system resources. In the ROS 2 version of the realsense driver, this is done by toggling the associated ROS parameters.
Issue
I found that if you toggle ROS parameters for different streams within a certain time of each other the driver will hang and never recover.

**Root Cause**

This appears to be because each parameter change triggers a "profile changed update". There is a mutex protected boolean flag and a condition variable that triggers the update in a separate thread.
When the "monitoring profile change" thread update, it makes a call to updateSensors(). This updates the underlying sensor configuration using via librealsense. This has it's own lock that is acquired. This update can take hundreds to milliseconds to over a second. While this is happening the profile change mutex is still being held. So another parameter change coming along can't acquire the lock.
Admittedly I'm not entirely sure why it never recovers but that is the observed behaviour. In the mix here there is also the built in parameter service waiting for the lock and the ROS 2 api doesn't really give you a good/clear timeout callback/return. In addition to this the realsense driver parameter are dynamic and base on different camera models. So it's hard to trace the exact flow of callbacks.

**Solution**

The solution here is to reset the profile change flags once entering the if branch. Then unlock/lock around the call to updateSensors. This unblocks incoming parameter updates while the sensor profile is updating. These are just boolean flags, so there's no queuing required. It'll just see an update required and update the sensors.

### DIFF
--- a/realsense2_camera/src/rs_node_setup.cpp
+++ b/realsense2_camera/src/rs_node_setup.cpp
@@ -62,6 +62,14 @@ void BaseRealSenseNode::monitoringProfileChanges()
                                 ))
             {
                 ROS_DEBUG("Profile has changed");
+                _is_profile_changed = false;
+                _is_align_depth_changed = false;
+
+                #if defined (ACCELERATE_GPU_WITH_GLSL)
+                    _is_accelerate_gpu_with_glsl_changed = false;
+                #endif
+
+                lock.unlock();
                 try
                 {
                     updateSensors();
@@ -70,12 +78,7 @@ void BaseRealSenseNode::monitoringProfileChanges()
                 {
                     ROS_ERROR_STREAM("Error updating the sensors: " << e.what());
                 }
-                _is_profile_changed = false;
-                _is_align_depth_changed = false;
-
-                #if defined (ACCELERATE_GPU_WITH_GLSL)
-                    _is_accelerate_gpu_with_glsl_changed = false;
-                #endif
+                lock.lock();
             }
         }
     };


sysrsbuild-gh:

realsense2_camera/src/rs_node_setup.cpp
```diff
@@ -70,12 +78,7 @@ void BaseRealSenseNode::monitoringProfileChanges()
                 {
                     ROS_ERROR_STREAM("Error updating the sensors: " << e.what());
                 }
-                _is_profile_changed = false;
-                _is_align_depth_changed = false;
-
-                #if defined (ACCELERATE_GPU_WITH_GLSL)
-                    _is_accelerate_gpu_with_glsl_changed = false;
-                #endif
+                lock.lock();
```
ashrafk93, line 81: **Bug – exception-safety gap:** `lock.lock()` is only reached when `updateSensors()` returns normally *or* throws a type derived from `std::exception`. If it throws anything else (e.g. a raw integer, a librealsense C-API error not in the `std::exception` hierarchy, or any other unchecked type), the exception bypasses this line and propagates out of the thread function. The `unique_lock` destructor is safe (it no longer owns the mutex), but the monitoring thread silently terminates, leaving the driver unable to recover from future parameter changes.

Consider wrapping with a catch-all that re-locks before rethrowing:
```cpp
lock.unlock();
try {
    updateSensors();
} catch (const std::exception& e) {
    ROS_ERROR_STREAM("Error updating the sensors: " << e.what());
} catch (...) {
    lock.lock();
    throw;
}
lock.lock();
```
Or use a scope-guard RAII wrapper so the re-lock is unconditional.

*Auto-generated by the rs-agentic review bot.*

ashrafk93 (commented):
The approach — resetting the flags before unlocking, dropping the mutex around `updateSensors()`, then re-acquiring — is a sound fix for the described deadlock. One exception-safety gap exists (see inline comment).

*Auto-generated by the rs-agentic review bot.*

Nir-Az:
@nnarain  thanks for the PR, any inputs for the code review comments please?

nnarain:
> @nnarain  thanks for the PR, any inputs for the code review comments please?

Need to take a closer look, I thought when I went through this I found no update was missed.

I'll get back to you shortly

nnarain:
Hm, getting back to this now. Thinking about how to address the review comment.

PR #3555, opened 2026-08-30
Occupancy: publish camera-computed grid + ~/occupancy_certainty; remove host-side raycasting

nivgo:
# Occupancy: parse self-describing MAP1 in-node + `~/occupancy_certainty`

## Summary

Moves occupancy post-processing out of the node and onto the camera. The
realsense2_camera node now **unpacks** what the D5xx firmware produces instead of
re-deriving it: it parses the self-describing **MAP1** occupancy payload in-node
(validated, CRC-checked), publishes the binarized grid on `~/occupancy` and the
graded per-cell confidence on a new `~/occupancy_certainty`. A single publisher
auto-detects three wire formats, so it works across firmware generations with no
SDK-version requirement beyond occupancy-stream support.

This replaces the previous in-node angular raycasting / FOV-masking / max-range
logic (that responsibility now lives in the camera), so the node
is a thin, well-bounded unpacker.

## How it works

```mermaid
flowchart TD
  F["occupancy frame bytes"] --> N{"raw_data null?"}
  N -->|yes| D1["drop"]
  N -->|no| M{"MAP1 magic?"}
  M -->|yes| P["parse_occg:<br/>CRC + geometry + bounds"]
  P -->|ok| S["splitCells"]
  P -->|invalid| D2["drop + throttled warn"]
  M -->|no| B{"size fits a byte grid?"}
  B -->|yes| BY["byte-legacy"] --> S
  B -->|no| BI["bit-legacy -> ~/occupancy only"]
  S --> O1["~/occupancy (binarized)"]
  S --> O2["~/occupancy_certainty (1..100)"]
```

**MAP1 frame** (self-describing, so geometry/origin travel with the data):

```
[ 20B frame_header ] magic "MAP1", version, data_type=OCCG, flags, payload_size,
                     profile_id, stream_generation, crc32
[ 32B occg_header  ] width(=forward X), height(=lateral Y), resolution_mm, cell_stride,
                     origin_x_mm, origin_y_mm, source_frame_id, timestamp_us, cell_count
[ int8 cells ...   ] row-major data[cy*width+cx];  -1 unknown / 0 free / 1..100 occupied
```

`splitCells` produces both outputs from the graded cells:

```
cell value   ~/occupancy (binary)                 ~/occupancy_certainty
  -1          -1 unknown                            -1
   0           0 free                                0
  1..100       100 if >= occupancy_occupied_threshold, else -1     the raw value
```

### Live in rviz

**`~/occupancy` and the new `~/occupancy_certainty` rendered together (D555):**

rviz_new_certainty_topic

## Changes

- **`occupancy_map1.h`** — table-free MAP1 parser: packed structs with size
  `static_assert`s, CRC-32/ISO-HDLC, and ordered bounds checks (frame length →
  payload_size → occg_header → `cell_count == width*height` geometry cross-check)
  that reject a malformed frame before any indexing.
- **`occupancy_grid_utils.h`** — `splitCells` (binarize + certainty split).
- **`base_realsense_node.cpp`** — `publishOccupancyFrame` three-carrier dispatch
  (MAP1 / byte-legacy / bit-legacy) + `publishOccupancyFromMap1`; drops invalid
  frames (never republished raw); removes the old raycasting/`occupancy_max_range`.
- **New topic** `~/occupancy_certainty` (byte-legacy + MAP1 formats); **new param**
  `occupancy_occupied_threshold` (default 100, range [1,100]).
- **Tests** — `test_occupancy_map1.cpp` (CRC reference vector, bad-CRC, truncated,
  geometry mismatch, and a **golden frame** pinning the byte↔position axis contract)
  and `test_occupancy_grid_utils.cpp`.

## Review feedback pre-empted

Concerns raised on the related producer PR (#461) and by the bots here, addressed:

- **Per-ray evidence, not global max** — decoding side has no raycasting; the
  camera owns it, and its FW PR keeps the per-ray bound.
- **Buffer-capacity / geometry validation before indexing** — `cell_count ==
  width*height` cross-check + size-ordered guards.
- **Integer-overflow** in the legacy path — bit-legacy now uses `size_t` for the
  size guard, `assign`, and index (matching the byte path).
- **Null data pointer** — guarded before all paths.
- **CRC-optional frames** — documented that production OCCG frames set `FLAG_CRC32`;
  a clear flag is trusted on geometry alone (still overflow-safe).

## Compatibility / breaking changes

- **Breaking for external subscribers of the old occupancy behavior:** in-node
  raycasting/FOV-masking and `occupancy_max_range` are removed; unknown/shadow
  semantics now come from the camera. `~/occupancy` message type is unchanged
  (`nav_msgs/OccupancyGrid`).
- **New topic** `~/occupancy_certainty` — additive; legacy 1-bit firmware publishes
  `~/occupancy` only.
- No SDK version requirement beyond occupancy-stream support (three-format auto-detect).

## Testing

- [x] `test_occupancy_map1` 7/7, `test_occupancy_grid_utils` 3/3
- [x] `colcon build --packages-select realsense2_camera` clean
- [x] Live on D555: `~/occupancy` + `~/occupancy_certainty` publish at ~18 Hz
      (occupancy stream enabled standalone)
- [x] README updated (new topic, param, three-format auto-detect)

### DIFF
--- a/realsense2_camera/include/occupancy_map1.h
+++ b/realsense2_camera/include/occupancy_map1.h
@@ -1,0 +1,143 @@
+// Copyright 2026 RealSense, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Self-contained, header-only parser for the MAP1 self-describing occupancy-grid
+// wire format (librealsense#15557 / device-mgr interface/MappingEp12.h). Ported
+// from the reviewed implementation in librealsense
+// (src/proc/occupancy-map1.h/.cpp, branch niv/occupancy-map1-consume) with the
+// same semantics; this copy has no dependency on librealsense internals
+// (rsutils, synthetic-stream, metadata-parser) -- the ROS node discriminates the
+// MAP1 payload purely by its bytes, not by SDK-side metadata/enums.
+
+#pragma once
+#include <cstdint>
+#include <cstddef>
+#include <cstring>
+
+namespace realsense2_camera {
+namespace map1 {
+
+#pragma pack(push, 1)
+struct frame_header {
+    uint32_t magic;              // 0x3150414D "MAP1"
+    uint16_t version;
+    uint8_t  data_type;          // 2 = occupancy grid
+    uint8_t  flags;              // bit 0: crc32 field is valid
+    uint32_t payload_size;
+    uint16_t profile_id;
+    uint16_t stream_generation;
+    uint32_t crc32;              // CRC-32/ISO-HDLC over payload_size bytes
+};
+struct occg_header {
+    uint16_t width;              // cells along +X (forward)
+    uint16_t height;             // cells along +Y (left); camera at height/2
+    uint16_t resolution_mm;
+    uint16_t cell_stride;
+    int32_t  origin_x_mm;
+    int32_t  origin_y_mm;
+    uint32_t source_frame_id;
+    uint64_t timestamp_us;
+    uint32_t cell_count;
+};
+#pragma pack(pop)
+static_assert(sizeof(frame_header) == 20, "MAP1 frame header ABI");
+static_assert(sizeof(occg_header) == 32, "MAP1 OCCG header ABI");
+
+constexpr uint32_t MAGIC = 0x3150414DU;
+constexpr uint8_t DATA_TYPE_OCCG = 2U;
+constexpr uint8_t FLAG_CRC32 = 1U;
+
+enum class parse_result { ok, not_map1, not_occg, truncated, bad_crc, bad_geometry, bad_version };
+
+struct occg_view {
+    occg_header header;
+    const int8_t* cells;
+};
+
+inline bool is_map1(const uint8_t* data, size_t size)
+{
+    if (data == nullptr || size < sizeof(frame_header)) return false;
+    uint32_t magic;
+    std::memcpy(&magic, data, sizeof(magic));
+    return magic == MAGIC;
+}
+
+// CRC-32/ISO-HDLC (poly 0xEDB88320, reflected, init/xorout 0xFFFFFFFF) -- same
+// algorithm/result as the standard zlib CRC-32. Table-free bit-at-a-time loop
+// (no dependency on librealsense's rsutils table implementation); mathematically
+// identical output for the same input.
+inline uint32_t crc32_iso_hdlc(const void* data, size_t size)
+{
+    uint32_t crc = 0xFFFFFFFFu;
+    const uint8_t* p = static_cast<const uint8_t*>(data);
+    for (size_t i = 0; i < size; ++i)
+    {
+        crc ^= p[i];
+        for (int bit = 0; bit < 8; ++bit)
+        {
+            const uint32_t mask = static_cast<uint32_t>(-static_cast<int32_t>(crc & 1u));
+            crc = (crc >> 1) ^ (0xEDB88320u & mask);
+        }
+    }
+    return ~crc;
+}
+
+inline parse_result parse_occg(const uint8_t* data, size_t size, occg_view* out)
+{
+    if (!is_map1(data, size)) return parse_result::not_map1;
+    frame_header fh;
+    std::memcpy(&fh, data, sizeof(fh));
+    if (fh.data_type != DATA_TYPE_OCCG) return parse_result::not_occg;
+    // Accept any minor revision of major version 1 (0x01xx); reject anything else
+    // (e.g. a future incompatible 0x02xx) rather than silently misinterpreting the
+    // rest of the frame under this struct layout.
+    if ((fh.version >> 8) != 0x01) return parse_result::bad_version;
+    // Widen to uint64_t before adding: fh.payload_size / oh.cell_count are
+    // device-controlled uint32_t values, and on a 32-bit size_t build
+    // `sizeof(...) + <uint32_t>` could wrap around and bypass these checks.
+    if ((uint64_t)size < (uint64_t)sizeof(fh) + fh.payload_size ||
+        fh.payload_size < sizeof(occg_header))
+        return parse_result::truncated;
+    const uint8_t* payload = data + sizeof(fh);
+    // Production occupancy (data_type OCCG) frames are expected to set FLAG_CRC32.
+    // When it is clear the CRC is skipped and the frame is trusted on the geometry/size
+    // checks below alone: a payload bit-flip then goes undetected, but still cannot
+    // overflow a consumer buffer (the cell_count cross-check guarantees that).
+    if ((fh.flags & FLAG_CRC32) &&
+        crc32_iso_hdlc(payload, fh.payload_size) != fh.crc32)
+        return parse_result::bad_crc;
+    occg_header oh;
+    std::memcpy(&oh, payload, sizeof(oh));
+    if ((uint64_t)fh.payload_size < (uint64_t)sizeof(oh) + oh.cell_count)
+        return parse_result::truncated;
+    // A zero-sized grid or an unsupported cell layout (cell_stride is currently
+    // always 1 per the wire spec; any other value means a layout this parser
+    // doesn't know how to interpret) are both malformed geometry.
+    if (oh.width == 0 || oh.height == 0 || oh.cell_stride != 1)
+        return parse_result::bad_geometry;
+    // width/height/cell_count are independently device-controlled fields; without
+    // this check a frame with a small width/height (small allocated destination
+    // buffer) and a large cell_count (large memcpy length) passes every check
+    // above -- especially with FLAG_CRC32 clear, where the CRC never runs -- and
+    // overflows whatever fixed-size buffer a consumer sized from width*height.
+    // width,height <= 0xFFFF so the product fits in uint32_t without overflow.
+    if (oh.cell_count != (uint32_t)oh.width * (uint32_t)oh.height)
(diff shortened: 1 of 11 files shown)


sysrsbuild-gh:

sysrsbuild-gh-agentic (commented):
**Auto-generated review by rs-agentic bot**

Good overall structure — the three-carrier auto-detection, header-only MAP1 parser with CRC+geometry validation, and the `splitCells` unit tests are solid.

Two bugs found in the legacy paths of `publishOccupancyFrame`:

**BUG 1 — signed `int` overflow in bit-legacy size guards and index arithmetic** (`base_realsense_node.cpp`)

`rows` and `cols` are `int` (cast from `rs2_metadata_type`). Three spots use `int` multiplication:

- Line ~1021: `(rows * cols + 7) / 8` — the size guard for the bit-legacy path. `n` is already a safe `size_t`; use `(n + 7) / 8u` instead.
- Line ~1029: `msg.data.assign(rows * cols, 0)` — same issue; use `msg.data.assign(n, 0)`.
- Line ~1034: `const int i = fw_row * cols + fw_col` — inner loop index, can overflow for large device-controlled grids; use `const size_t i = static_cast<size_t>(fw_row) * cols + fw_col`.

All three share the same root cause: device-controlled metadata values multiplied as `int`. The byte-legacy path correctly uses `n` everywhere already.

**BUG 2 — `msg.data` used uninitialized in the MAP1 path** (`base_realsense_node.cpp`, `publishOccupancyFromMap1`)

`splitCells` calls `occupancy_out.resize(n)` and `certainty_out.resize(n)` on the passed vectors. `msg.data` and `certainty_msg.data` are default-constructed (empty) before the call, so `resize` will value-initialize the new elements to `0` rather than leaving them uninitialized — this is actually safe. Flagging this for awareness: the correctness relies on `splitCells` writing every element via its own loop (`for i in [0, n)`). If future code changes `splitCells` to skip elements (sparse update), the caller's pre-fill assumption would silently break. Consider `assign(n, 0)` / `assign(n, -1)` before the call as defensive pre-fill, consistent with the bit-legacy path.

**Enhancement — FLAG_CRC32=0 skips all payload integrity checking** (`occupancy_map1.h`, line 102)

See inline comment.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -963,97 +977,108 @@ void BaseRealSenseNode::publishOccupancyFrame(rs2::frame f, const rclcpp::Time&
...
-    std::vector<uint8_t> bin_occluded(N_bins, 0);
-    std::vector<std::pair<int,float>> pending;  // (bin, x) of this row's obstacles
-    pending.reserve(cols);
+    // Legacy 1-bit grids never publish certainty - nothing to do if only
+    // ~/occupancy_certainty has subscribers.
+    if (!occ_wanted)
+        return;
 
-    for (int fw_row = rows - 1; fw_row >= 0; --fw_row)
+    if (raw_size < static_cast<size_t>((rows * cols + 7) / 8))
```
sysrsbuild-gh-agentic: **[BUG] `rows * cols` signed-int overflow — auto-generated by rs-agentic bot**

`rows` and `cols` are `int`; their product is `int` arithmetic and can overflow for device-controlled metadata values. The safe `size_t` variable `n` is already computed above. Use:
```cpp
if (raw_size < (n + 7) / 8u)
```
Same applies to `msg.data.assign(rows * cols, 0)` a few lines below (use `n`) and `const int i = fw_row * cols + fw_col` in the inner loop (use `size_t i`).

realsense2_camera/include/occupancy_map1.h
```diff
@@ -0,0 +1,128 @@
...
+    // rest of the frame under this struct layout.
+    if ((fh.version >> 8) != 0x01) return parse_result::bad_version;
+    // Widen to uint64_t before adding: fh.payload_size / oh.cell_count are
+    // device-controlled uint32_t values, and on a 32-bit size_t build
+    // `sizeof(...) + <uint32_t>` could wrap around and bypass these checks.
+    if ((uint64_t)size < (uint64_t)sizeof(fh) + fh.payload_size ||
+        fh.payload_size < sizeof(occg_header))
+        return parse_result::truncated;
+    const uint8_t* payload = data + sizeof(fh);
+    if ((fh.flags & FLAG_CRC32) &&
```
sysrsbuild-gh-agentic, line 117: **[Enhancement] FLAG_CRC32 clear skips all payload integrity checking — auto-generated by rs-agentic bot**

When `FLAG_CRC32` is not set, a truncated-but-geometry-valid or bit-flipped payload passes all checks silently. The geometry cross-check (`cell_count == width * height`) catches one class of corruption, but the cell data itself is unverified. This is acceptable if the wire spec guarantees that `data_type == 2` frames always set `FLAG_CRC32`; if not, a firmware bug sending unflagged frames would deliver garbage cells without any log warning. Worth documenting (or adding an assertion) that production frames always set this flag.

realsense2_camera/src/base_realsense_node.cpp
sysrsbuild-gh-agentic: change approved

realsense2_camera/include/occupancy_map1.h
sysrsbuild-gh-agentic: change approved

sysrsbuild-gh-agentic (commented):
**Auto-review (synchronize push) — rs-agentic bot**

Reviewed the new commit. Previous findings (signed overflow in legacy bit-path, FLAG_CRC32 behavior) are resolved and approved.

No new bugs found in this push:
- MAP1 parser bounds checks and `cell_count == width × height` cross-check are correct.
- `splitCells` threshold comparison is correct for all int8_t values in [−1, 100].
- Byte-legacy path uses `size_t n` throughout — no overflow.
- Parse-error log-and-drop path prevents MAP1 bytes from falling through to the legacy unpack.

One minor note (not a bug): `publishOccupancyFromMap1` always runs `splitCells` for both output vectors even when only one topic has subscribers. For typical grid sizes this is negligible.

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -920,28 +922,59 @@ bool BaseRealSenseNode::shouldPublishCameraInfo(const stream_index_pair& sip)
...
+        // unlike the legacy paths below, which relayout fw_row/fw_col. The self-describing
+        // wire format owns the axis convention; HW-validated on D555.
+        publishOccupancyFromMap1(view, t);
         return;
     }
-    const float tan_half_hfov = (static_cast<float>(depth_info_it->second.width) * 0.5f)
-                                / static_cast<float>(depth_info_it->second.k.at(0));
 
-    auto frame_as_uint8_arr = (uint8_t*)f.get_data();
+    ROS_DEBUG("Publishing Occupancy Grid Frame");
```
OhadMeir: Maybe this should too be throttled. Will add a log entry for every frame for a camera publishing a non-MAP1 occupancy.

OhadMeir (dismissed):
Looks great, one minor comment

realsense2_camera/launch/rs_launch.py
```diff
@@ -68,7 +68,6 @@
                            {'name': 'motion_fps',                   'default': '0', 'description': "'motion stream samples per second'"},
                            {'name': 'unite_imu_method',             'default': "0", 'description': '[0-None, 1-copy, 2-linear_interpolation]'},
                            {'name': 'clip_distance',                'default': '-2.', 'description': "''"},
-                           {'name': 'occupancy_max_range',          'default': '2.5', 'description': "'max reliable range for occupancy raycasting in meters; cells beyond this are unknown (-1)'"},
                            {'name': 'angular_velocity_cov',         'default': '0.01', 'description': "''"},
```
sysrsbuild-gh-agentic, line 72: **[BUG] `occupancy_occupied_threshold` not exposed as a launch argument — auto-generated by rs-agentic bot**

`occupancy_max_range` was removed here but `occupancy_occupied_threshold` was not added. Users cannot set it via `ros2 launch ... occupancy_occupied_threshold:=50` the way other params like `clip_distance` work. Add:
```python
{'name': 'occupancy_occupied_threshold', 'default': '100', 'description': 'Binarization threshold [1-100] for graded occupancy cells'}
```

realsense2_camera/src/base_realsense_node.cpp
```diff
@@ -920,28 +922,59 @@ bool BaseRealSenseNode::shouldPublishCameraInfo(const stream_index_pair& sip)
...
+
     auto cols = static_cast<int>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_GRID_COLUMNS));
     auto rows = static_cast<int>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_GRID_ROWS));
     auto cell_size = static_cast<float>(f.get_frame_metadata(RS2_FRAME_METADATA_OCCUPANCY_CELL_SIZE)) / 100.0f; // cm -> m
+    const size_t n = static_cast<size_t>(rows) * static_cast<size_t>(cols);
+    if (rows <= 0 || cols <= 0)
```
sysrsbuild-gh-agentic, line 971: **[BUG] `n` computed before the negative-metadata guard — auto-generated by rs-agentic bot**

`static_cast<size_t>(rows)` wraps a negative `rows` to a huge value; multiplying two such values overflows `size_t` (wraps, but silently). The guard on the next lines catches this before `n` is used, but the inline comment `"n == 0"` is wrong for negative values. Move the `n` computation after the guard:
```cpp
if (rows <= 0 || cols <= 0) { ... return; }
const size_t n = static_cast<size_t>(rows) * static_cast<size_t>(cols);
```

sysrsbuild-gh-agentic (commented):
Auto-generated review by rs-agentic bot (synchronize event).

realsense2_camera/src/base_realsense_node.cpp
nivgo: Thanks Ohad! good catch, throttled it.

realsense2_camera/launch/rs_launch.py
sysrsbuild-gh-agentic: change approved

realsense2_camera/src/base_realsense_node.cpp
sysrsbuild-gh-agentic: change approved

OhadMeir (approved):

PR #3556, opened 2026-09-01
Default all streams of an untagged sensor to one shared frame rate

AviaAv:
Tracked on: [RSDEV-13629]

### DIFF
--- a/realsense2_camera/src/profile_manager.cpp
+++ b/realsense2_camera/src/profile_manager.cpp
@@ -606,7 +606,8 @@ std::map<stream_index_pair, rs2::stream_profile> MotionProfilesManager::getFallb
             fallback_profiles = fps_profiles.second;
     }
 
-    ROS_INFO_STREAM("No default profile found. Using the lowest frame rate shared by the most streams.");
+    ROS_INFO_STREAM("No default profile found. Using the lowest frame rate shared by the most streams: "
+                    << fallback_profiles.begin()->second.fps());
     if (fallback_profiles.size() < all_sips.size())
         ROS_WARN_STREAM("No frame rate is shared by all the motion streams, "
                         << (all_sips.size() - fallback_profiles.size()) << " of them keep their own.");
--- a/realsense2_camera/src/profile_manager.cpp
+++ b/realsense2_camera/src/profile_manager.cpp
@@ -133,9 +133,20 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
 
     if (sip_default_profiles.empty())
     {
-        ROS_INFO_STREAM("No default profile found. Setting the first available profile as the default one.");
-        rs2::stream_profile first_profile = _all_profiles.front();
-        sip_default_profiles[{first_profile.stream_type(), first_profile.stream_index()}] = first_profile;
+        ROS_INFO_STREAM("No default profile found. Using the lowest frame rate that all the streams offer.");
+        std::map<int, std::map<stream_index_pair, rs2::stream_profile>> profiles_per_fps;
+        for (auto profile : _all_profiles)
+        {
+            stream_index_pair sip(profile.stream_type(), profile.stream_index());
+            profiles_per_fps[profile.fps()].emplace(sip, profile);
+        }
+
+        // One rate for all the streams - some sensors cannot stream theirs at different ones.
+        for (auto& fps_profiles : profiles_per_fps)
+        {
+            if (fps_profiles.second.size() > sip_default_profiles.size())
+                sip_default_profiles = fps_profiles.second;
+        }
     }
 
     return sip_default_profiles;
--- a/realsense2_camera/src/profile_manager.cpp
+++ b/realsense2_camera/src/profile_manager.cpp
@@ -133,7 +133,7 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
 
     if (sip_default_profiles.empty())
     {
-        ROS_INFO_STREAM("No default profile found. Using the lowest frame rate that all the streams offer.");
+        ROS_INFO_STREAM("No default profile found. Using the lowest frame rate shared by the most streams.");
         std::map<int, std::map<stream_index_pair, rs2::stream_profile>> profiles_per_fps;
         for (auto profile : _all_profiles)
         {
--- a/realsense2_camera/src/profile_manager.cpp
+++ b/realsense2_camera/src/profile_manager.cpp
@@ -132,24 +132,16 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
     }
 
     if (sip_default_profiles.empty())
-    {
-        ROS_INFO_STREAM("No default profile found. Using the lowest frame rate shared by the most streams.");
-        std::map<int, std::map<stream_index_pair, rs2::stream_profile>> profiles_per_fps;
-        for (auto profile : _all_profiles)
-        {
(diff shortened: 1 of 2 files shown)


realsense2_camera/src/profile_manager.cpp
```diff
@@ -133,9 +133,20 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
 
     if (sip_default_profiles.empty())
     {
-        ROS_INFO_STREAM("No default profile found. Setting the first available profile as the default one.");
-        rs2::stream_profile first_profile = _all_profiles.front();
-        sip_default_profiles[{first_profile.stream_type(), first_profile.stream_index()}] = first_profile;
+        ROS_INFO_STREAM("No default profile found. Using the lowest frame rate that all the streams offer.");
```
sysrsbuild-gh-agentic: ⚠️ **[rs-agentic bot]** Misleading log message — the algorithm below picks the fps group with the **most streams**, not the **lowest** fps. If the fps with the most streams is not the lowest available one, the message is incorrect. Consider updating to something like: `"No default profile found. Using the frame rate shared by the most streams."`

realsense2_camera/src/profile_manager.cpp
```diff
@@ -133,9 +133,20 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
...
+        for (auto profile : _all_profiles)
+        {
+            stream_index_pair sip(profile.stream_type(), profile.stream_index());
+            profiles_per_fps[profile.fps()].emplace(sip, profile);
+        }
+
+        // One rate for all the streams - some sensors cannot stream theirs at different ones.
+        for (auto& fps_profiles : profiles_per_fps)
+        {
+            if (fps_profiles.second.size() > sip_default_profiles.size())
```
sysrsbuild-gh-agentic: ⚠️ **[rs-agentic bot]** If `_all_profiles` is empty, `profiles_per_fps` will also be empty, the loop never runs, and `sip_default_profiles` remains empty — the function returns an empty map silently. Callers may not handle this. Consider adding a warning log and/or a guard before the loop:
```cpp
if (profiles_per_fps.empty())
{
    ROS_WARN_STREAM("No profiles available to select a default from.");
    return sip_default_profiles;
}
```

sysrsbuild-gh-agentic (commented):
Auto-generated review by rs-agentic bot. Two potential bugs found in the fallback profile selection logic.

realsense2_camera/src/profile_manager.cpp
Nir-Az: Agree with the bot here — the loop picks the fps bucket with the most streams, and "lowest" only breaks ties between equal-sized buckets. Suggest "Using the lowest frame rate shared by the most streams."

🤖 Generated by AI

realsense2_camera/src/profile_manager.cpp
```diff
@@ -133,9 +133,20 @@ std::map<stream_index_pair, rs2::stream_profile> ProfilesManager::getDefaultProf
...
+        ROS_INFO_STREAM("No default profile found. Using the lowest frame rate that all the streams offer.");
+        std::map<int, std::map<stream_index_pair, rs2::stream_profile>> profiles_per_fps;
+        for (auto profile : _all_profiles)
+        {
+            stream_index_pair sip(profile.stream_type(), profile.stream_index());
+            profiles_per_fps[profile.fps()].emplace(sip, profile);
+        }
+
+        // One rate for all the streams - some sensors cannot stream theirs at different ones.
+        for (auto& fps_profiles : profiles_per_fps)
```
Nir-Az: This picks the fps shared by the most streams, not necessarily one that covers all of them — if no rate is common to every stream (say accel 63/250 with gyro 200/400) it returns a single stream and the other one keeps its own min fps, quietly reintroducing the mismatch this PR is fixing. Can't happen on D400/D500 since they tag motion defaults, but a ROS_WARN when the result is smaller than the sensor's stream count would save someone a long debug session on the next untagged SKU.

🤖 Generated by AI

Nir-Az:
Since getDefaultProfiles() sits on the base class, this changes VideoProfilesManager too. For an untagged video sensor with a single stream every fps bucket ties at size 1, so the default silently becomes the lowest fps the sensor offers rather than the first enumerated profile. Could we keep the base fallback as _all_profiles.front() and move the bucket logic into a virtual getFallbackProfiles() overridden in MotionProfilesManager? Same fix for motion, no blast radius on video.

Last one: if accel and gyro share no rate at all, the loop returns one of them and the other keeps its own min, so we're back to mismatched rates at stream start with nothing in the log. A ROS_WARN when the result covers fewer streams than the sensor has would make that visible.

🤖 Generated by AI

realsense2_camera/src/profile_manager.cpp
AviaAv: Done in 80cc138f — reworded to "Using the lowest frame rate shared by the most streams."

realsense2_camera/src/profile_manager.cpp
AviaAv: _all_profiles cannot be empty here — getDefaultProfiles() throws "Wrong commands sequence. No profiles set." on entry if it is (profile_manager.cpp:122), and MotionProfilesManager::registerFPSParams returns early on the same condition. So profiles_per_fps always has at least one entry and the map is never empty. Leaving it as is to avoid dead code.

realsense2_camera/src/profile_manager.cpp
AviaAv: Agreed, fixed in 80cc138f.

realsense2_camera/src/profile_manager.cpp
AviaAv: Added in c4e98870 — ROS_WARN when the chosen set covers fewer streams than the sensor has, naming how many keep their own rate. Silent on the D430 GMSL since all its motion streams share 50.

realsense2_camera/src/profile_manager.cpp
sysrsbuild-gh-agentic: change approved

realsense2_camera/src/profile_manager.cpp
sysrsbuild-gh-agentic: change approved

realsense2_camera/src/profile_manager.cpp
sysrsbuild-gh-agentic: change approved

Nir-Az: Warn reads right, thanks. One leftover — 80cc138f is titled "Say which frame rate the fallback picked" but the log still prints a fixed string. Worth appending the value, e.g. `<< fallback_profiles.begin()->second.fps()`, since that's the number you'd want in hand when the warn fires right after it. Non-blocking.

🤖 Generated by AI

Nir-Az (dismissed):
Structure looks good now — base fallback is unchanged so the video path is untouched, and the shared-rate logic is confined to motion where it belongs. Only leftover is the non-blocking log suggestion above.

🤖 Generated by AI

realsense2_camera/src/profile_manager.cpp
AviaAv: Done in 0317e44e — the log line now ends with the chosen rate. On the D430 GMSL it prints "...shared by the most streams: 50".

sysrsbuild-gh-agentic: change approved

Nir-Az (approved):

PR #3558, opened 2026-09-03
Skip rosbag tests when the installed librealsense2 cannot play compressed db3

Nir-Az:
**Overview:** Makes the rosbag tests skip, instead of fail, when the installed librealsense2 is too old to play back the compressed `.db3` recordings they now use.

Tracked on [RSDSO-21716]

## Why
The `pre-release` workflow has failed on every commit since #3539, on every branch. It installs `ros-humble-librealsense2 2.58.4` from `repo.ros2.org/ubuntu/testing`, and the migrated tests feed it recordings whose frames live on `/compressed` and `/compressedDepth` topics. Support for reading those landed in librealsense PR #15556, which is not in v2.58.4, so the SDK reader logs

```
ros2_reader.cpp:156 read_next_data: unknown message type on topic: /device_0/sensor_1/Color_0/image/metadata
```

for every metadata message and delivers no frames. 8 of 11 tests fail or time out. `main.yml` is unaffected because it builds librealsense from the `development` branch.

`test_rosbag_db3_play_test.py` used to guard against exactly this ("rosdistro's librealsense2 may predate the bag-to-db3 converter"), but that probe did not survive the migration.

## Summary
- Added `realsense2_camera/test/conftest.py` (43 lines, 13 of them the license header), which skips every `rosbag`-marked test when the installed librealsense2 cannot read compressed `.db3`.
- The check probes the installed library for the `/compressedDepth` topic suffix rather than comparing versions: a source build of `development` reports 2.58.0, lower than the released 2.58.4 that lacks the support, so a version floor would skip the tests exactly where they pass.
- Returns True when it cannot tell, so the tests run.
- No test file changes; all eight affected tests already carry `@pytest.mark.rosbag`.

The guard becomes inert on its own once a librealsense2 package containing #15556 reaches the ROS servers.

## Test plan
- [ ] `pre-release` (humble/jazzy/kilted) reports the rosbag tests as skipped instead of the current 8 failures.
- [ ] `CI` and `CI - Rolling` stay green — they build librealsense from `development`, so the probe must let the tests run there.

Both are readable from this PR's own checks; the second is the one that proves the probe does not over-skip.

---
🤖 Generated by AI

### DIFF
--- a/realsense2_camera/test/conftest.py
+++ b/realsense2_camera/test/conftest.py
@@ -25,7 +25,8 @@ def librealsense_path():
     '''
     try:
         ctypes.CDLL('librealsense2.so')
-        return next(l.split()[-1] for l in open('/proc/self/maps') if 'librealsense2.so' in l)
+        with open('/proc/self/maps') as maps:
+            return next(l.split()[-1] for l in maps if 'librealsense2.so' in l)
     except Exception:
         pass
     libs = [lib for d in os.environ.get('LD_LIBRARY_PATH', '').split(':') if d
--- a/realsense2_camera/test/conftest.py
+++ b/realsense2_camera/test/conftest.py
@@ -11,9 +11,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import subprocess
+import ctypes
+import glob
+import os
 
 import pytest
+
+
+def librealsense_path():
+    ''' Path of the librealsense2 the camera node will load. Resolved through the dynamic
+        loader, since the ROS packages install into /opt/ros/<distro>/lib, which is reached
+        via LD_LIBRARY_PATH and is not in the ldconfig cache.
+    '''
+    try:
+        ctypes.CDLL('librealsense2.so')
+        return next(l.split()[-1] for l in open('/proc/self/maps') if 'librealsense2.so' in l)
+    except Exception:
+        pass
+    libs = [lib for d in os.environ.get('LD_LIBRARY_PATH', '').split(':') if d
+            for lib in glob.glob(os.path.join(d, 'librealsense2.so*')) if os.path.getsize(lib)]
+    return libs[0] if libs else None
 
 
 def sdk_plays_compressed_db3():
@@ -23,13 +40,13 @@ def sdk_plays_compressed_db3():
         branch reports 2.58.0 - lower than the released 2.58.4 that lacks the support.
         Returns True when it cannot tell, so the tests run.
     '''
+    path = librealsense_path()
+    if not path:
+        return True
     try:
-        libs = subprocess.run(['ldconfig', '-p'], capture_output=True, text=True).stdout
-        path = next(l.rsplit('=>', 1)[-1].strip() for l in libs.splitlines()
-                    if 'librealsense2.so' in l and '=>' in l)
         with open(path, 'rb') as lib:
             return b'/compressedDepth' in lib.read()
-    except Exception:
+    except OSError:
         return True
 
 
--- a/realsense2_camera/test/conftest.py
+++ b/realsense2_camera/test/conftest.py
@@ -11,47 +11,33 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import functools
 import subprocess
 
 import pytest
 
 
-@functools.lru_cache(maxsize=1)
 def sdk_plays_compressed_db3():
-    ''' Whether the installed librealsense2 can play back the compressed .db3 recordings
-        the rosbag tests use.
-
-        Those recordings keep their frames on /compressed and /compressedDepth topics, which
-        librealsense only learned to read in PR #15556. The librealsense2 packages on the ROS
-        servers can still predate it, and the reader then logs "unknown message type" for every
-        metadata message and delivers no frames.
-
-        This probes the installed library for the topic suffix instead of comparing versions,
-        because a source build of the development branch reports 2.58.0 - lower than the
-        released 2.58.4 that lacks the support - so a version floor would skip the tests
-        exactly where they do work. Fails open: if the library cannot be located, run them.
+    ''' Whether the installed librealsense2 can read the /compressed and /compressedDepth
+        topics of the .db3 test recordings (librealsense PR #15556). Probes the library for
+        the topic suffix rather than its version, since a source build of the development
+        branch reports 2.58.0 - lower than the released 2.58.4 that lacks the support.
+        Returns True when it cannot tell, so the tests run.
     '''
     try:
-        ldconfig = subprocess.run(['ldconfig', '-p'], capture_output=True, text=True).stdout
-    except (OSError, subprocess.SubprocessError):
+        libs = subprocess.run(['ldconfig', '-p'], capture_output=True, text=True).stdout
+        path = next(l.rsplit('=>', 1)[-1].strip() for l in libs.splitlines()
+                    if 'librealsense2.so' in l and '=>' in l)
+        with open(path, 'rb') as lib:
+            return b'/compressedDepth' in lib.read()
+    except Exception:
         return True
-    for line in ldconfig.splitlines():
-        if 'librealsense2.so' in line and '=>' in line:
-            try:
-                with open(line.split('=>')[-1].strip(), 'rb') as lib:
-                    return b'/compressedDepth' in lib.read()
-            except OSError:
-                return True
-    return True
 
 
 def pytest_collection_modifyitems(config, items):
     if sdk_plays_compressed_db3():
         return
-    skip_rosbag = pytest.mark.skip(
-        reason="installed librealsense2 cannot play compressed .db3 recordings "
-               "(needs librealsense PR #15556)")
+    skip = pytest.mark.skip(reason="installed librealsense2 cannot play compressed .db3 "
+                                   "recordings (needs librealsense PR #15556)")
     for item in items:
         if 'rosbag' in item.keywords:
-            item.add_marker(skip_rosbag)
+            item.add_marker(skip)
--- a/realsense2_camera/test/conftest.py
+++ b/realsense2_camera/test/conftest.py
@@ -1,0 +1,57 @@
+# Copyright 2026 RealSense, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import functools
+import subprocess
+
+import pytest
+
+
+@functools.lru_cache(maxsize=1)
+def sdk_plays_compressed_db3():
+    ''' Whether the installed librealsense2 can play back the compressed .db3 recordings
+        the rosbag tests use.
+
+        Those recordings keep their frames on /compressed and /compressedDepth topics, which
+        librealsense only learned to read in PR #15556. The librealsense2 packages on the ROS
+        servers can still predate it, and the reader then logs "unknown message type" for every
+        metadata message and delivers no frames.
+
+        This probes the installed library for the topic suffix instead of comparing versions,
+        because a source build of the development branch reports 2.58.0 - lower than the
+        released 2.58.4 that lacks the support - so a version floor would skip the tests
+        exactly where they do work. Fails open: if the library cannot be located, run them.
+    '''
+    try:
+        ldconfig = subprocess.run(['ldconfig', '-p'], capture_output=True, text=True).stdout
+    except (OSError, subprocess.SubprocessError):
+        return True
+    for line in ldconfig.splitlines():
+        if 'librealsense2.so' in line and '=>' in line:
+            try:
+                with open(line.split('=>')[-1].strip(), 'rb') as lib:
+                    return b'/compressedDepth' in lib.read()
+            except OSError:
+                return True
+    return True
+
+
+def pytest_collection_modifyitems(config, items):
+    if sdk_plays_compressed_db3():
+        return
+    skip_rosbag = pytest.mark.skip(
+        reason="installed librealsense2 cannot play compressed .db3 recordings "
+               "(needs librealsense PR #15556)")
+    for item in items:
+        if 'rosbag' in item.keywords:
+            item.add_marker(skip_rosbag)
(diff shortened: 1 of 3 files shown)


realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+        servers can still predate it, and the reader then logs "unknown message type" for every
+        metadata message and delivers no frames.
+
+        This probes the installed library for the topic suffix instead of comparing versions,
+        because a source build of the development branch reports 2.58.0 - lower than the
+        released 2.58.4 that lacks the support - so a version floor would skip the tests
+        exactly where they do work. Fails open: if the library cannot be located, run them.
+    '''
+    try:
+        ldconfig = subprocess.run(['ldconfig', '-p'], capture_output=True, text=True).stdout
```
sysrsbuild-gh-agentic: ⚠️ **BUG (auto-generated by rs-agentic bot):** No timeout on `subprocess.run`. If `ldconfig` hangs (e.g. broken ldcache), test collection blocks indefinitely.

```python
subprocess.run(['ldconfig', '-p'], capture_output=True, text=True, timeout=10)
```
Catch `subprocess.TimeoutExpired` alongside the existing exceptions and return `True`.

realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+    except (OSError, subprocess.SubprocessError):
+        return True
+    for line in ldconfig.splitlines():
+        if 'librealsense2.so' in line and '=>' in line:
+            try:
+                with open(line.split('=>')[-1].strip(), 'rb') as lib:
+                    return b'/compressedDepth' in lib.read()
```
sysrsbuild-gh-agentic: ⚠️ **BUG (auto-generated by rs-agentic bot):** `lib.read()` loads the entire `.so` into memory — librealsense2.so can be 30–100 MB. Use `mmap` for a zero-copy scan:

```python
import mmap
with open(path, 'rb') as lib:
    with mmap.mmap(lib.fileno(), 0, access=mmap.ACCESS_READ) as mm:
        return b'/compressedDepth' in mm
```

realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+        exactly where they do work. Fails open: if the library cannot be located, run them.
+    '''
+    try:
+        ldconfig = subprocess.run(['ldconfig', '-p'], capture_output=True, text=True).stdout
+    except (OSError, subprocess.SubprocessError):
+        return True
+    for line in ldconfig.splitlines():
+        if 'librealsense2.so' in line and '=>' in line:
+            try:
+                with open(line.split('=>')[-1].strip(), 'rb') as lib:
```
sysrsbuild-gh-agentic: 💡 **Enhancement (auto-generated by rs-agentic bot):** Only the **first** `librealsense2.so` entry from `ldconfig` is checked and the loop exits. `ldconfig -p` can list multiple entries (versioned symlinks: `.so.2`, `.so.2.58.4`). The first hit might be a symlink to the real file, which is fine, but if it isn't readable the code falls through to `return True` (fail-open) instead of trying the next entry. Consider continuing the loop on `OSError` rather than returning immediately.

sysrsbuild-gh-agentic (commented):
Auto-review by rs-agentic bot. Good approach overall — probing the binary instead of comparing versions is the right call. A few items worth addressing.

Nir-Az:
@AviaAv handing this over to you since it follows your db3 migration. The pre-release job has been red on every branch since #3539: it installs `ros-humble-librealsense2 2.58.4` from the ROS servers, and I confirmed the hosted `outdoors_1color.db3` carries its frames only on `/image/data/compressed` and `/image/data/compressedDepth` (no plain `/image/data` topic at all), which 2.58.4's stream-topic regex `/(data|metadata)$` filters out — so no frame is built and every metadata message ends up logging `unknown message type`. Reading those topics came in librealsense #15556, which landed three days after v2.58.4 was tagged, so nothing is wrong with the bags or the tests, they just need an SDK newer than anything released.

This PR guards it with a conftest that skips the `rosbag`-marked tests when the installed library lacks the support, probing for the topic suffix rather than a version floor because a source build of `development` reports 2.58.0 and would otherwise skip in the main CI where it works. Placement is your call though — it could equally be a `pytest.skip` inside the shared `launch_descr_*` fixtures or a marker on each test class, and you know that test layout better than I do. Also worth deciding whether it's worth guarding at all versus just waiting for the next librealsense2 release to reach the ROS servers, given pre-release is informational and self-heals.

Nir-Az:
The first attempt did not fire — pre-release stayed red and my skip reason never appeared in the log, so I pulled the actual deb apart to see why. Two findings: the `/compressedDepth` string check is a valid discriminator (absent from the released `librealsense2.so.2.58.4`, present in a development build), but the ROS package installs to `/opt/ros/humble/lib/x86_64-linux-gnu/`, which is reached via LD_LIBRARY_PATH and is not in the ldconfig cache — so my lookup found nothing and failed open, and the tests ran exactly as before.

b142f56a resolves the library through the dynamic loader instead (`/proc/self/maps` after a `CDLL`, falling back to LD_LIBRARY_PATH), which I verified against the real released `.so` (returns False, tests skip), a library carrying the string (returns True, tests run) and no library at all (True, fail open). Waiting on this run to confirm it in the container.

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: Still present in the refactored code — `lib.read()` still loads the full `.so` into memory.

realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def librealsense_path():
+    ''' Path of the librealsense2 the camera node will load. Resolved through the dynamic
+        loader, since the ROS packages install into /opt/ros/<distro>/lib, which is reached
+        via LD_LIBRARY_PATH and is not in the ldconfig cache.
+    '''
+    try:
+        ctypes.CDLL('librealsense2.so')
+        return next(l.split()[-1] for l in open('/proc/self/maps') if 'librealsense2.so' in l)
```
sysrsbuild-gh-agentic: ⚠️ **BUG (auto-generated by rs-agentic bot):** `open('/proc/self/maps')` is not wrapped in a `with` block — the file handle leaks until GC collects the generator. Fix:
```python
with open('/proc/self/maps') as maps:
    return next(l.split()[-1] for l in maps if 'librealsense2.so' in l)
```

realsense2_camera/test/conftest.py
Nir-Az: The ldconfig call is gone in b142f56a - the library is resolved through the loader now, so there is no subprocess left to time out.

realsense2_camera/test/conftest.py
Nir-Az: Keeping the plain read - it is an 11 MB file read once at collection (I measured the released librealsense2.so.2.58.4 at 11,087,992 bytes, not 30-100 MB). Worth noting the suggested snippet is also wrong: "b\"...\" in mm" on an mmap object does not do a substring search and quietly returns False even when the bytes are present, so it would have silently disabled the guard - it needs mapped.find(...) != -1.

realsense2_camera/test/conftest.py
Nir-Az: Also gone with the ldconfig lookup in b142f56a.

realsense2_camera/test/conftest.py
Nir-Az: Fixed in 4bdd88b4 - the maps read is inside a with block now.

sysrsbuild-gh-agentic: change approved

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: You are correct on both counts — `b'/compressedDepth' in mm` iterates the mmap as bytes (integers) and always returns `False`, silently disabling the guard. The mmap suggestion was wrong. And 11 MB read once at collection time is not a concern. The plain `lib.read()` approach is fine. Acknowledged.

Nir-Az:
@AviaAv re-pinging for a review — this is ready now. All checks are green on 4bdd88b4, including the three pre-release jobs that have been red on every branch since #3539, and the bot threads are answered. The pre-release runs also dropped from 26-28 min to 11-17 min, and the log shows the rosbag files skipping in under a second each with everything else still running, so the guard is doing what it says rather than passing by luck.

Still your call whether to take it at all: pre-release is not a required check and `ros2-development` has no branch protection, so the alternative is simply living with the red until a librealsense2 carrying #15556 reaches the ROS servers. If you do want it, the placement is also yours — conftest hook as written, a skip inside the shared `launch_descr_*` fixtures, or a marker per test class.

realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+        return True
+
+
+def pytest_collection_modifyitems(config, items):
+    if sdk_plays_compressed_db3():
+        return
+    skip = pytest.mark.skip(reason="installed librealsense2 cannot play compressed .db3 "
+                                   "recordings (needs librealsense PR #15556)")
+    for item in items:
+        if 'rosbag' in item.keywords:
```
AviaAv: Over-skips two tests that don't depend on the SDK version:

- `test_rosbag_db3_play_test.py:35` — its own header says "Verify vanilla `ros2 bag play` (not librealsense)"; it reads the `.db3` through rosbag2's own reader. It passed in `pre-release` before this PR, and is skipped now (Test #9, 0.61s).
- `test_rosbag_depth_tests.py:174` `test_non_existing_rosbag` — expects 0 chunks, never calls `run_test`, passes on any SDK.

The marker is right for both (they do use a recording), so exempt rather than strip:

```python
        if 'rosbag' in item.keywords and 'no_librealsense' not in item.keywords:
```

plus `@pytest.mark.no_librealsense` on those two.

🤖 Generated by AI

realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+        via LD_LIBRARY_PATH and is not in the ldconfig cache.
+    '''
+    try:
+        ctypes.CDLL('librealsense2.so')
+        with open('/proc/self/maps') as maps:
+            return next(l.split()[-1] for l in maps if 'librealsense2.so' in l)
+    except Exception:
+        pass
+    libs = [lib for d in os.environ.get('LD_LIBRARY_PATH', '').split(':') if d
+            for lib in glob.glob(os.path.join(d, 'librealsense2.so*')) if os.path.getsize(lib)]
```
AviaAv: `os.path.getsize(lib)` sits outside the `try`, so it breaks fail-open instead of extending it: a dangling `librealsense2.so*` symlink in any `LD_LIBRARY_PATH` directory raises `OSError` into pytest collection — a collection error, not a skip and not a run.

The size filter also changes nothing; a zero-byte `.so` fails the substring check anyway. Drop it:

```python
    libs = [lib for d in os.environ.get('LD_LIBRARY_PATH', '').split(':') if d
            for lib in glob.glob(os.path.join(d, 'librealsense2.so*'))]
```

🤖 Generated by AI

realsense2_camera/test/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+        return True
+    try:
+        with open(path, 'rb') as lib:
+            return b'/compressedDepth' in lib.read()
+    except OSError:
+        return True
+
+
+def pytest_collection_modifyitems(config, items):
+    if sdk_plays_compressed_db3():
```
AviaAv: When this skips, nothing in the log says so. 19 tests disappear from the run with no reason and no path printed — so if the probe ever resolves the wrong `.so`, it looks identical to an old SDK.

One print fixes it:

```python
def pytest_collection_modifyitems(config, items):
    path = librealsense_path()
    if sdk_plays_compressed_db3(path):
        return
    print(f"rosbag tests skipped: {path} lacks compressed .db3 support")
```

with `sdk_plays_compressed_db3` taking the path instead of calling `librealsense_path()` itself.

🤖 Generated by AI

AviaAv (commented):
Approach and probe verified — `v2.58.4` has zero `compressedDepth` occurrences in `src/media/ros2/`, `development` has them in 4 files, so the discriminator is sound. Source-build CI ran all rosbag tests (36 tests, 4 pre-existing skips); `pre-release` now skips them. Both test-plan items hold.

Three comments inline: one real over-skip, one uncaught `OSError`, one observability gap.

🤖 Generated by AI

realsense2_camera/test/conftest.py
Nir-Az: Good catch - fixed in 56eb005b. Both now carry @pytest.mark.no_librealsense and the hook exempts it, so they run in pre-release again while keeping the rosbag marker.

realsense2_camera/test/conftest.py
Nir-Az: Fixed in 56eb005b - size filter dropped. You are right that it broke the fail-open property it looked like it was helping.

realsense2_camera/test/conftest.py
Nir-Az: Fixed in 56eb005b - sdk_plays_compressed_db3 takes the path now and the hook prints it on skip. This one bit me already: when the first attempt silently failed open, the log looked identical to an old SDK and I had to unpack the deb to tell the difference.

Nir-Az:
@AviaAv all three addressed in 56eb005b, and the branch is rebased onto current `ros2-development`. The two SDK-independent tests now carry `no_librealsense` and are exempt from the skip, the size filter is gone, and the hook prints the resolved path when it skips — that last one had already cost me a debugging round, since a silent fail-open looked exactly like an old SDK in the log.

Checks are re-running; worth a look at whether `test_rosbag_db3_play_test` and `test_non_existing_rosbag` come back as passing in pre-release rather than skipped, which is the direct check on your first comment.

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved — `no_librealsense` exemption added in conftest.py and `@pytest.mark.no_librealsense` applied to both tests in the new commit.

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved — `os.path.getsize` removed; glob fallback is clean with no size check.

realsense2_camera/test/conftest.py
sysrsbuild-gh-agentic: change approved — `print("rosbag tests skipped: %s lacks compressed .db3 support" % path)` is present in the new conftest.py.

AviaAv (approved):
LGTM

PR #3561, opened 2026-09-14
Enable only the streams the device tags as default

AviaAv:
Tracked on: [RSDEV-13627]

`enable_<stream>` now defaults to whether the SDK tags that stream as default, instead of `true` for everything the sensor exposes. A D401 GMSL tags Depth + Color 0 + Color 1 and not infrared - Color 1 switches both imagers to raw Bayer, so IR cannot run with it. Enabling everything asks for a set the device cannot open, and bare `ros2 run` fails there.

## Deltas

`ros2 launch` - no change. `rs_launch.py` passes `enable_infra1`/`enable_infra2` explicitly, which overrides the registered default.

`ros2 run` - untagged streams lose their default. Measured:

| device | before | after |
|---|---|---|
| D401 GMSL | fails to start | Depth + Color 0 + Color 1, node up |
| D455 | Depth, IR1, IR2, Color, IMU | Depth, Color, IMU |
| D585 2C | Depth, IR1, IR2, Color 1, Color 2, IMU | Depth, Color 1, Color 2, IMU |

Overridable: `--ros-args -p enable_infra1:=true -p enable_infra2:=true`.

IR drops out because no infrared profile is tagged default on any device today - the `RS2_STREAM_INFRARED` tags in `d400-factory.cpp` ask for `RS2_FORMAT_RGB8`, which these cameras don't expose. Fixing those tags would restore IR here, and would also move viewer and `pipeline.start()` defaults.

## Alternative

Default `enable_infra1`/`enable_infra2` to `false` in C++, matching `rs_launch.py`. Smaller blast radius and no dependence on tags, but it hardcodes stream-specific policy instead of asking the device, costs IR on bare `ros2 run` just the same, and doesn't help the next device that can't stream everything it exposes.


sysrsbuild-gh-agentic (commented):
Auto-generated review by rs-agentic bot.

Two issues worth addressing before merge — see inline comments.

realsense2_camera/src/profile_manager.cpp
```diff
@@ -95,7 +96,7 @@ void ProfilesManager::registerSensorUpdateParam(std::string template_name,
                 // Disabling Infra 0 stream by default
                 params[sip] = std::make_shared<T>(false);
             else
-                params[sip] = std::make_shared<T>(value);
+                params[sip] = std::make_shared<T>(value && sip_default_profiles.count(sip));
         }
```
sysrsbuild-gh-agentic, line 100: ⚠️ **BUG (auto-generated by rs-agentic bot)**

`registerSensorUpdateParam` is explicitly instantiated for both `T=bool` **and** `T=int` (see the explicit instantiation lines at the bottom of this file). For `T=int`, the expression `value && sip_default_profiles.count(sip)` applies logical AND, which yields `0` or `1` — not the original `value`. If any `T=int` caller passes a non-0/1 value, it gets silently clamped.

Consider:
```cpp
params[sip] = std::make_shared<T>(sip_default_profiles.count(sip) ? value : T{});
```
This preserves `value` for default streams and falls back to the zero-value of `T` for non-default ones, which is correct for both `bool` and `int`.

realsense2_camera/src/profile_manager.cpp
```diff
@@ -86,6 +86,7 @@ void ProfilesManager::registerSensorUpdateParam(std::string template_name,
     // This function registers parameters that their modification requires a sensor update.
     // For each pair of stream-index, Function add a parameter to <params>, if does not exist yet, and advertise it by <template_name>.
     // parameters in <params> are dynamically being updated.
+    auto sip_default_profiles = getDefaultProfiles();
```
sysrsbuild-gh-agentic, line 89: ⚠️ **Potential issue (auto-generated by rs-agentic bot)**

`getDefaultProfiles()` throws `std::runtime_error` if `_all_profiles` is empty. Calling it unconditionally here introduces a hidden precondition on every caller of `registerSensorUpdateParam` — existing callers satisfy it, but future callers might not.

Also: for `MotionProfilesManager`, when **no** motion profiles are tagged default, `getDefaultProfiles()` falls back to only the **first** profile in `_all_profiles`. If both GYRO and ACCEL are present but untagged, only one ends up in the fallback map — the other stream would be disabled by default silently.

Consider moving the call inside the `if (params.find(sip) == params.end())` block (where the result is actually used), and add a comment about the `_all_profiles` precondition.

realsense2_camera/src/profile_manager.cpp
AviaAv: Fixed in caa80526 - `sip_default_profiles.count(sip) ? value : T{}` keeps the caller's value for tagged streams instead of collapsing it through `&&`.

realsense2_camera/src/profile_manager.cpp
AviaAv: Motion sensors don't use the base-class fallback: `MotionProfilesManager` overrides `getFallbackProfiles()` and returns the largest group of motion streams sharing a frame rate, so when gyro and accel are both untagged they both come back (a D430 GMSL gets both at 50 Hz). The single-profile fallback quoted here is `ProfilesManager::getFallbackProfiles`, which motion never reaches.

On the `_all_profiles` precondition - `getDefaultProfiles()` is already called unconditionally by `registerVideoSensorParams`, and both `registerProfileParameters` implementations populate `_all_profiles` before registering any parameter (the motion one also returns early when it is empty). A sensor exposing no profiles at all fails earlier than this.

realsense2_camera/src/profile_manager.cpp
sysrsbuild-gh-agentic: change approved

realsense2_camera/src/profile_manager.cpp
sysrsbuild-gh-agentic: Acknowledged — the explanation covers both concerns. `MotionProfilesManager` overriding `getFallbackProfiles` prevents the single-profile fallback issue, and `_all_profiles` being pre-populated means the throw precondition is always satisfied by the time this is reached. No further action needed.

Nir-Az (approved):

PR #3564, opened 2026-09-18
Proper cleanup in the driver's destructor

nnarain:
**Context**

In our application, I using realsense drivers as component nodes. In some cases I have to unload the component node, perform some operation and reload the drivers. I found that the underlying device context was not being closed on the destruction of the component node, and this was causing a seg fault.

```c++
void RealSenseNodeFactory::closeDevice()
{
    _is_alive = false;
    if (_query_thread.joinable())
    {
        _query_thread.join();
    }

    stopDevice();
    if (_device)
    {
        ROS_INFO("Closing RealSense device...");
        RSContextSingletonWrapper::getInstance().set_devices_changed_callback([](rs2::event_information&) {});

        // To go into unconfigured lifecycle state for lifecycled node we have to also disconnect the device
        _device = rs2::device();
    }
}
```

`closeDevice` already sets the `_is_alive` boolean and joins the thread. So I'm opted to just `closeDevice` from the `RealSenseNodeFactory`'s destructor.


sysrsbuild-gh:

realsense2_camera/src/realsense_node_factory.cpp
```diff
@@ -99,11 +99,7 @@ RealSenseNodeFactory::RealSenseNodeFactory(const std::string & node_name, const
 
 RealSenseNodeFactory::~RealSenseNodeFactory()
 {
-    _is_alive = false;
-    if (_query_thread.joinable())
-    {
-        _query_thread.join();
-    }
+    closeDevice();
```
sysrsbuild-gh-agentic, line 102: **[BUG – auto-generated by rs-agentic bot]**

`closeDevice()` can throw (e.g. `_realSenseNode.reset()`, RealSense SDK calls inside). Any uncaught exception escaping a destructor calls `std::terminate()`.

Suggest wrapping:
```cpp
RealSenseNodeFactory::~RealSenseNodeFactory()
{
    try { closeDevice(); } catch (...) {}
}
```
Or mark `closeDevice()` as `noexcept` (if all callsites allow it).

sysrsbuild-gh-agentic (commented):
*Auto-generated review by rs-agentic bot.*

